PR #798, opened 2023-03-10
DOCS-389 Create doc about remotes

JessamyT:


docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+
+Robots are organized into *parts*, where each part represents a computer (a single-board computer like a Raspberry Pi or a desktop, laptop, or other computer) running `viam-server`, the hardware [components](/components/) attached to it, and any [services](/services/) or other resources running on it.
+Every robot has a main part.
+Multi-part robots also have one or more *sub-parts* representing additional computers running `viam-server`.
+
+There are two ways to link robot parts:
+
+- **Sub-part**: If you have two computers within the *same robot*, use one as the main part and [connect the other to it as a sub-part](#configure-a-sub-part).
+
+- **Remote**: To connect two computers that are parts of *different robots*, [add one robot part as a remote part of the other robot](#configure-a-remote).
```
edaniels, line 23: Within robot parts, this makes sense calling it remotes. We are still missing one slight thing which is to just talk about a remote that is either running on some non viam-server SDK or a viam-server running outside of App. We don't have to make large mention of them, but it is important to spell it out in one of our `###`s.

edaniels (changes requested):
This is so much better

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
+---
+title: "Parts, Sub-Parts and Remotes"
```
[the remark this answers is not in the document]
JessamyT: What do you want it to be renamed to? `parts-remotes.md` or something?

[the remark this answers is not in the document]
edaniels: maybe parts-and-remotes!

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+description: "Connect robots to each other."
+tags: ["server", "components", "services"]
+---
+
+When robots can communicate with each other, they can share resources and operate collaboratively.
+This document explains how to establish secure connections between robots.
+
+### Robot parts
+
+Robots are organized into *parts*, where each part represents a computer (a single-board computer like a Raspberry Pi or a desktop, laptop, or other computer) running `viam-server`, the hardware [components](/components/) attached to it, and any [services](/services/) or other resources running on it.
```
mcvella, line 15:
```suggestion
Robots are organized into *parts*, where each part represents a computer (a [single-board computer](/installation/prepare) or a desktop, laptop, or other computer) running `viam-server`, the hardware [components](/components/) attached to it, and any [services](/services/) or other resources running on it.

```
We want to stop calling out Raspberry Pi specifically as our aim is to support many SBCs

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+- **Sub-part**: If you have two computers within the *same robot*, use one as the main part and [connect the other to it as a sub-part](#configure-a-sub-part).
+
+- **Remote**: To connect two computers that are parts of *different robots*, [add one robot part as a remote part of the other robot](#configure-a-remote).
+
+Connections between robots are established using the best network path available.
+
+When you configure a remote or a sub-part, the main robot part can access all the components and services configured on the remote robot part as though they were entities of the main robot part.
+This is a one-way connection: The main robot part can access the resources of the remote robot part, but the remote robot cannot access the resources of the robot part remoting into it.
+
+<img src="../img/remotes/remotes-diagram.png" alt="Example of a remote and a two part robot where the main (and only) part of robot 1 remotes into the main part of robot 2, and thus has access to all resources of robot 2." >
```
mcvella, line 30: The arrows being two way arrows do not clearly indicate that these are one-way connections.

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+
+Use the parts drop-down menu in the top banner of your robot's page on the [Viam app](https://app.viam.com) to add a new sub-part:
+
+![The Viam app interface with the part drop-down open. A new part called "my-sub-part" is being created.](../img/remotes/sub-part-config.png)
+
+<br>
+
+{{% alert title="Note" color="note" %}}
+
+When you create a sub-part, a connection is established between the main part and the sub-part without you needing to explicitly configure a remote.
```
mcvella, line 48: TBH I am not sure if this is still needed.  It feels already clear-cut to me.

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+
+![The Viam app CONFIG tab with the REMOTES sub-tab open.](../img/remotes/remote-create.png)
+
+1. Give the remote a name (you can just use the name of the other robot part, for example, "my-other-robot-main") and click **Create Remote**.
+2. Paste the `address` (for example, `my-other-robot-main.abc1de23f4.viam.cloud`) into the **Address** field.
+3. Click **Add Auth** and paste the `secret` from the other robot's **CODE SAMPLE** tab into the **Auth Key** field.
+
+![The Viam app CONFIG tab with a remote configured.](../img/remotes/remote-config.png)
+
+## Usage
```
mcvella, line 76: OK the heading level here is confusing to me.  Is this relevant to sub-parts and remote connected parts?  If so, what do we say "remote part" - can we call them that only when they are connected explicitly with a remote?  If this is supposed to be only for remotes, we need to make it more clear - "Usage" sounds generic to both.

mcvella (changes requested):

docs/manage/remotes.md
JessamyT: Yeah, I thought about that, but which way should the arrows go? The remote gets requests and sends data right?

docs/manage/remotes.md
JessamyT: Yeah fair enough but idk whether to remove it. Do you think it hurts anything?

docs/manage/remotes.md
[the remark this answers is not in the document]
JessamyT: I don't want to confuse people into thinking that by establishing a connection, they make two sub-parts of each other or two remotes of each other or something

docs/manage/remotes.md
JessamyT: You're right. How about:
```suggestion
## Use a remote resource with the SDKs
```

But actually I'm not clear: for sub-parts, is there a prepending of the part name? If so, I'll need to genericize this section. @edaniels ?

docs/manage/remotes.md
```diff
@@ -0,0 +1,92 @@
...
+
+![The Viam app CODE SAMPLE tab with Remotes selected and a copyable JSON snippet with the name, address and secret of the robot part.](../img/remotes/remote-address.png)
+
+5. Go to the Viam app robot page of the robot part from which you want to establish a remote connection.
+   This is the robot part that will be able to access the resources of the other robot part.
+6. Click the **CONFIG** tab, and then click the **REMOTES** sub-tab.
+
+![The Viam app CONFIG tab with the REMOTES sub-tab open.](../img/remotes/remote-create.png)
+
+1. Give the remote a name (you can just use the name of the other robot part, for example, "my-other-robot-main") and click **Create Remote**.
```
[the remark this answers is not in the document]
JessamyT: Because the linter was sneaky :p I will fix it :)

docs/manage/remotes.md
edaniels: for sub parts, yes when talking to the main part

docs/manage/remotes.md
JessamyT: I've gotten mixed messages about whether we are supposed to talk about running viam-server without connecting to the app.

Sorry what does ### mean? Does it mean in some doc somewhere or in a section here?

I'm not super confident in my understanding of how else you would use it or how much of an edge case that is. Should I include a note box at the bottom of "Configure a Remote" that says
"You can also configure a remote to:
 - An instance of `viam-server` that is not connected to the Viam app
 - A server other than `viam-server`"
 but then would we need to include instructions for how? How do they find the address and secret?

docs/manage/remotes.md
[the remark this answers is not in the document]
JessamyT: Okay I don't like saying that a part configures something since I think of the human as doing the configuring. I will change it to "Part A of Robot 1 has Part A of Robot 2 as a remote part." 

I don't understand what X would be. Resource?

JessamyT:
This is being migrated to https://github.com/viamrobotics/docs/pull/872 because of a docs breakage thing

docs/manage/remotes.md
edaniels: hmm. maybe let's just keep it as is for now and get feedback instead. I don't have any great ideas for how to section it up better right now

docs/manage/remotes.md
npentrel: yeah I think we can remove it - will add the suggestion on the other PR

JessamyT: Removed on the other PR

docs/manage/remotes.md
JessamyT: Done on the other PR

docs/manage/remotes.md
JessamyT: Updated on the other PR

PR #851, opened 2023-03-21
[OLD] Will close once confirmed there were no pending comments

npentrel:
Hi @bhaney - following the refactoring of the camera docs, I've started on these as well. It's a bit of a work in progress but I'd like to see what you think of the detection page before I work on the other ones.


npentrel:
Apologies - moving to a new PR - I'll keep this one open in case you had already started working on it. Since we published the docs some workflows got broken

PR #874, opened 2023-03-24
Remove exif

mcvella:


npentrel (approved):
what is exif? The test aren't fixed yet but this seems like a safe thing to merge anyway.

PR #916, opened 2023-03-29
DOCS 571: update SDK page with commands for installation and running code v2

sguequierre:
v2 of https://github.com/viamrobotics/docs/pull/811/files


docs/program/SDK-as-client.md
```diff
@@ -1,44 +1,53 @@
 ---
-title: "Using Our SDKs for a Client Application"
```
sguequierre, line 2: Definitely don't have to make this change, just a suggestion to make this more applicable from where in the docs we are linking to this page

docs/program/SDK-as-client.md
```diff
@@ -1,44 +1,53 @@
...
 - [Python SDK](https://python.viam.dev/)
-
 - [Go SDK](https://pkg.go.dev/go.viam.com/rdk)
-
 - [TypeScript SDK](https://ts.viam.dev/)
 
-## Quick Start Examples
+Click on the links above to read more about installation and usage of each SDK.
+
+## Installation
```
npentrel, line 24: tabs please

docs/program/SDK-as-client.md
```diff
@@ -186,6 +192,23 @@ main();
 {{% /tab %}}
 {{< /tabs >}}
 
+{{% alert title="Caution" color="caution" %}}
```
npentrel, line 195: This should stay above because it's important people see it - moving it here hides it a bit

docs/program/SDK-as-client.md
```diff
@@ -186,6 +192,23 @@ main();
...
+
+{{% /alert %}}
+
+## Run Your Code
+
+After saving the above code and adding control logic with Viam's API methods, run your program to control your Viam-connected robot.
+
+For example:
+
+- Python: `python3 ~/myCode/myViamFile.py`
```
npentrel, line 208: tabs - but also for typescript that is wrong I think? The way we use typescript is with a front end. So maybe remove that. The person to ask is @maximpertsov

maximpertsov: Correct, typescript must be run in a web browser which requires some additional setup - we should definitely remove the `tsc ...` line and maybe have a link to the quick-start example instead. What do you think about that @npentrel ?

maximpertsov: sadly there is no one-liner for how to run the typescript SDK at the moment :(

npentrel: Thanks Maxim! In that case I do think we just need to remove that and link to quick start docs if available or tutorials

maximpertsov: Yeah, for now we can just point user to the vanilla example tutorial you wrote up, which is already on this page.

docs/program/SDK-as-client.md
sguequierre: cool cool, lgtm!

docs/program/SDK-as-client.md
sguequierre: @maximpertsov @npentrel thanks guys! will make these changes

docs/program/SDK-as-client.md
sguequierre: np, done!

docs/program/SDK-as-client.md
```diff
@@ -183,13 +201,19 @@ async function main() {
...
-{{< /tabs >}}
+## Run Your Code
+
+After saving the above code and adding control logic with Viam's API methods, run your program to control your Viam-connected robot.
+
+For example:
+
+- Python: `python3 ~/myCode/myViamFile.py`
+- Go: `go run ~/myCode/myViamFile.py`
+- Typescript: [see execution demo](https://github.com/viamrobotics/viam-typescript-sdk/tree/main/examples/vanilla)
```
npentrel, line 212:
```suggestion
- Typescript: [see sample code](https://github.com/viamrobotics/viam-typescript-sdk/tree/main/examples/vanilla)
```

npentrel (approved):
LGTM % final comment

PR #926, opened 2023-03-30
Fix HTML test on light up bot link

JessamyT:


JessamyT:
This didn't fix it so I'm closing it

PR #959, opened 2023-04-03
DOCS-154: Add macOS instructions for webcam

npentrel:


docs/components/camera/webcam.md
```diff
@@ -20,10 +20,12 @@ Enter a name for your camera, select the type `camera`, and select the `webcam`
 <img src="../img/create-webcam.png" alt="Creation of webcam camera in the Viam app config builder." style="max-width:500px" />
 
 Fill in the attributes for your webcam view.
-If you click on the **Video Path** field while your robot is live, a drop down will autopopulate with identified camera paths.
+If you click on the **Video Path** field while your robot is live, a drop down autopopulates with identified camera paths.
 
 <img src="../img/configure-webcam.png" alt="Configuration of a webcam camera in the Viam app config builder." />
 
+Use the following configuration and fill in the attributes for your webcam view:
```
JessamyT, line 27:
```suggestion
Use the following configuration and fill in the attributes for your webcam:
```
Also these are mostly optional so maybe overwhelming?

docs/components/camera/webcam.md
```diff
@@ -20,10 +20,12 @@ Enter a name for your camera, select the type `camera`, and select the `webcam`
 <img src="../img/create-webcam.png" alt="Creation of webcam camera in the Viam app config builder." style="max-width:500px" />
 
 Fill in the attributes for your webcam view.
```
JessamyT, line 22:
```suggestion
Fill in the attributes for your webcam.
```

docs/components/camera/webcam.md
npentrel: That's fair but not really part of this PR. I'm also not currently sure how to address that because i think we do need a place where we list all the potential parameters... I guess an example that has only the common attributes would probably work

npentrel: I guess i can add that here

JessamyT: Yeah I think the example is good

JessamyT (approved):

PR #978, opened 2023-04-05
DOCS-645 Remove WorldState from arm and gantry API

JessamyT:


docs/services/frame-system.md
```diff
@@ -226,7 +226,7 @@ This *supplemental transform* is the missing link to be able to transform a pose
 
 Both TransformPose and FrameSystemConfig optionally take in these supplemental transforms.
 
-Functions of some services and components take in a WorldState parameter (like `ArmMoveToPosition`).
+Functions of some services take in a WorldState parameter (like `motion_service.move(component_name, destination, world_state)`).
```
spieswl, line 229: Nice catch

JessamyT: Thanks :)

docs/services/motion.md
```diff
@@ -147,7 +147,7 @@ This can be any pose, from the perspective of any component whose location is kn
 It will be converted into the frame of the component named in `component_name` when passed to `MoveToPosition`.
 
 **`world_state`**: This data structure is structured identically to what is described above in "Move", and is intended to behave the same.
```
spieswl, line 149: [major] L149 here indicates they are intended to behave the same, but in the implementation this is no longer true. `MoveSingleComponent` wraps the arm's `MoveToPosition` call, so the obstacle avoidance and collision checking features normally associated with passing in WorldState information are no longer available through `MoveSingleComponent`.

Actually, there might be some value in specifying that the Motion `Move` method is the ONLY way to have collision checking and obstacle avoidance given the recent changes.

spieswl (changes requested):
One detail on the Motion service docs should be addressed. Everything else looks perfect.

To your question about **Describing the Robot's Working Environment**, that is still valid information since the WorldState used there is passed into the Motion Service `Move` function (which is unchanged and still supports collision checking and obstacle avoidance).

docs/services/motion.md
JessamyT: So does it not actually take a world_state parameter? or does it take it and not use it?

spieswl: `MoveSingleComponent` still takes in a `WorldState` (see https://github.com/viamrobotics/rdk/blob/main/services/motion/builtin/builtin.go#L143) but only for the supplemental transforms that might be needed to plan arm motion with the Frame System (L163 of that same file)

spieswl: And I say "plan" but its really just to figure out if there's any math to be done for poses with respect to the arm and goal frames.

docs/services/motion.md
```diff
@@ -127,27 +127,34 @@ motion.move(component_name=armRes,destination=PoseInFrame(reference_frame="myArm
...
 
 An example of when this may be useful is if the user has implemented their own version of an arm, and wishes to use their own motion planning for it.
 That user will implement `MoveToPosition` on that arm using whatever method they desire to plan to the specified pose, and then can use this API endpoint to pass the destination in the frame of any other robot component.
 
+{{% alert title="Note" color="note" %}} <a id="move-vs-movetoposition">
+
 If this method is called with an arm which uses Viam’s motion planning on the backend, then this method is equivalent to using `robot.TransformPose` to transform the destination into the frame of the arm, and then calling `MoveToPosition` on the arm directly.
+Note that `arm.MoveToPosition` does not use `world_state`, so collision checking and obstacle avoidance _will not_ be performed.
+
+If you need collision checking and obstacle avoidance, use [`Move`](#move).
```
spieswl, line 142: Nice touch adding the alert.

spieswl (approved):
LGTM!

docs/services/motion.md
JessamyT: ah okay ty!

npentrel (approved):

sguequierre (approved):

PR #1016, opened 2023-04-11
Refactor fleet

npentrel:
# Description


docs/data/_index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: "Data"
+linkTitle: "Data"
+weight: 41
```
npentrel, line 4:
```suggestion
weight: 19
```

docs/fleet/_index.md
```diff
@@ -0,0 +1,98 @@
...
+
+- [Setup](#setup)
+- [Config](#config)
+- [Logs](#logs)
+- [History](#history)
+- [Code Sample](#code-sample)
+- [Control](#control)
+- [Security](#security)
+
+Soon you will also be able to [deploy code and machine learning models](#package-deployment) to robots.
```
JessamyT, line 25: I initially thought this meant that after reading this page I'd know how to deploy code etc.

Maybe this shouldn't be here at all if it doesn't exist yet? And certainly not as a highlighted feature (edit: though I now see we're already doing this on the current version anyway)

JessamyT (commented):
High level, I have the following thoughts:
1. What is "Organization" versus "Fleet Management" in the high level left nav? These seem redundant and like they should stay/be combined, plus "Organization" vs "Organizations" could be confusing. Maybe "Organization" -> "Fleet Management" and this "Fleet Management" page goes away or goes into the robot page?
2. I think the "Fleet management" page that immediately starts talking about tabs is odd because why do I care what the tabs are if I don't understand the overall concept? I think the current fleet management page does a better job of overviewing
3. The "Manage Robots" page seems a bit like a hodgepodge since it's called "Manage robots" but that title could apply to an explanation of locations and organizations, but then it goes all the way down into some configuration and API usage details.

npentrel:
1. Organization vs Fleet Management: Jon has given me the feedback that Fleet Management has a specific meaning in the industry and it doesn't include Organizations, Locations, Permissions and all of that. To me what you are saying makes perfect sense but if our audience doesn't think about it that way it makes sense to change it. 
2. I've updated the fleet page a bit to make it less abrupt. Page still needs love but it's less bad. Will create a ticket to further change.
3. Agreed on the manage robot page  - I've reordered the content a little but still needs more work. Will create a ticket to further change.

npentrel:
I will merge this for now and add the tickets

PR #1024, opened 2023-04-12
DOCS-515: Update Fragments page

npentrel:


docs/configuration/fragments.md
```diff
@@ -28,13 +28,23 @@ To add a fragment to a robot:
 - Look through the list of available fragments and click **ADD** next to any fragments you want to add to your robot.
 - Click **Save Config** at the bottom of the screen.
 
-{{% alert title="Note" color="note" %}}
+![The fragments subtab](../img/fragments-tab.png)
 
-The components or other resources included in the fragment will *not* appear in the **COMPONENTS** sub-tab or in the `components` section of your config.
-You will simply see the fragment ID in the `fragments` section of your config, but all the resources *will* appear on the **CONTROL** tab when your robot comes on line, and they will be accessible from SDK code as regular resources of your robot.
+The components or other resources included in the fragment will appear as read-only in the **COMPONENTS** sub-tab.
```
sguequierre, line 33: This sentence is a bit tricky: for my own clarification, wouldn't other resources just be services? and could these appear elsewhere from components sub-tab?

docs/configuration/fragments.md
```diff
@@ -28,13 +28,23 @@ To add a fragment to a robot:
...
+![The fragments subtab](../img/fragments-tab.png)
 
-The components or other resources included in the fragment will *not* appear in the **COMPONENTS** sub-tab or in the `components` section of your config.
-You will simply see the fragment ID in the `fragments` section of your config, but all the resources *will* appear on the **CONTROL** tab when your robot comes on line, and they will be accessible from SDK code as regular resources of your robot.
+The components or other resources included in the fragment will appear as read-only in the **COMPONENTS** sub-tab.
 
-{{% /alert %}}
+![A fragment in the components subtab](../img/fragment-components.png)
+
+In the `Raw JSON` configuration, you will simply see the fragment ID in the `fragments` section:
```
sguequierre, line 37:
```suggestion
In the `Raw JSON` configuration, you will see the fragment ID in the `fragments` section:
```
Opt.

npentrel: oh very good catch - not at all opt :D

docs/configuration/fragments.md
npentrel: it could also be processes or other resources. I think I need to update the glossary entry for that and link that here and not call out components specifically

sguequierre (approved):

PR #1031, opened 2023-04-12
fix typos in Jetson Orin documentation

penguinland:
The docs didn't quite match up with the Jetson Orin on my desk:
- Pins 3 and 5 are I2C bus 7, as shown by plugging something into those pins and running `i2cdetect -y -a -r 7`
- Pins 27 and 28 are I2C bus 1, as shown by an analogous check.
- I did _not_ swap the I2C `jetson-io.py` IDs, since they look correct (based on running `jetson-io.py` myself).
- The SPI bus is number 0, as found by reading from an ADC plugged into it. I'm happy to share my config if you want; setting this up is more complicated than I'd prefer.
- Things in the `/dev` column should start with slashes, to show that they're absolute paths.
- I changed the device names to the nearest things that actually exist in `/dev`. Are those the right devices? I'm unsure; our code doesn't interact with them directly. but they're the devices that exist on my Orin which are the closest to what was already in the docs, and I trust whoever wrote the previous version. 

I've never made a PR in this repo before. Once it's merged, what else do I need to do to get it deployed?

### DIFF
--- a/docs/installation/prepare/jetson-agx-orin-setup.md
+++ b/docs/installation/prepare/jetson-agx-orin-setup.md
@@ -75,9 +75,9 @@ If this command fails, try using `wget http://packages.viam.com/apps/viam-server
 
 | Data Sheet ID | GPIO Header Pin | Viam Bus ID | `jetson-io.py` ID | `/dev` Path ID | Notes |
 | ------------- | --------------- | ----------- | ----------------- | ----------- | ----- |
-| I2C_GP2_DAT, I2C_GP2_CLK | 3, 5 | `1` | `i2c2` | `dev/i2c2` | |
-| I2c_GP5_DAT, I2C_GP5_CLK | 27, 28 | `7` | `i2c8` | `dev/i2c8` | |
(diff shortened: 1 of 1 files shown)


sguequierre (commented):
Thank you for making a pr and for testing this out, these changes look good to me! I added the previous version of this table to this document in conjunction with @thegreatco from his setup notes, so I am going to ask him for review on this as well. I think the /dev path changes are accurate but I am not 100% sure about the bus IDs switch-up.

Oh and re. deployment --> once this is merged it should be deployed to production automatically!

penguinland:
This branch became out of date because someone else merged something else. I've rebased it, and will continue rebasing as necessary. There's still just the single commit.

sguequierre (approved):
LGTM but would suggest waiting for Pete to review/approve when he gets back from holiday before merging.

thegreatco:
I had opened https://viam.atlassian.net/browse/DOCS-262 during my investigation. I found that Pins 3/5 were i2c bus 1 and 27/28 were i2c bus 7. I guess my notes were wrong or something changed with nvidia?

thegreatco (approved):

penguinland:
My guess is that the notes were wrong because the notes are how I'd design the system myself and they seem perfectly reasonable. :stuck_out_tongue:

PR #1049, opened 2023-04-14
DOCS-662/555/391: SLAM service refactor

sguequierre:
* nested tabs for modes
* change attributes in templates to make pages more accurate
* add config builder tab & better instructions there
* simplify language at points
* update rplidar module instructions for intel/mac chip arch differences
* removing ORBSLAM documentation

PR #1068, opened 2023-04-18
DOCS-432 Update Docs with Results of Map Representation Project

jeremyrhyde:
This PR corrects one of the troubleshooting steps in the slam cartographer tutorial. Specifically, map representation did away with the JPEG endpoint and the issue discussed in this section, while still present, has changed slightly.

JIRA Ticket: https://viam.atlassian.net/browse/DOCS-432

### DIFF
--- a/docs/services/slam/run-slam-cartographer.md
+++ b/docs/services/slam/run-slam-cartographer.md
@@ -333,9 +333,7 @@ The STL file for an adapter plate is available on [GitHub](https://github.com/vi
 
 ### Issue: Maps JPEG not Appearing in UI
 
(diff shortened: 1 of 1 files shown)


sguequierre:
Hi @jeremyrhyde! I am adding this in to the equivalent file in the current big SLAM docs refactor that should be landing tomorrow: see https://github.com/viamrobotics/docs/pull/1049/files#r1170577816 --> Is that ok? Thanks for updating with this!

sguequierre (approved):
Thanks for adding this in! I added it in to the SLAM refactor branch going on right now as well, but feel free to merge this in as is: LGTM

PR #1069, opened 2023-04-18
Add installation alias

JessamyT:
Found this one in an old tutorial google doc

### DIFF
--- a/docs/installation/_index.md
+++ b/docs/installation/_index.md
@@ -11,6 +11,7 @@ aliases:
     - /getting-started/linux-install/
     - /installation/linux-install/
     - /installation/install/
+    - /getting-started/installation/
 
 ---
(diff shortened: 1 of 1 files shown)


docs/installation/_index.md
```diff
@@ -11,6 +11,7 @@ aliases:
     - /getting-started/linux-install/
     - /installation/linux-install/
     - /installation/install/
+    - /getting-started/installation/
```
npentrel, line 14: did we get feedback on this?

npentrel (approved):

docs/installation/_index.md
JessamyT: Not sure what you mean? I happened upon it when I clicked a link in an old intern tutorial google doc, and realized we don't have this one in the aliases but it might be in other external places besides this one google doc because it was the link for a while.

PR #1099, opened 2023-04-21
Update tutorial to avoid blinking confusion

npentrel:


JessamyT (approved):
LGTM. Why do the angle brackets let the markdown render properly? Seems fine but I thought it didn't used to work

PR #1118, opened 2023-04-25
Update URLs for install

npentrel:


npentrel:
We don't handle the app front end, though, I think a find and replace there is probably fairly easy for you to do :)

PR #1125, opened 2023-04-26
Try to make motor pin information more clear

JessamyT:


docs/components/motor/gpio/_index.md
```diff
@@ -118,40 +118,32 @@ The following attributes are available for `gpio` motors:
...
-Nested within `pins`:
+Inside the `pins` struct you need to configure **two or three** of the following, depending on your motor driver:
 
 | Name | Type | Inclusion | Description |
 | ---- | ---- | --------- | ----------- |
-| `a` | string | **Required** | See [Pin Information](#pin-information). Corresponds to "IN1" on many driver data sheets. Pin number such as "36." Viam uses board pin numbers, not GPIO numbers. |
-| `b` | string | **Required** | See [Pin Information](#pin-information). Corresponds to "IN2" on many driver data sheets. Pin number such as "36." Viam uses board pin numbers, not GPIO numbers. |
-| `dir` | string | **Required** | See [Pin Information](#pin-information). Pin number such as "36." Viam uses board pin numbers, not GPIO numbers. |
-| `pwm` | string | **Required** | See [Pin Information](#pin-information). Pin number such as "36." Viam uses board pin numbers, not GPIO numbers. |
+| `a` | string | **Required** for some drivers | Corresponds to "IN1" on many driver data sheets. Pin number such as "36." Viam uses board pin numbers, not GPIO numbers. |
```
sguequierre, line 137:
```suggestion
| `a` | string | **Required** for some drivers | Pin number of the GPIO pin on the board this driver's "IN1" pin is connected to. |
```
You could link to the "pin number" glossary index I'm adding in board pr here too to make the GPIO differentiation more clear. I think indicating the connection is good because otherwise its somewhat confusing where the number is coming from 

Regardless of deleting the last sentence I don't think you need to repeat it on every attribute!

sguequierre: Would suggest the same below & can indicate board PWM designation etc.

docs/components/motor/gpio/_index.md
```diff
@@ -118,40 +118,32 @@ The following attributes are available for `gpio` motors:
...
+## `pins`
 
-The attribute `max_rpm`is not required or available for [encoded](/components/motor/gpio/encoded-motor/) `gpio` motors.
+There are three common ways for your computer to communicate with a brushed DC motor driver chip.
+**Your motor driver data sheet will specify which one to use.**
 
-{{% /alert %}}
+- PWM/DIR: Two pins: One digital input (such as a GPIO pin) sends a [pulse width modulation (PWM)](https://en.wikipedia.org/wiki/Pulse-width_modulation) signal to the driver to control speed while another digital input sends a high or low signal to control the direction.
+  - Configure `pwm` and `dir`.
+- In1/In2 (also known as A/B): Two pins: One digital input is set to high and another set to low turns the motor in one direction and vice versa, while speed is controlled with PWM through one or both pins.
```
sguequierre, line 128:
```suggestion
- In1/In2: Two pins on the motor, often labeled In1 and In2 or A/B, connected to two GPIO pins on the board. One pin is set to high (active, >0), and another is set to low (inactive, 0V) to turn the motor in different directions depending on which pin is high. Speed is controlled with PWM through one or both pins.
```
Same suggestion format above--- No matter what I think nesting the colons should be edited!

sguequierre (changes requested):
Great changes but a couple comments, I think could be shortened and wiring could be made more clear!

docs/components/motor/gpio/_index.md
JessamyT: Okay we can add that link once that PR lands.
```suggestion
| `a` | string | **Required** for some drivers | Pin number this motor driver's "IN1" or "A" pin is wired to. Viam uses board pin numbers, not GPIO numbers. |
```

JessamyT: I think especially with the wiring diagrams and matching JSONs it should be okay?

docs/components/motor/gpio/_index.md
```diff
@@ -118,40 +118,32 @@ The following attributes are available for `gpio` motors:
...
 
-{{% alert title="Caution" color="caution" %}}
+## `pins`
 
-The attribute `max_rpm`is not required or available for [encoded](/components/motor/gpio/encoded-motor/) `gpio` motors.
+There are three common ways for your computer to communicate with a brushed DC motor driver chip.
+**Your motor driver data sheet will specify which one to use.**
 
-{{% /alert %}}
+- PWM/DIR: Two pins: One digital input (such as a GPIO pin) sends a [pulse width modulation (PWM)](https://en.wikipedia.org/wiki/Pulse-width_modulation) signal to the driver to control speed while another digital input sends a high or low signal to control the direction.
```
JessamyT, line 126:
```suggestion
- PWM/DIR: Use this if one of your motor driver's pins takes a [pulse width modulation (PWM)](https://en.wikipedia.org/wiki/Pulse-width_modulation) signal to the driver to control speed while another pin takes a high or low signal to control the direction.
```

sguequierre (approved):
LGTM

PR #1145, opened 2023-05-01
DOCS-706: Update transformation order

npentrel:


docs/components/camera/transform.md
```diff
@@ -74,59 +74,66 @@ The following attributes are available for `transform` views:
...
+        "low_threshold_pct": float,
+        "blur_radius_px": float
+    }
 }
 ```
 
 **Attributes:**
 
-- None
+- `high_threshold_pct`: The high threshold of ??? : between 0.0 - 1.0.
```
sguequierre, line 115: ??? I think needs to be clarified here and below! Sorry as I don't know what it is either

sguequierre (changes requested):
Would LGTM when ??? is addressed!

docs/components/camera/transform.md
npentrel: Oh amazing 😂

npentrel, line 115:
```suggestion
- `high_threshold_pct`: The high threshold value: between 0.0 - 1.0.
```

docs/components/camera/transform.md
```diff
@@ -74,59 +74,66 @@ The following attributes are available for `transform` views:
...
+- `low_threshold_pct`: The low threshold of ??? : between 0.0 - 1.0.
```
npentrel, line 116:
```suggestion
- `low_threshold_pct`: The low threshold value: between 0.0 - 1.0.
```

docs/components/camera/transform.md
npentrel: basing on this https://en.wikipedia.org/wiki/Canny_edge_detector

sguequierre (approved):

PR #1160, opened 2023-05-03
DOCS-719: Add additional example for data capture

npentrel:
Hi @agreenb - I think we were already using triplets for remote objects but I've added the camera one as an additional example - can you confirm everything is correct?


nataliajacobowitz:
this all looks good to me!

agreenb (approved):
Thank you!

PR #1190, opened 2023-05-08
DOCS-214: document utilizing extra params

sguequierre:
* I'm not super happy with the low level of detail this provides on utilizing within drivers but I think it's still an improvement? needs review: from what I've found these are not frequently used within engineering, but meant to provide the option in the api wrapper


docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
+---
+title: "Using Extra Params with Viam's SDKs"
+linkTitle: "Extra Params"
+weight: 40
+type: "docs"
+description: "Using extra parameters with Viam's SDKs."
+icon: "/services/img/icons/sdk.svg"
+tags: ["client", "sdk"]
```
sguequierre, line 8:
```suggestion
tags: ["sdk", "extra", "extend"]
```

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
...
+
+{{% alert title="Note" color="note" %}}
+
+If passing an object of type `None`, you do not have to specify `None` in the method call.
+
+{{% /alert %}}
+{{% /tab %}}
+{{% tab name="Go" %}}
+
+`extra (map[string]interface{})` indicates you are required to pass in an object of either type `map[string]interface{}` or type `nil` as a parameter when calling this method.
```
sguequierre, line 49:
```suggestion
`extra (map[string]interface{})` indicates you are required to pass in an object of either type `map[string]interface{}` or `nil` as a parameter when calling this method.
```

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
...
+For example:
+
+```go {class="line-numbers linkable-line-numbers"}
+func main() {
+    ... // Connect to the robot
+
+    // Get your resource from the robot
+    yourResource, err := YourResource.FromRobot(robot, "your-resource")
+
+    // Define a mapping containing extra information
```
sguequierre, line 62:
```suggestion
    // Define a map containing extra information
```

npentrel:
This looks generally good but I wonder if it should be framed as "Use extra Params with Modular Resources" because you'd only want to look at it if you're working with a custom resource? I think also probably worth getting reviewed by maybe Eric?

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
+---
+title: "Using Extra Params with Viam's SDKs"
+linkTitle: "Extra Params"
+weight: 40
+type: "docs"
+description: "Using extra parameters with Viam's SDKs."
```
npentrel, line 6:
```suggestion
description: "Use the extra parameter on the component API methods of modular resources."
```

Though I think it's not only component api methods that support it?

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
...
+icon: "/services/img/icons/sdk.svg"
+tags: ["sdk", "extra", "extend"]
+---
+
+Many component API methods have the option to pass in `extra` parameters.
+These are typed as `Optional[Dict[str, Any]]` in the Python SDK and `map[string]interface{}` in the Go SDK.
+
+Here's how to [pass correctly typed objects to](#define) and [utilize](#utilize) these `extra` parameters.
```
npentrel, line 14:
```suggestion
Here's how to [define](#define) and [utilize](#utilize) these `extra` parameters.
```

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Define
```
npentrel, line 16: If we can come  up with an example from a use case that would be better I think

sguequierre: @npentrel Hmmm, I think I can add the code from the ticket: https://viam.atlassian.net/jira/software/c/projects/DOCS/boards/31?modal=detail&selectedIssue=DOCS-214&quickFilter=40 to the "Utilize" section and adapt define-- will try that out

sguequierre: @npentrel Updated and reordered, LMK what you think!

sguequierre: @npentrel Going to page eric for review on this and based on his feedback, I also think it'd be a good idea to have it in /extend instead, and link to this from within the modular resources documentation?

edaniels:
Swapping myself for @randhid here!

docs/program/sdks/use-extra-params.md
sguequierre: @npentrel Eric recommended @randhid as a good resource for this!

docs/program/sdks/use-extra-params.md
sguequierre: Made more general with resource

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,140 @@
...
+How to [utilize](#utilize) and [define](#define) the `extra` parameters that many {{< glossary_tooltip term_id="resource" text="resource" >}} API methods offer in the Go and Python SDKs.
+
+## Utilize
+
+Use these `extra` parameters to pass information to a {{< glossary_tooltip term_id="resource" text="resource's" >}} driver that is not specified as a parameter in the [built-in resource type's API specification](/program/extend/modular-resources/#apis).
+
+To do this, you must code your own modified implementation of the resource type's API for a model.
+See [Extend Viam with Modular Resources](/program/extend/modular-resources/) for more information and [instructions](/program/extend/modular-resources/#use-a-modular-resource-with-your-robot) on modifying API specifications.
+
+An example of how to check the values of keys in an `extra` parameter of a [built-in resource API method](/program/sdks/#add-control-logic), the sensor component's [Readings](/components/sensor/#readings):
```
JessamyT, line 20:
```suggestion
An example of how to check the values of keys in an `extra` parameter of a [built-in resource API method](/program/sdks/#add-control-logic), the sensor component's [Readings](/components/sensor/#readings) method:
```
[opt]

docs/program/sdks/use-extra-params.md
```diff
@@ -0,0 +1,140 @@
...
+description: "Using the extra parameter on resource API methods with Viam's SDKs."
+icon: "/services/img/icons/sdk.svg"
+tags: ["sdk", "extra", "extend"]
+---
+
+How to [utilize](#utilize) and [define](#define) the `extra` parameters that many {{< glossary_tooltip term_id="resource" text="resource" >}} API methods offer in the Go and Python SDKs.
+
+## Utilize
+
+Use these `extra` parameters to pass information to a {{< glossary_tooltip term_id="resource" text="resource's" >}} driver that is not specified as a parameter in the [built-in resource type's API specification](/program/extend/modular-resources/#apis).
```
JessamyT, line 15:
```suggestion
Use `extra` parameters to pass information to a {{< glossary_tooltip term_id="resource" text="resource's" >}} driver that is not specified as a parameter in the [built-in resource type's API specification](/program/extend/modular-resources/#apis).
```

docs/program/sdks/use-extra-params.md
sguequierre: Agree sentence is a little awk but I don't think we need to specify method twice!

docs/program/sdks/use-extra-params.md
randhid: Not sure if I was pinged for a better use case or to verify that this works?

I usually do 

```go
out, ok:= extra["fieldname"].type //type is bool, string, something
if !ok {
return error
}
if out == something {
   doAfunction()
}
```

Which is equivalent to your go code example.

randhid: Let me know how I can be more helpful

sguequierre: @randhid No worries, just wanted you to verify the page's accuracy!

PR #1192, opened 2023-05-10
DOCS-739: Remove SLAM Service is an experimental feature

npentrel:


sguequierre (approved):
Yay for not being experimental!

PR #1210, opened 2023-05-15
DOCS-733: make context and error typing consistent

sguequierre:
* change `Context` to `ctx`
* remove naming from `error` 

(Match RDK specs)


JessamyT (approved):
Nice much better :D

PR #1215, opened 2023-05-15
Turn off MD034 no bare URLs lint rule

JessamyT:


npentrel (changes requested):
Let's discuss on slack

JessamyT:
Yes that's true!

Closing because we actually want this rule per discussion.

PR #1227, opened 2023-05-16
Update services and viam in 3 min image

npentrel:
Guess this also formatted that file... I think it should be fine


sguequierre (approved):
LGTM assuming html format changes are also intentional!

PR #1234, opened 2023-05-16
Fix nested type link formatting for MD034

JessamyT:
Per https://github.com/DavidAnson/markdownlint/blob/v0.28.2/doc/md034.md, nested square brackets should be escaped. Applied this, and also removed some backticks from types for consistency.


npentrel (changes requested):
Sorry - I thought you meant put slashed at the end of links - what does this do? escape the angular brackets and remove the code styling? I'm not sure either makes sense - the brackets don't need removing for hugo to do its thing and the code styling is there on purpose - though we don't have it everywhere. I'm sorry but since this is an issue only on your end, we need to find a solution that fixes the thing on your end or changes the markdownlint file only on your end

npentrel (approved):

PR #1239, opened 2023-05-17
Add joint positions imports to arm doc

JessamyT:


biotinker (approved):
LGTM

docs/components/arm/_index.md
```diff
@@ -86,6 +90,11 @@ from viam.components.arm import Arm
 ```go
 import (
   "go.viam.com/rdk/components/arm"
+  // To use MoveToPosition:
+  "go.viam.com/rdk/referenceframe"
+  "go.viam.com/rdk/spatialmath"
+  // To use MoveToJointPositions:
+  armapi "go.viam.com/api/component/arm/v1"
```
sguequierre, line 97: Why are you naming just this here? Not against it but (if I'm not confused haha) I think it's probably better without for simplicity
```suggestion
"go.viam.com/api/component/arm/v1"
```

sguequierre (changes requested):

docs/components/arm/_index.md
JessamyT: I just did it because it was done here https://docs.viam.com/tutorials/services/accessing-and-moving-robot-arm/#joint-position-commands but it doesn't have to be! But would there be other things called v1 or something?

docs/components/arm/_index.md
```diff
@@ -86,6 +90,11 @@ from viam.components.arm import Arm
 ```go
 import (
   "go.viam.com/rdk/components/arm"
+  // To use MoveToPosition:
+  "go.viam.com/rdk/referenceframe"
+  "go.viam.com/rdk/spatialmath"
+  // To use MoveToJointPositions:
```
JessamyT, line 96:
```suggestion
  // To use MoveToJointPositions ("armapi" name optional, but necessary if importing other packages called "v1"):
```
@sguequierre Do you think this is a good idea?

docs/components/arm/_index.md
JessamyT: @biotinker what do you suggest here? I kind of want to be consistent with the motion tutorials and such, but also want this to be generic, so curious what you think.

biotinker: IMO we should encourage our users to write software "properly", which in this case would mean renaming.

Someone could blindly copy the code if we didn't do the rename and get it to work, sure. But then if they try expand on our example code, if it's called `v1`, then we've laid a trap for them to hit and get a surprise error for no reason.

What would the rationale be for not renaming it?

PR #1245, opened 2023-05-18
DOCS-787: Add userflow AI assistant

npentrel:


andf-viam (commented):
The new UserFlow icon still appears to overlay the existing feedback icon. I am able to see the two icons superimposed on each other when using DarkReader, but without I only see the new UserFlow one. However, click a little to it's right and you'll still bring up the feedback widget.

I heard that Marketing is considering moving the feedback widget icon due to interference with the Viam App's red error messaging, also in the lower-right. So perhaps this will resolve itself? For now it does appear a bit wonky with the two together in the same corner.

Everything else about this is amazing!

- Using DarkReader, the two icons are superimposed:
	Screenshot 2023-05-23 at 11 14 08
- You can see the bottom&right border of the existing feedback icon, and clicking it opens the feedback icon:
	Screenshot 2023-05-23 at 11 14 29
	Screenshot 2023-05-23 at 11 15 17

andf-viam (approved):
Sounds good! LGTM then!

PR #1250, opened 2023-05-18
Clarify default values

JessamyT:
A suggestion based on this ticket https://viam.atlassian.net/browse/DATA-1377

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -170,6 +170,13 @@ Components of the same model are supported using the same low-level code.
 - `attributes`: A struct to define things like how the component is wired to the robot, its dimensions, and other specifications; attributes vary widely between models.
   See the [component documentation](/components/) for a given component type and model for more details.
 
+{{% alert title="Tip" color="tip" %}}
+
+Some optional attributes have default values.
+If you omit these attributes from your config, or if you include them but leave their values empty, RDK will apply the default values even though the default values are not shown in the config.
+
+{{% /alert %}}
+
 - `depends_on`: Any components that a given component relies upon, and that must be initialized on boot before this component is initialized.
   Many built-in components have convenient implicit dependencies, in which case `depends_on` can be left blank.
   For example, a [`gpio` motor](/components/motor/gpio/) depends on the `board` to which it is wired, but it has a dedicated `board` attribute and `viam-server` will automatically initialize that board before it looks for the motor.
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -173,7 +173,7 @@ Components of the same model are supported using the same low-level code.
 {{% alert title="Tip" color="tip" %}}
 
 Some optional attributes have default values.
-If you omit these attributes from your config, or if you include them but leave their values empty, RDK will apply the default values even though the default values are not shown in the config.
+If you omit these attributes from your config, or if you include them but leave their values empty, `viam-server` will apply these default values at runtime, even though they are not reflected in the configuration file.
 
 {{% /alert %}}
 


docs/manage/configuration.md
```diff
@@ -170,6 +170,13 @@ Components of the same model are supported using the same low-level code.
 - `attributes`: A struct to define things like how the component is wired to the robot, its dimensions, and other specifications; attributes vary widely between models.
   See the [component documentation](/components/) for a given component type and model for more details.
 
+{{% alert title="Tip" color="tip" %}}
+
+Some optional attributes have default values.
+If you omit these attributes from your config, or if you include them but leave their values empty, RDK will apply the default values even though the default values are not shown in the config.
```
npentrel, line 176:
```suggestion
If you omit these attributes from your config, or if you include them but leave their values empty, the RDK will apply the default values even though the default values are not shown in the config.
```

Also not sure whether it's are not shown or are not in the config

JessamyT: I've always thought it was "RDK" without "the." This isn't part of our style guide--maybe we should nail it down one way or the other?
I don't understand the question; I guess they're not in the config but they sort of are in a way behind the scenes maybe? But the point is that the user won't see them configured, which is what I meant to convey here.

JessamyT: (Have heard other people say "RDK" without an article as well, but maybe that's just a Bend office thing??)

andf-viam: If we want RDK to evoke SDK, we should consider matching its usage: `the SDK`, `our SDK`, `the Python SDK`. Omitting the article is perhaps more appropriate for named products, like `ROS` or `RoboDevoKito`.

I don't mind either personally, as long as we are consistent throughout. Sounds like a check in with Product? Sounds like you were thinking the same thing!

`are not shown` -- in this case, I often see the phrase `are not reflected` to indicate that the user won't see the values where they otherwise might expect them. So maybe: `the RDK will apply these default values, even though they are not reflected in the configuration shown under the **Config** tab.` or something similar.

JessamyT: Asking!

I like that suggestion, or maybe "...will apply these default values at runtime, even though they are not reflected in the configuration file"

JessamyT:
```suggestion
If you omit these attributes from your config, or if you include them but leave their values empty, `viam-server` will apply these default values at runtime, even though they are not reflected in the configuration file.
```
@mcvella Since you're part of that Slack thread bringing you in. Does it make more sense to say `viam-server` or the RDK in here?

mcvella: Why not just say:

If you omit these attributes from your config, or if you include them but leave their values empty, default values will be applied at runtime, even though they are not reflected in the configuration file.

andf-viam: Jessamy looking for `RDK` vs `the RDK` guidance!

JessamyT: Well, I didn't say it that way because that would be passive voice, which we try to avoid to reduce vagueness.

JessamyT: I can just apply
```suggestion
If you omit these attributes from your config, or if you include them but leave their values empty, default values will be applied at runtime, even though they are not reflected in the configuration file.
```
but similar cases appear throughout the docs so I do think this conversation is broadly applicable and therefore worthwhile!
If it is the case that how default values are applied is actually super complicated/technical and can't be accurately described by saying that `viam-server` does it, then perhaps passive voice really is the best option here, which would be good to know for this and other similar cases.

JessamyT: Ooh per Slack conversation, will use `viam-server`

npentrel (approved):

PR #1255, opened 2023-05-19
Add note about ENA and ENB not being enable pins

JessamyT:
This is a weird case and I don't want users thinking that enable pins typically equate to PWM pins.

### DIFF
--- a/docs/try-viam/rover-resources/rover-tutorial-fragments.md
+++ b/docs/try-viam/rover-resources/rover-tutorial-fragments.md
@@ -51,6 +51,14 @@ The fragment adds the following components to your robot's JSON configuration:
   * Within the motor attributes, board: "local", encoder: "Lenc", ticks per rotation: "996".
   * Within the component pin assignment, type: In1/In2, A/In1: "11 GPIO 17", B/In2: "13 GPIO 27", PWM: "15 GPIO 22".
   * Depends on local and Lenc.
+
+{{% alert title="Note" color="note" %}}
+
+This particular motor driver has pins labeled "ENA" and "ENB."
+Typically, this would suggest that they should be configured as enable pins, but on this specific driver these function as PWM pins, so we configure them as such.
+
+{{% /alert %}}
+
 * A wheeled Viam [base](/components/base/) with attributes:
   * Right Motors: right
   * Left Motors: left


docs/try-viam/rover-resources/rover-tutorial-fragments.md
```diff
@@ -51,6 +51,14 @@ The fragment adds the following components to your robot's JSON configuration:
   * Within the motor attributes, board: "local", encoder: "Lenc", ticks per rotation: "996".
   * Within the component pin assignment, type: In1/In2, A/In1: "11 GPIO 17", B/In2: "13 GPIO 27", PWM: "15 GPIO 22".
   * Depends on local and Lenc.
+
+{{% alert title="Note" color="note" %}}
+
+This particular motor driver has pins labeled "ENA" and "ENB."
```
andf-viam, line 57: One nit about location: `motor driver` as a phrase appears [here](https://docs.viam.com/try-viam/rover-resources/rover-tutorial/#motor-driver), but not anywhere on this page. On this page, does this alert refer to the `Right gpio motor` and `Left gpio motor`?

Should this alert instead live closer to where we talk about wiring these motors? Perhaps the pinout table below the wiring diagram [here](https://docs.viam.com/try-viam/rover-resources/rover-tutorial/#connect-the-wires)? There is a `Tip` there that seems to be closely related to this alert: maybe the two should live close to one another, or be merged in some fashion?

Alert text itself LGTM! Approving, and leaving placement up to you!

andf-viam (approved):
LGTM % one thought about placement. Thanks!

docs/try-viam/rover-resources/rover-tutorial-fragments.md
JessamyT: Yeah it applies to both right and left gpio motors.

RE the phrase motor driver not appearing here, I think that's okay because it's already been introduced before folks will come to this doc, and because this doc needs refactoring anyway (I made a ticket for it yesterday). I'm just making this lil PR now because it came up yesterday and I was confused about it, so I could certainly see users being confused too--the PWM pin configuration is a very frequently asked question and cause of problems!

Hmm my take is that we don't talk about it being a PWM pin on that first page in that tip or in the diagram or paper leaf, so I think it makes more sense to talk about here, when actually configuring the thing. Will merge as-is and we can change later if necessary!

PR #1265, opened 2023-05-22
DOCS-786: Remove long inline code

npentrel:


JessamyT (approved):
LGTM. There are some other imports like this where I was thinking of doing this too--could add them to this PR if you feel so inclined.

npentrel:
Not sure which ones but we can fix them as we see them :)

PR #1287, opened 2023-05-25
DOCS-603: add note for power management api error

sguequierre:
* Add a requested note for an issue with the set power mode method in the board API.
* Clean up and fix board status method parameters, I had added power mode in there erroneously 🤦‍♀️


andf-viam (approved):
LGTM with one optional wording suggestion.

PR #1290, opened 2023-05-25
change BaseClient to Base in python base samples

sguequierre:
* these samples were added before we had gotten clarification that Base effectively wraps BaseClient, this had slipped by me until now!


andf-viam (approved):
LGTM with one question: does [try-viam-sdk.md](https://github.com/viamrobotics/docs/blob/main/docs/tutorials/get-started/try-viam-sdk.md) need the same treatment? Not sure myself!

PR #1291, opened 2023-05-25
remove unnecessary capitalization 

sguequierre:
* Randomly noticed this -> remove caps post colon  on /manage/parts-and-remotes/, I think that is correct?


andf-viam (approved):
You are absolutely correct, thank you!! LGTM!

PR #1309, opened 2023-05-31
DOCS-722: Lazy load video gifs

npentrel:


layouts/partials/scripts.html
```diff
@@ -73,6 +73,7 @@
 <!-- scripts for prism -->
 <script src='{{ "js/prism.js" | relURL }}'></script>
 {{ end }}
+<script type="module" src='{{ "js/yall.js" | relURL }}'></script>
```
mrloureed, line 76: Is this a lazy loading library. You may be able to write a script that uses intersection observer instead.

npentrel: yes - using https://github.com/malchata/yall.js . Does it look like I am importing that correctly? I could probably write my own logic... yes... but why reinvent the wheel?

layouts/docs/content.html
```diff
@@ -26,7 +26,7 @@ <h4 class="alert-heading">DRAFT</h4>
 			<div class="col hover-card hover-card-small">
 			  <a href="{{.RelPermalink}}">
 				{{ if .Params.webmSrc }}
-				  <video autoplay loop muted playsinline preload="auto" alt="{{ .Params.videoAlt }}" width="100%">
+				  <video autoplay loop muted playsinline alt="{{ .Params.videoAlt }}" width="100%" class="lazy">
```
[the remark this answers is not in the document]
npentrel: these are always gif-like videos

npentrel:
Hey @mrloureed I got this to work with [lozad.js](https://github.com/ApoorvSaxena/lozad.js/). I've decided I'd prefer using a library over creating my own logic with intersection observers - since I won't be able to do it better than these and this is super lightweight. If you can give this a quick review when you're back to check that it's working and that I'm including this as it should be, I'd appreciate that :).

mrloureed (approved):
LGTM!

PR #1355, opened 2023-06-06
Link to RPLidar module docs from SLAM docs

dannenberg:
i struggled to find this page and just saw a discord user who was also struggling to. i think this may help. 

i was also considering reworking the requirements and configuration of the cartographer page to make the rplidar steps harder to miss, but that is a less trivial reworking.


sguequierre (approved):
LGTM Good idea

PR #1359, opened 2023-06-07
DOCS-822: Add I2C error to RPi troubleshooting

andf-viam:
Add I2C error encountered when missing the `raspi-config` step to RPi troubleshooting section.

- Because this is specific to the Pi only, opted not to include this in the main Troubleshooting section.
- Took the opportunity to clean up Next Steps section, which referred to a Troubleshooting item as having already been completed.


andf-viam:
Testing a strange push issue. Closing this ticket in favor of next.

PR #1367, opened 2023-06-07
Remove text shadow from code-terminal

andf-viam:
Testing a CSS fix to the code shadows in `code-terminal` code blocks.


npentrel (approved):
Oh perfect - I noticed something was off before but never figured out what

PR #1372, opened 2023-06-08
DOCS-839: Explore welcome page intro

andf-viam:
Consider some alternate options to how we introduce our Welcome page.

Staged options to choose from:

- [Absolute basic](https://docs-test.viam.dev/438c5ec82347f0e731417b7295117700e9182672/public/index.html)
	- Swapped title to first sentence, and trimmed title to absolute minimal.
- [Absolute basic, with text separator](https://docs-test.viam.dev/93439c77e90d2553e45e10edf2c92f6752d03a6c/public/index.html)
	- Same as above, but with single intro sentence between two image sections.
- [Two more carousels](https://docs-test.viam.dev/e66668c9d417e75a4a871134273097d539e9c7a2/public/index.html)
	- Added image carousels for components and services, and swapped board to be below getting started tiles.

Questions for team:

- Do we find additional carousels useful? Is the single sentence text intro to each necessary?
- Is the paragraph break between getting started tiles and the components carousel useful, or duplicative?
	- I felt that two large image scroll sections (getting started tiles & three image carousels) blend together without a text separator
- How do we feel about the order of presented resources? Should the 4 getting started tiles should be before or after the carousels?
	- My thoughts: before, given content, but after, given space consumed; with a lean towards before (content is king, they say).

TODO:

- Investigate why there is an extra newline between the components and services carousels!

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -107,7 +107,7 @@ Viam is a complete software platform for robots that runs on Linux and macOS and
 </div>
 <br>
 
-Explore the following resources to jump right in:
+Explore the following to learn more about the Viam platform or to try it out for yourself:
 
 <div class="container td-max-width-on-larger-screens">
   <div class="row">
@@ -127,7 +127,7 @@ Explore the following resources to jump right in:
             <ol style="padding-inline-start: 1.1rem">
             <li><a href="manage/configuration/">Configure your robot<a> or <a href="manage/fleet/">fleet</a></li>
             <li><a href="installation/">Install Viam on your robot</a></li>
-            <li>Configure robot <a href="components/">components</a> and add <a href="services/">services</a></li>
+            <li><a href="components/">Configure robot components</a> and <a href="services/">add services</a></li>
             <li><a href="manage/fleet/robots/#control">Control and test your robot</a></li>
             </ol>
             {{<gif webm_src="img/blink.webm" mp4_src="img/blink.mp4" alt="A blinking L.E.D. connected to a Raspberry Pi">}}
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -11,7 +11,6 @@ sitemap:
 ---
 
 Welcome to the Viam Documentation!
-
 Viam is a complete software platform for robots that runs on any 64-bit Linux OS and macOS.
 Viam supports a wide variety of systems, including:
 
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -11,63 +11,7 @@ sitemap:
 ---
 
 Welcome to the Viam Documentation!
-Viam is a complete software platform for robots that runs on any 64-bit Linux OS and macOS.
-
-Explore the following resources to jump right in:
-
-<div class="container td-max-width-on-larger-screens">
-  <div class="row">
-    <div class="col landing-hover-card">
-        <div class="landing-hover-card-padding yellow">
-            <h4>Learn and Try</h4>
-            <p style="text-align: left;">
-            Learn about <a href="viam/">the Viam platform in 3 minutes</a> and then
-            <a href="try-viam/">drive a Viam rover</a> from the comfort of your home or follow along with a <a href="tutorials/"> tutorial</a>.</p>
-            {{<gif webm_src="img/rover.webm" mp4_src="img/rover.mp4" alt="A Viam Rover moving about">}}
-        </div>
-    </div>
-    <div class="col landing-hover-card ">
-        <div class="landing-hover-card-padding purple">
-        <h4>Configure your robots</h4>
-        <div style="text-align: left">
-            <ol style="padding-inline-start: 1.1rem">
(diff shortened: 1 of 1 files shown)


npentrel:
I like the changes in the beginning with the welcome.

Answers:
**Do we find additional carousels useful? Is the single sentence text intro to each necessary?**

Carousels are kind of [an anti-pattern](https://thegood.com/insights/ecommerce-image-carousels/). I think it suits our purposes for the boards because if people want to look at other boards they can but if they don't the impression they get from the ones that are there are enough. I don't think the same holds true for services and components where I would always want all to be visible. Which we could do but that may be putting too much information in front of users.

**Is the paragraph break between getting started tiles and the components carousel useful, or duplicative?**
Definitely too much duplication but I do like the paragraph most about this. I think what I'd suggest though is to try and merge the content with the second tile (configure your robots). Maybe remove the second item (since it's somewhat covered by 1.) and then say Configure your robot's components and services? You could also split the div with the cards into two and add a paragraph in between.

**Investigate why there is an extra newline between the components and services carousels!**
That's probably something to do with the margin on the h4 but I can't easily see what's going wrong either. 

For the ordering with the boards - they could go below the fold. That's not a bad idea but I'd love to hear what @sguequierre and @JessamyT think on that

JessamyT:
**Are carousels useful?...**
I think more carousels is too many carousels, and they're a little hard to use on mobile. I agree with Naomi's points. These lists are also not analogous to boards because they're lists of types, not models, so it's just enough info to be too much while also not being enough to understand fully what we actually support.

**Paragraph break?**
Doesn't this become moot if we don't keep the component and service carousels? Either way I think it's better if we can get people to just go read Viam in 3 minutes.

**RE the line of text between the board carousel and the other cards:** I think you're right Andrew that they currently sort of blend together, but I wonder if we could just put a thin horizontal line, or a blank line break, in there instead of more text, since I think it's hard to come up with text that is actually useful there.

andf-viam:
Thanks for the great feedback @npentrel and @JessamyT !

I've removed the extra carousels (were fun to explore!) and merged helpful paragraph text into the second card. I think we still need the word `installation` prominently findable in that card, so I've reshuffled things a little. Now: install before configure (general mindset) even though you can perform these tasks in inverse order (Viam approach, via app, then DL configured conf file).

Explored a fifth card to try to explore, say, "Install & Configure" vs "Control & Extend" but it looks much better with 4 cards. Maybe if we had a good idea for a 6th card also? Don't want to overdo the landing page though. 

Let me know what you think! Happy to fine-tune.

docs/_index.md
```diff
@@ -105,6 +105,10 @@ Viam supports a wide variety of systems, including:
   <div class="prev" style="display: block">‹</div>
   <div class="next" style="display: block">›</div>
 </div>
+<br>
+
+Viam is designed to be flexible and extensible, and to save you time from prototype to production.
```
JessamyT, line 110: This strikes me as potentially too sales-y for the docs page.

andf-viam: Your probably right. Trying for that elusive padding:  `Explore the following resources to jump right in:` just seems like too little between sections.

npentrel:
So I think I'm ok with not featuring modular resources prominently right now because it is more of an advanced usage and we don't have many yet. Once we do we may want to add something at the bottom that features popular modular resources or something like that.

docs/_index.md
```diff
@@ -122,10 +125,10 @@ Viam supports a wide variety of systems, including:
         <h4>Configure your robots</h4>
         <div style="text-align: left">
             <ol style="padding-inline-start: 1.1rem">
-            <li><a href="/manage/fleet/robots/">Set up your robot</a> or<a href="manage/fleet/"> fleet</a></li>
             <li><a href="installation/">Install Viam on your robot</a></li>
-            <li><a href="manage/configuration/">Configure your robot</a></li>
-            <li><a href="manage/fleet/robots/#control">Test your robot</a></li>
+            <li><a href="manage/configuration/">Configure your robot<a> or <a href="manage/fleet/">fleet</a></li>
```
npentrel, line 129: This ordering doesn't make sense. Configuring the fleet happens before installing viam.

docs/_index.md
```diff
@@ -105,6 +105,9 @@ Viam supports a wide variety of systems, including:
   <div class="prev" style="display: block">‹</div>
   <div class="next" style="display: block">›</div>
 </div>
+<br>
+
+Explore the following resources to jump right in:
```
JessamyT, line 110: Oooh let's not use the word "resources" here since it's got another meaning. How about
```suggestion
Get started with Viam:
```
I don't love that either but it's less potentially confusing? Open to other things for sure

andf-viam: Oop! You're definitely correct! Will fixup -- thanks!

npentrel (approved):
LGTM % final comment

PR #1383, opened 2023-06-12
Capitalize V for volt in MAIV tutorial

JessamyT:
(nit)

PR #1390, opened 2023-06-14
DOCS-850: Add rule to capitalize V for Voltage

npentrel:
we can't include voltage in the units role without it triggering for alt texts but we can make sure it gets capitalized. The former hasn't been an issue tbf


andf-viam (approved):
LGTM, thanks!

EDIT: Nevermind, didn't see https://github.com/viamrobotics/docs/pull/1383. All good!

PR #1396, opened 2023-06-14
Make quick-fixes on new navigation service docs 

sguequierre:
* change capitalization and formatting
* remove empty comment

### DIFF
--- a/docs/services/navigation/_index.md
+++ b/docs/services/navigation/_index.md
@@ -24,7 +24,7 @@ You must configure a [base](/components/base/) with a [movement sensor](/compone
 
 Navigate to the **Config** tab of your robot's page in [the Viam app](https://app.viam.com).
 Click on the **Services** subtab and navigate to the **Create service** menu.
-Select the type `Navigation` and enter a name for your service.
+Select the type `navigation` and enter a name for your service.
 
 Click **Create service**:
 
@@ -34,7 +34,6 @@ Click **Create service**:
 {{% tab name="JSON Template" %}}
 
 ```json {class="line-numbers linkable-line-numbers"}
-// 
 {
     "name": "your-navigation-service",
     "type": "navigation", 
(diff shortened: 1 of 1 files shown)


docs/services/navigation/_index.md
```diff
@@ -34,7 +34,6 @@ Click **Create service**:
 {{% tab name="JSON Template" %}}
 
 ```json {class="line-numbers linkable-line-numbers"}
-// 
 {
```
sguequierre, line 37:
```suggestion
// "services": [  ...
{
```
For whoever reviews this; would something like the above or `// <INSERT SERVICES JSON>` or similar be good for making it more clear what users need to do here?

npentrel: I like the way it is in https://docs.viam.com/services/vision/detection/ ? Have a look and see if you like that or if that gives you other ideas

npentrel (commented):
Let me know when ready for re-review

docs/services/navigation/_index.md
```diff
@@ -99,7 +101,7 @@ The following attributes are available for `Navigation` services:
 
 | Name | Type | Inclusion | Description |
 | ---- | ---- | --------- | ----------- |
-| `store` | obj | **Required** | The configuration to use for the Navigation data storage. There are two parameters: `type` (`"memory"` or `"mongodb"`) and `config`. <br> Examples: `"store": [{"type": "memory"}]`, `"store": [{ "type": "mongodb", "config": {"uri": "mongodb://127.0.0.1:27017"}]` |
+| `store` | obj | **Required** | The configuration to use for the Navigation data storage. There are two parameters: `type` (`"memory"` or `"mongodb"`) and `config`. <br> Examples: <br> `"store": [{"type": "memory"}]` <br> `"store": [{ "type": "mongodb", "config": {"uri": "mongodb://127.0.0.1:27017"}]` |
```
npentrel, line 104:
```suggestion
| `store` | obj | **Required** | The configuration object to use for the Navigation data storage. Either `[{"type": "memory"}]` or `[{ "type": "mongodb", "config": {"uri": "mongodb://127.0.0.1:27017"} }]`. |
```

The reason I think that's better is because for memory you don't need the second parameter.

PR #1399, opened 2023-06-15
DOCS-845: Clarify coordinate system definitions

JessamyT:
# Description

Clarify base API coordinate system.

Remove note in frame system doc because the frame system has no inherent "forward" etc.; it's all relative and user-defined.

### DIFF
--- a/docs/components/base/_index.md
+++ b/docs/components/base/_index.md
@@ -156,9 +156,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
 **Parameters:**
 
 - `angle` [(float)](https://docs.python.org/3/library/functions.html#float): The angle to spin in degrees.
-Positive implies a given number of degrees to the left.
-- `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The angular velocity at which to spin in degrees per second.
 Positive implies turning to the left.
+- `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The angular velocity at which to spin in degrees per second.
+Given a positive angle and a positive velocity, the base turns to the left (for built-in base models).
 
 **Returns:**
 
@@ -180,9 +180,9 @@ await my_base.spin(angle=10, velocity=1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `angleDeg` [(float64)](https://pkg.go.dev/builtin#float64): The angle to spin in degrees.
-Positive implies a given number of degrees to the left.
-- `degsPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The angular velocity at which to spin in degrees per second.
 Positive implies turning to the left.
+- `degsPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The angular velocity at which to spin in degrees per second.
+Given a positive angle and a positive velocity, the base turns to the left (for built-in base models).
 - `extra` [(map\[string\]interface{})](https://go.dev/blog/maps): Extra options to pass to the underlying RPC call.
 
 **Returns:**
@@ -218,7 +218,7 @@ Set the linear and angular power of the base, represented as a percentage of max
 - `angular` [(Vector3)](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.Vector3): The percentage of max power of the base's angular propulsion.
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
   Use the Z component of this vector to spin left or right when controlling a wheeled base.
-  Positive "Z" values imply spinning to the left.
+  Positive "Z" values imply spinning to the left (for built-in base models).
 
 **Returns:**
 
@@ -260,7 +260,7 @@ Negative "Y" values imply moving backwards.
 - `angular` [(r3.Vector)](https://pkg.go.dev/github.com/golang/geo/r3#Vector): The percentage of max power of the base's angular propulsion.
 In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
 Use the Z component of this vector to spin left or right when controlling a wheeled base.
-Positive "Z" values imply spinning to the left.
+Positive "Z" values imply spinning to the left (for built-in base models).
 - `extra` [(map\[string\]interface{})](https://go.dev/blog/maps): Extra options to pass to the underlying RPC call.
 
 **Returns:**
--- a/docs/components/base/_index.md
+++ b/docs/components/base/_index.md
@@ -91,9 +91,9 @@ Move the base in a straight line across the given distance (mm) at the given vel
 **Parameters:**
 
 - `distance` [(int)](https://docs.python.org/3/library/functions.html#int): The distance to move in millimeters.
-Negative implies backwards.
+Positive implies forwards.
 - `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The velocity at which to move in millimeters per second.
-Negative implies backwards.
+Positive implies forwards.
 
 **Returns:**
 
@@ -118,9 +118,9 @@ await my_base.move_straight(distance=10, velocity=-1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `distanceMm` [(int)](https://pkg.go.dev/builtin#int): The distance to move the base in millimeters.
-Negative implies backwards.
+Positive implies forwards.
 - `mmPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The velocity at which to move the base in millimeters per second.
-Negative implies backwards.
+Positive implies forwards.
 - `extra` [(map\[string\]interface{})](https://go.dev/blog/maps): Extra options to pass to the underlying RPC call.
 
 **Returns:**
@@ -152,9 +152,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
 **Parameters:**
 
 - `angle` [(float)](https://docs.python.org/3/library/functions.html#float): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies forwards.
 - `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies forwards.
 
 **Returns:**
 
@@ -176,9 +176,9 @@ await my_base.spin(angle=10, velocity=1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `angleDeg` [(float64)](https://pkg.go.dev/builtin#float64): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies forwards.
 - `degsPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies forwards.
 - `extra` [(map\[string\]interface{})](https://go.dev/blog/maps): Extra options to pass to the underlying RPC call.
 
 **Returns:**
@@ -209,11 +209,11 @@ Set the linear and angular power of the base, represented as a percentage of max
 - `linear` [(Vector3)](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.Vector3): The percentage of max power of the base's linear propulsion.
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
   Viam's coordinate system considers +Y to be the forward axis (+/- X left/right, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
-  Negative "Y:" values imply moving backwards.
+  Positive "Y" values imply moving forwards.
 - `angular` [(Vector3)](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.Vector3): The percentage of max power of the base's angular propulsion.
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
   Use the Z component of this vector to spin left or right when controlling a wheeled base.
-  Negative "Z:" values imply spinning to the right.
+  Positive "Z" values imply spinning to the left.
 
 **Returns:**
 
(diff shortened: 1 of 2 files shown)


docs/components/base/_index.md
```diff
@@ -152,9 +152,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
 **Parameters:**
 
 - `angle` [(float)](https://docs.python.org/3/library/functions.html#float): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies forwards.
```
Otterverse, line 155: what does "foward" mean when spinning? It's not a head-over-heels somersault, so...?

docs/components/base/_index.md
```diff
@@ -152,9 +152,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
...
 - `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies forwards.
```
Otterverse, line 157: Same as above... what is "forward" in terms of spin?

docs/services/frame-system/_index.md
```diff
@@ -86,13 +86,6 @@ The `Orientation` parameter offers `Types` for ease of configuration, but the Fr
 
 {{% /alert %}}
 
-{{% alert title="Tip" color="tip" %}}
-
-Viam's coordinate system considers `+X` to be forward, `+Y` to be left, and `+Z` to be up.
-You can use [the right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule) to determine the orientation of these axes.
-
-{{% /alert %}}
-
```
Otterverse, line 95: I think this still need to be defined somewhere. Removing the definition just makes it MORE ambiguous. Frames themselves are entirely neutral, but we should have a standard for how to attach them to physical objects so that they line up consistent with things humans would consider "forward" or "left" or "positive spin". This was the real issue in my original complaint. We have to say (somewhere, or even multiple places for different things) HOW a frame attaches to a physical object with human-like assumptions about intersectionality, otherwise our components are no longer interchangeable.

tl;dr I believe it should be absolutely clear, to a viam novice reading docs alone, how orientation/direction maps to a component and human assumptions (even if that varies per type/model.)

docs/components/base/_index.md
JessamyT: Ooh you're so right. Would it be positive = turning left?
```suggestion
Positive implies turning to the left.
```

JessamyT: This wording came from the Python docs (so those also need to be changed when we figure this out--I'll follow up), and the Go docs don't specify either.

JessamyT: I *think* this is a safe assumption because https://github.com/viamrobotics/rdk/blob/main/components/base/wheeled/wheeled_base.go#L254 spin tells the left motors to go negative RPM

docs/components/base/_index.md
JessamyT:
```suggestion
Positive implies turning to the left.
```
Ahhh I see you could spin around either way you want to to reach a given position.

docs/components/base/_index.md
```diff
@@ -176,9 +176,9 @@ await my_base.spin(angle=10, velocity=1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `angleDeg` [(float64)](https://pkg.go.dev/builtin#float64): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies forwards.
```
JessamyT, line 179:
```suggestion
Positive implies a given number of degrees to the left.
```

docs/components/base/_index.md
```diff
@@ -176,9 +176,9 @@ await my_base.spin(angle=10, velocity=1)
...
 - `degsPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies forwards.
```
JessamyT, line 181:
```suggestion
Positive implies turning to the left.
```

docs/services/frame-system/_index.md
JessamyT: Hmm. The note as-is is wrong though since Viam has no such consistent coordinate system as far as I understand. As you noted, bases have y forward, but then I think we have grippers with x "forward" (depending on what you even mean by forward :D )
Whereas with differential drive bases that can only move efficiently along one axis, arms and grippers and things have no such inherent axis.

docs/components/base/_index.md
```diff
@@ -91,9 +91,9 @@ Move the base in a straight line across the given distance (mm) at the given vel
 **Parameters:**
 
 - `distance` [(int)](https://docs.python.org/3/library/functions.html#int): The distance to move in millimeters.
-Negative implies backwards.
+Positive implies forwards.
```
npentrel, line 94: Why don't we have both for additional clarity?

docs/services/frame-system/_index.md
npentrel: I think we can potentially solve that by adding specificity to the alert and explaining how it works for different components. I think @Otterverse's point is valid that this does help understanding. Can you work out how it works for each involved resource and potentially also check with the SME's whether we can make it consistent if it is not?

npentrel (changes requested):

docs/services/frame-system/_index.md
JessamyT: What I did in this PR was based on conversations yesterday with Peter. For bases, Y is forwards, but for everything else, there is no inherent forwards. Should we say that?

docs/components/base/_index.md
JessamyT: Okay, can do. Though at a quick glance, with it as-is, they'll read "positive...forwards" instead of seeing all the words "positive, forwards, negative, backwards" and needing to actually read the words carefully, so idk if it helps?

docs/services/frame-system/_index.md
```diff
@@ -86,13 +86,6 @@ The `Orientation` parameter offers `Types` for ease of configuration, but the Fr
 
 {{% /alert %}}
```
JessamyT, line 88:
```suggestion

{{% alert title="Tip" color="tip" %}}

For [base components](../../../components/base/), Viam considers `+X` to be to the right, `+Y` to be forwards, and `+Z` to be up.
You can use [the right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule) to understand rotation about any of these axes.

For non-base components, there is no inherent concept of "forward," so it is up to the user to define frames that make sense in their application.

{{% /alert %}}
```

docs/services/frame-system/_index.md
```diff
@@ -88,8 +88,10 @@ The `Orientation` parameter offers `Types` for ease of configuration, but the Fr
 
 {{% alert title="Tip" color="tip" %}}
 
-Viam's coordinate system considers `+X` to be forward, `+Y` to be left, and `+Z` to be up.
-You can use [the right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule) to determine the orientation of these axes.
+For [base components](../../../components/base/), Viam considers `+X` to be to the right, `+Y` to be forwards, and `+Z` to be up.
```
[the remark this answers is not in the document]
Otterverse: I'm fine with that.

Otterverse (approved):
LGTM

docs/components/base/_index.md
JessamyT: (done though)

docs/components/base/_index.md
```diff
@@ -152,9 +156,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
 **Parameters:**
 
 - `angle` [(float)](https://docs.python.org/3/library/functions.html#float): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies a given number of degrees to the left.
```
JessamyT, line 159:
```suggestion
Positive implies turning to the left.
```

docs/components/base/_index.md
```diff
@@ -152,9 +156,9 @@ Turn the base in place, rotating it to the given angle (degrees) at the given an
...
 - `velocity` [(float)](https://docs.python.org/3/library/functions.html#float): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies turning to the left.
```
JessamyT, line 161:
```suggestion
Given a positive angle and a positive velocity, the base turns to the left (for built-in base models).
```

docs/components/base/_index.md
```diff
@@ -176,9 +180,9 @@ await my_base.spin(angle=10, velocity=1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `angleDeg` [(float64)](https://pkg.go.dev/builtin#float64): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies a given number of degrees to the left.
 - `degsPerSec` [(float64)](https://pkg.go.dev/builtin#float64): The angular velocity at which to spin in degrees per second.
-Negative implies backwards.
+Positive implies turning to the left.
```
JessamyT, line 185:
```suggestion
Given a positive angle and a positive velocity, the base turns to the left (for built-in base models).
```

docs/components/base/_index.md
```diff
@@ -176,9 +180,9 @@ await my_base.spin(angle=10, velocity=1)
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
 - `angleDeg` [(float64)](https://pkg.go.dev/builtin#float64): The angle to spin in degrees.
-Negative implies backwards.
+Positive implies a given number of degrees to the left.
```
JessamyT, line 183:
```suggestion
Positive implies turning to the left.
```

docs/components/base/_index.md
```diff
@@ -209,11 +213,12 @@ Set the linear and angular power of the base, represented as a percentage of max
...
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
   Viam's coordinate system considers +Y to be the forward axis (+/- X left/right, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
-  Negative "Y:" values imply moving backwards.
+  Positive "Y" values imply moving forwards.
+  Negative "Y" values imply moving backwards.
 - `angular` [(Vector3)](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.Vector3): The percentage of max power of the base's angular propulsion.
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
   Use the Z component of this vector to spin left or right when controlling a wheeled base.
-  Negative "Z:" values imply spinning to the right.
+  Positive "Z" values imply spinning to the left.
```
JessamyT, line 221:
```suggestion
  Positive "Z" values imply spinning to the left (for built-in base models).
```

docs/components/base/_index.md
```diff
@@ -247,12 +252,15 @@ await my_base.set_power(linear=Vector3(x=0,y=0,z=0), angular=Vector3(x=0,y=0,z=-
...
+  In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
 Viam's coordinate system considers +Y to be the forward axis (+/- X left/right, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
-Negative "Y:" values imply moving backwards.
+Positive "Y" values imply moving forwards.
+Negative "Y" values imply moving backwards.
 - `angular` [(r3.Vector)](https://pkg.go.dev/github.com/golang/geo/r3#Vector): The percentage of max power of the base's angular propulsion.
 In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
-Use the Z component of this vector to spin left or right when controlling a wheeled base. Negative "Z:" values imply spinning to the right.
+Use the Z component of this vector to spin left or right when controlling a wheeled base.
+Positive "Z" values imply spinning to the left.
```
JessamyT, line 263:
```suggestion
Positive "Z" values imply spinning to the left (for built-in base models).
```

docs/components/base/_index.md
```diff
@@ -208,12 +212,13 @@ Set the linear and angular power of the base, represented as a percentage of max
 
 - `linear` [(Vector3)](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.Vector3): The percentage of max power of the base's linear propulsion.
   In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
-  Viam's coordinate system considers +Y to be the forward axis (+/- X right/left, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
-  Negative "Y:" values imply moving backwards.
+  Viam's coordinate system considers +Y to be the forward axis (+/- X left/right, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
```
JessamyT, line 215:
```suggestion
  Viam's coordinate system considers +Y to be the forward axis (+/- X right/left, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
```

docs/components/base/_index.md
```diff
@@ -247,12 +252,15 @@ await my_base.set_power(linear=Vector3(x=0,y=0,z=0), angular=Vector3(x=0,y=0,z=-
 **Parameters:**
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
-- `linear` [(r3.Vector)](https://pkg.go.dev/github.com/golang/geo/r3#Vector): The percentage of max power of the base's linear propulsion. In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
+- `linear` [(r3.Vector)](https://pkg.go.dev/github.com/golang/geo/r3#Vector): The percentage of max power of the base's linear propulsion.
+  In the range of -1.0 to 1.0, with 1.0 meaning 100% power.
 Viam's coordinate system considers +Y to be the forward axis (+/- X left/right, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
```
JessamyT, line 257:
```suggestion
Viam's coordinate system considers +Y to be the forward axis (+/- X right/left, +/- Z up/down), so use the Y component of this vector to move forward and backward when controlling a wheeled base.
```

npentrel (approved):

docs/services/frame-system/_index.md
JessamyT, line 91:
```suggestion
For [base components](../../components/base/), Viam considers `+X` to be to the right, `+Y` to be forwards, and `+Z` to be up.
```

PR #1402, opened 2023-06-16
Fix motion planning tutorial table dimensions

JessamyT:
Make consistent between the Python and Go tabs, and with the description.

### DIFF
--- a/docs/tutorials/services/plan-motion-with-arm-gripper.md
+++ b/docs/tutorials/services/plan-motion-with-arm-gripper.md
@@ -203,7 +203,7 @@ worldState := &referenceframe.WorldState{
 
 This example adds a "table" with the assumption that you mounted your robot arm to an elevated surface.
 The 2000 millimeter by 2000 millimeter dimensions ensure that a sufficiently large box is constructed, regardless of the real physical footprint of your mounting surface.
-The origin at -19 mm along the Z axis (half the table's thickness) positions the table so that its top surface is at 0.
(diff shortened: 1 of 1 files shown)


docs/tutorials/services/plan-motion-with-arm-gripper.md
```diff
@@ -203,6 +203,7 @@ worldState := &referenceframe.WorldState{
 
 This example adds a "table" with the assumption that you mounted your robot arm to an elevated surface.
 The 2000 millimeter by 2000 millimeter dimensions ensure that a sufficiently large box is constructed, regardless of the real physical footprint of your mounting surface.
+The origin at -19 mm along the Z axis (half the table's thickness) positions the table so that its top surface is at 0.
```
andf-viam, line 206:
```suggestion
The -19 mm origin along the Z axis (which is computed as half the table's thickness of 38mm) positions the table so that its top surface is at the 0 position.
```

To match the previous sentence's structure, if desired, with some optional language ideas.

andf-viam (approved):
LGTM, thanks for catching! Left an optional language tweak if you agree.

docs/tutorials/services/plan-motion-with-arm-gripper.md
JessamyT: I think that makes less sense since that's only one component of the origin, but made me think about it more and I think this is better:
```suggestion
Setting the Z component of the origin to -19 mm (half the table's thickness) conveniently positions the top surface of the table at 0.
```

andf-viam: That looks good to me!

PR #1404, opened 2023-06-16
Update foam-dart-launcher.md

Kschappacher:
There was some inaccurate information in the tutorial around some of the wiring. These commits should fix those issues


sguequierre (approved):
Thanks for catching this and making the pr!

PR #1408, opened 2023-06-20
DOCS-748: Document the ur5e arm

sguequierre:
* Document example configuration for an ur5e arm
* Update the main arm page to include this model 
* Fix speed/accel attribute names on xArms (not sure when this became inaccurate) 
* Update the weights of the arm models so they display well in toolbar


docs/components/arm/xarmlite.md
```diff
@@ -83,7 +83,7 @@ The following attributes are available for `xArmLite` arms:
 | --------- | ---- | ----------| ----------- |
 | `host`  | string | **Required** | IP address of the arm's system on your network. Find this when setting up your xArm. |
 | `port`  | int | Optional | Port number of the arm's system. Find this when setting up your xArm. <br> Default: `502` |
-| `speed` | float | Optional | Desired maximum speed of joint movement in degrees/sec. <br> Default: `20.0` |
-| `acceleration`  | float | Optional | Desired maximum acceleration of joint movement in degrees/sec<sup>2</sup>. <br> Default: `50.0` |
+=| `speed_degs_per_sec` | float | Optional | Desired maximum speed of joint movement in degrees/sec. <br> Default: `20.0` |
```
sguequierre, line 86:
```suggestion
| `speed_degs_per_sec` | float | Optional | Desired maximum speed of joint movement in degrees/sec. <br> Default: `20.0` |
```

docs/components/arm/ur5e.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+The following attributes are available for `ur5e` arms:
+
+| Attribute | Type | Inclusion | Description |
+| --------- | ---- | ----------| ----------- |
+| `speed_degs_per_sec`  | string | **Required** | Desired maximum speed of joint movement in degrees/sec. <br> Range: `[.1, 1]` |
```
sguequierre, line 79:
```suggestion
| `speed_degs_per_sec`  | float | **Required** | Desired maximum speed of joint movement in degrees/sec. <br> Range: `[.1, 1]` |
```

docs/components/arm/img/ur5e-ui-config.png
JessamyT, line 1: maybe put a number between .1 and 1 in the speed since 0 would be invalid

docs/components/arm/ur5e.md
```diff
@@ -0,0 +1,80 @@
...
+
+{{% /tab %}}
+{{< /tabs >}}
+
+The following attributes are available for `ur5e` arms:
+
+| Attribute | Type | Inclusion | Description |
+| --------- | ---- | ----------| ----------- |
+| `speed_degs_per_sec`  | float | **Required** | Desired maximum speed of joint movement in degrees/sec. <br> Range: `[.1, 1]` |
+| `host`  | string | **Required** | The IP address of the arm's system on your network. Find this when setting up your ur5e. |
```
JessamyT, line 80:
```suggestion
| `host`  | string | **Required** | The IP address of the arm's system on your network. Find this when setting up your UR5e. |
```

JessamyT (commented):
Looks good except for two nits

docs/components/arm/img/ur5e-ui-config.png
sguequierre: It's not in the validate function so wouldn't return an error should be ok!

JessamyT: I'm confused--it wouldn't function properly, and the table below says this is a required attribute with a range of .1 to 1?

JessamyT: Which validate function? I'm looking at this one that looks like it would indeed throw an error. https://github.com/viamrobotics/rdk/blob/5c7649e6e51a52f5f7dc36f7547ec4e9a18ff056/components/arm/universalrobots/ur.go#L45 In either case, seems it makes sense to show a valid rather than an invalid example there.

sguequierre: @JessamyT Ah, you're right-- was confused, changed!

sguequierre: I do think that if it autopopulates like `0` and we're telling them to fill out the attributes in instructions later on it shouldn't be too much of a blocker on pages like these, almost seems cleaner

JessamyT: Hmm I mean it's not the biggest deal but on all the other arm models it looks like we have valid config builder examples in the screenshots (exception being Yahboom which has example strings). We should probably all align these at some point.

JessamyT (approved):

PR #1409, opened 2023-06-21
DOCS-461: Create motion planning with constraints tutorial

JessamyT:


npentrel:
It may be worth getting this merged & reviewed without the Go code for now

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Define start and end poses
+
+You need to tell the robot where to pick up the cup, and where to put it down.
+
+While the Motion Service has the mathematical ability to plan complex motion from one position to another, around an obstacle, this is computationally intensive.
+The program will run much more efficiently if you add waypoints to more tightly define the motion.
+The way points in the code below are defined such that the path between each consecutive point can be linear, without intersecting the tissue box (or other obstacle).
+
+Though we are passing these waypoints in manually, the Motion Service will throw an error if we accidentally pass in a pose that would cause the arm or gripper to hit any obstacles.
```
andf-viam, line 235: I would bump this sentence to just after the code sample, leaving `in the code below` to immediately precede the code, and `though we are <doing a thing>` to just after we `<did that thing>`.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+```
+
+## Add a motion constraint
+
+To keep the cup upright as the arm moves it from one place on the table to another, create a [linear constraint](../../../services/motion/constraints/#linear-constraint).
+Then, when you tell the robot to move the cup from one upright position to another, the gripper will move linearly and the upright orientation of the cup will be maintained throughout the planned path.
+
+You could try using an [orientation constraint](../../../services/motion/constraints/#orientation-constraint) instead, which would also constrain the orientation.
+However, since this opens up many more options for potential paths, it is much more computationally intensive than the linear constraint.
+
```
andf-viam, line 261: Since we just described the two possible constraint types we could use, let's clarify explicitly which one we'll be using, esp since most recently-described option is not the one we're using. Maybe:

>The code below creates a linear constraint, which ensures that the robot moves the cup along the predetermined route while avoiding the obstacles we've defined and while keeping the cup upright:

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+print(f"At end pose")
+
+# Put down the cup
+await my_gripper.open()
+```
+
+## Full tutorial code
+
+The following code contains everything covered in this tutorial in addition to the connect function, and the resource access code from the last tutorial that you still need here.
+Be sure to change the robot secret and address values to match your actual robot credentials, and change all relevant parameters such as `z_offset` and other dimensions and poses to match your hardware.
```
andf-viam, line 283:
```suggestion
Be sure to change the `<ROBOT SECRET PAYLOAD>` and `<ROBOT ADDRESS>` placeholders shown in the code to match your actual robot credentials, and change all relevant parameters such as `z_offset` and other dimensions and poses to match your hardware.
```

Or similar, just to help the user identify what's being replaced. Maybe no brackets?

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+# Create a WorldState that includes the table and tissue box obstacles, and the cup transform
+world_state = WorldState(obstacles=[obstacles_in_frame], transforms = transforms)
+```
+
+## Define start and end poses
+
+You need to tell the robot where to pick up the cup, and where to put it down.
+
+While the Motion Service has the mathematical ability to plan complex motion from one position to another, around an obstacle, this is computationally intensive.
+The program will run much more efficiently if you add waypoints to more tightly define the motion.
```
andf-viam, line 232: nit: I was expecting something like the following to more directly answer "well, what can I do to make my motion planning less computationally intensive?":

```suggestion
The program will run much more efficiently if you add waypoints to more tightly define the motion, which allows the math engine to work less hard on mathing.
```

Haha, or maybe more likely:

```suggestion
The program will run much more efficiently if you add waypoints to more tightly define the motion, which allows the Motion planning service to make use of simple pathing algorithms for each movement stage rather than computing paths for each stage individually each time.
```

Obviously, I don't know the answer here, just offering an opportunity for following a likely question with a likely answer.

andf-viam (commented):
This is awesome! I've left some language and layout suggestions for your consideration - ping me if you want to chat about anything!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+Be careful when instructing robot arms to move.
+Before running any code, ensure your robotic arm has enough space and that there are no obstacles.
+Also pay attention to your surroundings, double-check your code for correctness, and make sure anyone nearby is aware and alert before issuing commands to your robot.
+{{< /alert >}}
+
+{{<gif webm_src="../../videos/motion_constraints.webm" mp4_src="../../videos/motion_constraints.mp4" poster="../../videos/motion_constraints.jpg" alt="An arm moving a cup from one side of a tissue box to the other, across a table. The cup stays upright." class="alignright" max-width="250px">}}
+
+Say you want your robot to pass you a cup of tea, but you don't want it to spill the water or bump into other objects on the table.
+
+If you followed along in [part 2 of the Motion Service tutorial series](../plan-motion-with-arm-gripper/), you used the [Motion Service](/services/motion/) to move a robot arm and end effector to desired positions.
```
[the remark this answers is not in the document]
JessamyT: Yeah...it's mentioned in the prereqs of each of them, and we do have similar situations with at least two other sets of tutorials. You're right it's worth a rethink. I didn't write the first two and didn't think this PR should involve refactoring but I'll make a ticket.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+If you are connecting to a real robotic arm during this tutorial, make sure your computer can communicate with the arm controller before continuing.
+
+Familiarize yourself with the concepts outlined in the second motion tutorial, [Plan Motion with an Arm and a Gripper](../plan-motion-with-arm-gripper/), before continuing.
+This tutorial picks up where **Plan Motion with an Arm and a Gripper** stops, so further examples depend on having a connected robot, client and service access, and other infrastructure in place.
+
+For reference, look at [the full code sample from the prior tutorial](../plan-motion-with-arm-gripper/#full-tutorial-code).
+
+## Configure your robot
+
+Use the robot configuration from [the previous tutorial](../plan-motion-with-arm-gripper/) for this tutorial, including the [arm](../../../components/arm/) and [gripper](../../../components/gripper/) components with [frames](../../../services/frame-system/) configured.
```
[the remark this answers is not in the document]
JessamyT: oof I wish GH didn't render that as a very long scroll code block haha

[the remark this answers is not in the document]
JessamyT: And yeah....that's why I included the whole config as an expander so they could just use that. I don't want to repeat all the steps since they should've already configured everything they need in the previous two tutorials. I think this is part of the overall "how do we deal with tutorial series" question.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+
+The Motion Service is one of the "built-in" services, so you don't need to do anything to enable it on your robot.
+
+{{% expand "Click for an example raw JSON config." %}}
```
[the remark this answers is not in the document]
JessamyT: Yes, it is, but why would you make a new robot?

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+
```
JessamyT, line 54:
```suggestion

If you go to your robot's **Config** tab on [the Viam app](https://app.viam.com/) and click **Raw JSON**, your config should resemble the following:

```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+        reference_frame="myGripper",
+        pose=Pose(x=0, y=0, z=45, o_x=0, o_y=-1, o_z=0, theta=0)),
+        physical_object=cup_geometry)
+]
+```
+
+The 155mm offset along the gripper's Z axis accounts for the radius of the gripper (45mm), plus the distance from the gripper's origin (where it meets the arm) and what you can think of as the palm of its hand.
+In this example, that distance is 110mm (see diagram above), but if you have a different gripper, change the offset accordingly.
+
+Now that you have created the table and tissue box obstacles and the cup transform, create a WorldState that includes all of them:
```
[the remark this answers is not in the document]
JessamyT: If this tutorial gets language agnostic though?

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+```python {class="line-numbers linkable-line-numbers"}
+# Create a WorldState that includes the table and tissue box obstacles, and the cup transform
+world_state = WorldState(obstacles=[obstacles_in_frame], transforms = transforms)
+```
+
+## Define start and end poses
+
+You need to tell the robot where to pick up the cup, and where to put it down.
+
+While the Motion Service has the mathematical ability to plan complex motion from one position to another, around an obstacle, this is computationally intensive.
```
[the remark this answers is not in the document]
JessamyT: I think your suggestion opens a possible interpretation along the lines of: "The Motion Service is generally computationally intensive, whether it's maneuvering around an obstacle or not." I will suggest a rewording of this paragraph to avoid this as well as your nit :)

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+The program will run much more efficiently if you add waypoints to more tightly define the motion.
+The way points in the code below are defined such that the path between each consecutive point can be linear, without intersecting the tissue box (or other obstacle).
```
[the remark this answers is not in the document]
JessamyT: It will avoid intersecting all obstacles. Perhaps that is not clear? I will reword.

docs/tutorials/services/constrain-motion.md
JessamyT: Hmm but maybe then they won't realize that the Motion Service is in fact preventing collisions, as opposed to thinking that they're doing all the work for it.

docs/tutorials/services/constrain-motion.md
JessamyT: I didn't want to make this super verbose and redundant, and I think the line-by-line comments can be more helpful since they say what each line is doing.
What are your thoughts on body text versus comments in the code?

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+## Add a motion constraint
+
+To keep the cup upright as the arm moves it from one place on the table to another, create a [linear constraint](../../../services/motion/constraints/#linear-constraint).
+Then, when you tell the robot to move the cup from one upright position to another, the gripper will move linearly and the upright orientation of the cup will be maintained throughout the planned path.
+
+You could try using an [orientation constraint](../../../services/motion/constraints/#orientation-constraint) instead, which would also constrain the orientation.
+However, since this opens up many more options for potential paths, it is much more computationally intensive than the linear constraint.
+
+```python {class="line-numbers linkable-line-numbers"}
+# Create a constraint to keep the cup upright as it moves
```
JessamyT, line 263:
```suggestion
# Create a linear constraint to keep the cup upright as it moves in a linear path
```

docs/tutorials/services/constrain-motion.md
JessamyT: There are brackets on the previous tutorials--idk

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+obstacles_in_frame = GeometriesInFrame(reference_frame="world",
+    geometries=[table_object, box_object])
+```
+
+## Determine the gripper's axes
+
+You need to figure out how the gripper's frame corresponds to its hardware.
+That is, you need to know which axis points from the "wrist" to the end of the gripper, which axis lines up with the direction the jaws actuate, and if the gripper was holding a cup, which axis would pass vertically through the cup.
+
+All example code below assumes +Z points from the base of the gripper to the point where its jaws close, +Y points towards the ground when the gripper is holding a cup from the side, and the X axis is the axis along which the jaws close, following the [right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule/).
```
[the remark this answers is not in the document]
JessamyT: Two things:
1. The fake vs real is about grippers rather than arms. They could use a real arm and a fake gripper (which is almost what I did, but with a nonfunctioning gripper stand-in, configured as a fake). Not sure if that changes your feedback or if it was a typo.
2. In either case, this is actually relevant to all users. Even if they're using a fake arm AND a fake gripper, a big piece of understanding and using the frame system and motion service is thinking about how axes align with objects. Even though their fake gripper is just an arbitrary rectangular box, it's important to think about how it would work if it was a real gripper. If their gripper is fake, they can assign those axes however they want, but I'd hope they'd at least be imagining things like "okay, this represents this, so I need to move it this way" as a learning exercise.

Given this, do you think it'd be good if I changed the fake paragraph to "...there is no real hardware to calibrate and you can [continue this tutorial], imagining that your fake gripper corresponds to the diagram above."?

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,410 @@
+---
+title: "Add Constraints and Transforms to a Motion Plan"
+linkTitle: "Add Motion Constraints"
+weight: 25
+type: "docs"
+description: "Use constraints and transforms with the Motion Service to move robot components in specific ways."
```
npentrel, line 6:
```suggestion
description: "Use constraints and transforms with the Motion Service to move robot components."
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,410 @@
...
+webmSrc: "/tutorials/videos/motion_constraints.webm"
+mp4Src: "/tutorials/videos/motion_constraints.mp4"
+images: ["/tutorials/videos/motion_constraints.jpg"]
```
npentrel, line 9: This would be nicer as the gif of the video.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,410 @@
+---
+title: "Add Constraints and Transforms to a Motion Plan"
+linkTitle: "Add Motion Constraints"
+weight: 25
+type: "docs"
+description: "Use constraints and transforms with the Motion Service to move robot components in specific ways."
+webmSrc: "/tutorials/videos/motion_constraints.webm"
+mp4Src: "/tutorials/videos/motion_constraints.mp4"
```
npentrel, line 8: The mp4 file is too large - also, you need to either retake this footage or crop it to fit a 4:3 aspect ratio. I know not all of them currently exactly match that spec but they are all in landscape mode.

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
npentrel: You can solve this by explicitly suggesting they add this JSON config to their config and then adding the info that this is from a previous tutorial after that

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,410 @@
...
+        }
+      }
+    }
+  ]
+}
+```
+
+{{% /expand %}}
+
+<!--
```
npentrel, line 131: what's happening with this comment?

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: Ah ha! I am indeed unclear on what to do with this JSON! This is why I think a brief intro is valuable here. When you state directly why this conf file is here, I can easily determine:

- Oh I have a robot already from last tutorial, I don't need this (or can compare to make sure mine is still the same)
- Wait, I totally just Googled to get here and didn't even know that there were previous tutorials in the series. Halp! I need a conf file in its entirety to get up to speed! I do need this! And: I probably will be creating a new robot, no?

docs/tutorials/services/constrain-motion.md
andf-viam: Definitely prefer explicit mention in-copy rather than relying on code comments to carry the teaching burden. Copy in code fences is not searchable in TypeSense (our search) or via external search engine's site crawlers (I think! Correct me if I'm wrong!).

Your code comments are great! But not everyone reads code comments, or reads them carefully enough to rely on them to convey core concepts. Certainly retain the code comments!! But a single sentence intro summarizing the goal and result of the 14 lines of code here helps put the code comments in context, presents conceptual content inline in the tutorial flow (caught by readers who skip code, and search engines alike!), and answers the "why" of why do I want to use this code: `which ensures that the robot moves the cup along the predetermined route while avoiding the obstacles we've defined and while keeping the cup upright` (or however you would like to choose to phrase such a summary!).

Certainly an additional paragraph would be too much, given that your line-by-line code comments cover the details well. But I think a single introductory sentence helps here, more than it risks undue verbosity.

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: 1. Yes `("ohhh my arm actually grips along the Z axis", etc)` should have been `("ohhh my gripper actually grips along the Z axis", etc)`
2. I agree that that level of thinking is good here! I wasn't advising any deletions, just rearrangement, but you've convinced me that this is relevant to all users, and should remain where it is. I withdraw my suggestion! 👍

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: Ah, I had thought we needed to define each obstacle in order for the hardware to know it exists, so that it can then avoid it. Is it able to avoid arbitrary objects -- like a toy dinosaur I place on the table -- without me defining its `dino_dims` and `dino_origin` ? If not, the phrasing `(or other obstacle)` lightly suggests that I can expect the arm to avoid _any_ other obstacle.

docs/tutorials/services/constrain-motion.md
JessamyT: If we're taking that off I think I'll take off "to move robot components" since the point there was to say why you are using constraints and transforms, but you don't need constraints and transforms to move robot components in unspecific ways :)

docs/tutorials/services/constrain-motion.md
JessamyT: Oh I thought we're supposed to also have this as a "poster" for previews? What is a poster?

docs/tutorials/services/constrain-motion.md
JessamyT: @andf-viam I did add this based on your first round of feedback--not enough?

docs/tutorials/services/constrain-motion.md
JessamyT: I heard that the visualizer stuff would significantly improve soon, and it isn't currently in a showcase-ready state.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+-->
+
+## Add an obstacle on the table
+
+In the last tutorial, you [added a table obstacle](../plan-motion-with-arm-gripper/#describe-the-robots-working-environment) to prevent the arm and gripper from hitting the table upon which the arm is mounted.
+In this tutorial, keep that table code and add an additional obstacle: a tissue box sitting on the table.
+
+We also added a `z_offset` parameter to the code so you can more easily calibrate your motion plans based on how high or low your arm is mounted compared to the table.
+
+You can adjust the dimensions and positions of the obstacles to describe your own scenario, but here is an example:
```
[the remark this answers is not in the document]
JessamyT: RE "we," it was used 12 times in the previous tutorial so I went for it here. I usually avoid it but given that that's in keeping with the rest of the series I didn't. Also, I'm unclear on our voice for tutorials now that we have such a range. So, not sure.

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
JessamyT: Okay! Still, now that you've called more attention to this I think I want to add in the "imagining that your fake gripper corresponds to the diagram above." to tie it together a bit!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+![A gripper mounted on an arm. The Z axis of the gripper points from the base of the gripper to the end of its jaws. The X axis points up through the gripper. The Y axis points in the direction along which the jaws open and close (following right-hand rule). The diagram also shows the global coordinate system with Z pointing up, X down the length of the horizontal gripper, and Y pointing horizontally in the opposite direction of the gripper's Y.](../../img/constrain-motion/gripper-diagram.png)
+
+If you are using a `fake` gripper, there is no real hardware to calibrate and you can [continue this tutorial](#use-a-transform-to-represent-a-drinking-cup).
+
+If you are using a real arm and gripper, use the **Control** tab in the [Viam app](https://app.viam.com/) to move the gripper, look at its reported orientations, and map them to its orientation in the real world.
+If the axes are different from those described above, take these differences into account in your code.
+
+## Use a transform to represent a drinking cup
+
+Imagine your cup is 120 millimeters tall and 90 millimeters in diameter.
```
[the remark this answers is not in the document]
JessamyT: ooh yeah totally

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
JessamyT: But in any case, I like your suggestion and am incorporating it! "We" still appears later on though--do you think it should come out?

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
JessamyT: Ohhh sorry I misunderstood your original comment. Yes, it's true, the only obstacles we have defined are the box and table. I guess I wanted it to be general in case someone defined a birthday cake and a present instead of a tissue box, or something. But you're right that's not so clear. FWIW the arm will also avoid hitting itself or other components, but that's different from "obstacles" as such, yes.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+{{< /alert >}}
+
+Before starting this tutorial, make sure you have the [Viam Python SDK](https://python.viam.dev/) <!-- or the [Viam Go SDK](https://pkg.go.dev/go.viam.com/rdk/robot/client#section-readme/) -->installed.
+
+If you are connecting to a real robotic arm during this tutorial, make sure your computer can communicate with the arm controller before continuing.
+
+Familiarize yourself with the concepts outlined in the second motion tutorial, [Plan Motion with an Arm and a Gripper](../plan-motion-with-arm-gripper/), before continuing.
+This tutorial picks up where **Plan Motion with an Arm and a Gripper** stops, so further examples depend on having a connected robot, client and service access, and other infrastructure in place.
+
+For reference, look at [the full code sample from the prior tutorial](../plan-motion-with-arm-gripper/#full-tutorial-code).
```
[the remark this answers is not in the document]
JessamyT: Ooh nice I like this a lot

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,408 @@
...
+The following diagram shows this, as well as the global coordinate system.
+
+![A gripper mounted on an arm. The Z axis of the gripper points from the base of the gripper to the end of its jaws. The X axis points up through the gripper. The Y axis points in the direction along which the jaws open and close (following right-hand rule). The diagram also shows the global coordinate system with Z pointing up, X down the length of the horizontal gripper, and Y pointing horizontally in the opposite direction of the gripper's Y.](../../img/constrain-motion/gripper-diagram.png)
+
+If you are using a `fake` gripper, there is no real hardware to calibrate and you can [continue this tutorial](#use-a-transform-to-represent-a-drinking-cup).
+
+If you are using a real arm and gripper, use the **Control** tab in the [Viam app](https://app.viam.com/) to move the gripper, look at its reported orientations, and map them to its orientation in the real world.
+If the axes are different from those described above, take these differences into account in your code.
```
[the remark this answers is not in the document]
JessamyT: Sounds like this can be resolved based on the other thread?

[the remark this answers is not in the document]
andf-viam: Sure, clear to resolve!

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: Up to you!

docs/tutorials/services/constrain-motion.md
andf-viam: Sure, this works for me, thank you!!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+obstacles_in_frame = GeometriesInFrame(reference_frame="world",
+    geometries=[table_object, box_object])
+```
+
+## Determine the gripper's axes
+
+You need to figure out how the gripper's frame corresponds to its hardware.
+That is, you need to determine which axis points from the "wrist" to the end of the gripper, which axis lines up with the direction the jaws actuate, and, when the gripper is holding a cup, which axis passes vertically through the cup.
+
+All example code below assumes +Z points from the base of the gripper to the point where its jaws close, +Y points towards the ground when the gripper is holding a cup from the side, and the X axis is the axis along which the jaws close, following the [right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule/).
```
JessamyT, line 189:
```suggestion
All example code below assumes +Z points from the base of the gripper to the point where its jaws close, +Y points towards the ground when the gripper is holding a cup from the side, and the X axis is the axis along which the jaws close, following the [right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule).
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+{{< alert title="Tip" color="tip" >}}
+You may be wondering how the orientations of the poses are determined.
+Our example gripper's frame is defined such that its orientation vector points from its "wrist" to the tip of its jaws.
+In the example code above, all poses have an [orientation vector](../../../internals/orientation-vector/) pointing along the positive X axis of the world frame, which is a horizontal orientation pointing "forwards" with respect to the xArm 6 base.
+When we tell the gripper to move to such a pose, its orientation vector moves to align with the orientation vector of the pose, so its jaws end up pointing along the global X axis, "forwards" from the robot base.
+This puts it in a good position for picking up and moving a cup.
+
+Additionally, for our gripper, setting `theta=0` about this particular orientation vector orients the gripper such that its jaws open and close horizontally.
+If we changed it to `theta=90` or `theta=270`, the gripper jaws would open vertically, not ideal for picking up a cup!
+{{< /alert >}}
```
andf-viam, line 274: This tip is a bit lengthy, and conveys some info that I think would be helpful to all users, not just curious tip-reading ones (namely the `orientation-vector` link). What do you think about including its most important bits inline, maybe something like (replacing lines 262-274):

> Note that all poses use the same [orientation vector](../../../internals/orientation-vector/), pointing along the positive X axis of the world frame, which is a horizontal orientation pointing "forwards" with respect to the xArm 6 base.
> When we tell the gripper to move to each pose, its orientation vector moves to align with the orientation vector of the pose, so that its jaws end up pointing along the global X axis, "forwards" from the robot base.
> We also set `theta=0` for each orientation vector to orient the gripper such that its jaws open and close horizontally.
> Together, these configurations put the gripper in a good position for picking up and moving the cup. 

I removed the reference to spilling the tea because we discuss that in the next section, but I could see it retained here also if you prefer.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+
+## Add a motion constraint
+
+To keep the cup upright as the arm moves it from one place on the table to another, create a [linear constraint](../../../services/motion/constraints/#linear-constraint).
+Then, when you tell the robot to move the cup from one upright position to another, the gripper will move linearly and the upright orientation of the cup will be maintained throughout the planned path.
```
andf-viam, line 279:
```suggestion
A linear constraint ensures that the gripper will move linearly when you tell the robot to move the cup from one upright position to another, and that it will maintain the upright orientation of the cup throughout the planned path.
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+way1_pose_in_frame = PoseInFrame(reference_frame="world", pose=way1_pose)
+way2_pose = Pose(x=300, y=-240, z=320+z_offset, o_x=1, o_y=0, o_z=0, theta=0)
+way2_pose_in_frame = PoseInFrame(reference_frame="world", pose=way2_pose)
+
+# Create a pose where the cup will be set down
+end_pose = Pose(x=300, y=-250, z=90.0+z_offset, o_x=1, o_y=0, o_z=0, theta=0)
+end_pose_in_frame = PoseInFrame(reference_frame="world", pose=end_pose)
+```
+
+Though we are passing these waypoints in manually, the Motion Service will throw an error if we accidentally pass in a pose that would cause the arm or gripper to hit any obstacles.
```
andf-viam, line 260:
```suggestion
Though we are passing these waypoints in manually, the Motion Service will throw an error if we accidentally pass in a pose that would cause the arm or gripper to hit any obstacles we've defined.
```

To match line 284.

andf-viam (approved):
LGTM % small language nits and some optional suggestions. Great work!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+- If you are connecting to a real robotic arm during this tutorial, make sure your computer can communicate with the arm controller before continuing.
+  Code examples in this tutorial use a [UFACTORY xArm 6](https://www.ufactory.cc/product-page/ufactory-xarm-6/), but you can use any [arm model](/components/arm/) including a [fake arm model](../../../components/arm/fake/).
+- Complete the previous tutorial, [Plan Motion with an Arm and a Gripper](../plan-motion-with-arm-gripper/), which configures the robot, client and service access, and other infrastructure we'll need for this tutorial.
+  For reference, see the [full code sample from the prior tutorial](../plan-motion-with-arm-gripper/#full-tutorial-code).
+
+## Configure your robot
+
+Use the same robot configuration from [the previous tutorial](../plan-motion-with-arm-gripper/) for this tutorial, including the [arm](../../../components/arm/) and [gripper](../../../components/gripper/) components with [frames](../../../services/frame-system/) configured.
+Make one change: Change the Z translation of the gripper to zero.
```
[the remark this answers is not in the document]
JessamyT: It's changing a 90 to a 0:
```suggestion
Make one change: Change the Z translation of the gripper frame from `90` to `0`.
```

Another option: Change the other tutorial. Currently, it tells people to add an offset but doesn't say why or what it represents. 

Also though, I hear that a change is coming along that will change the frame configuration interface so we'll need to update both this and the other tutorial for that anyway.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+
+The Motion Service is one of the "built-in" services, so you don't need to do anything to enable it on your robot.
+
+{{% expand "Click to see what your raw JSON config should look like." %}}
+
+If you go to your robot's **Config** tab on [the Viam app](https://app.viam.com/) and click **Raw JSON**, your config should resemble the following.
+
+If you're creating a fresh configuration instead of using your robot from the previous tutorials, you can copy paste this into your robot's **Raw JSON** field.
```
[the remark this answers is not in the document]
JessamyT: Yeah, i like that!

JessamyT, line 53:
```suggestion
If you completed the previous tutorial, your robot's configuration should match the following.
You can view your robot configuration in [the Viam app](https://app.viam.com/) under the **Config** tab by clicking **Raw JSON**.

If instead you are creating a new robot for this tutorial, copy the following configuration into the Raw JSON field:
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+Account for this by giving the table a **Translation** of `-10` in the Z direction, moving the top of the table to 0.
+
+![The Frame System subtab of the Components tab with a 1000 wide, 1000 deep, 200 millimeter thick box representing the top of the table.](../../img/constrain-motion/frame-table.jpg)
+
+### Visualize obstacles on the table
+-->
+
+## Modify your robot's working environment
+
+In the last tutorial, you [defined a table obstacle](../plan-motion-with-arm-gripper/#describe-the-robots-working-environment) to prevent the arm and gripper from hitting the table upon which the arm is mounted.
```
JessamyT, line 156:
```suggestion
In the previous tutorial, you [defined a table obstacle](../plan-motion-with-arm-gripper/#describe-the-robots-working-environment) to prevent the arm and gripper from hitting the table upon which the arm is mounted.
```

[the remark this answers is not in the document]
JessamyT:
```suggestion
In the previous tutorial, you [defined a table obstacle](../plan-motion-with-arm-gripper/#describe-the-robots-working-environment) to prevent the arm and gripper from hitting the table upon which the arm is mounted.
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+In this tutorial, you'll expand on the code that describes your robot's working environment in two ways:
+
+- Define an additional obstacle: a tissue box sitting on the table.
+- Add a `z_offset` parameter to the `table_origin` and to the new `box_origin`.
+  This makes it easier to calibrate your motion plans based on how high or low your arm is mounted compared to the table.
+
+The following example shows the two obstacles defined with the expanded configuration.
+Adjust the dimensions and positions of the obstacles to describe your own scenario:
```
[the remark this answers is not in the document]
JessamyT: You wrote this as a suggestion lolol

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+box_object = Geometry(center=box_origin, box=RectangularPrism(dims_mm=box_dims))
+
+obstacles_in_frame = GeometriesInFrame(reference_frame="world",
+    geometries=[table_object, box_object])
+```
+
+## Determine the gripper's axes
+
+You need to figure out how the gripper's frame corresponds to its hardware.
+That is, you need to determine which axis points from the "wrist" to the end of the gripper, which axis lines up with the direction the jaws actuate, and, when the gripper is holding a cup, which axis passes vertically through the cup.
```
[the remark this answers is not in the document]
JessamyT: Well, in the next paragraph I used the language "points" because there it's talking about the positive Z axis specifically. And used the same language here for consistency. Do you think that's okay? Whereas, it's near impossible to explain succinctly which way the other positive axes are "pointing," so I didn't bother.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+box_dims = Vector3(x=120.0, y=80.0, z=100.0)
+box_object = Geometry(center=box_origin, box=RectangularPrism(dims_mm=box_dims))
+
+obstacles_in_frame = GeometriesInFrame(reference_frame="world",
+    geometries=[table_object, box_object])
+```
+
+## Determine the gripper's axes
+
+You need to figure out how the gripper's frame corresponds to its hardware.
```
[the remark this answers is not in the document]
JessamyT: Maybe this to be more specific?
```suggestion
You need to identify how the gripper's frame corresponds to its hardware so that you can write code that creates your desired behavior.
```

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: Agreed that this is definitely a tricky concept to convey. `points` is fine, especially if used consistently. Feel free to keep as is here, it is definitely understandable in present form.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+If you are using a real arm and gripper, use the **Control** tab in the [Viam app](https://app.viam.com/) to move the gripper, look at its reported orientations, and map them to its orientation in the real world.
+If the axes are different from those described above, take these differences into account in your code.
+
+## Use a transform to represent a drinking cup
+
+Imagine your cup is 120 millimeters tall with a radius of 45 millimeters.
+You need to take this space into account to avoid bumping objects on the table with the cup.
+
+You can pass transforms to the [Motion Service `move` method](../../../services/motion/#move) to represent objects that are connected to the robot but are not actual robotic components.
+To represent the drinking cup held in your robot's gripper, add this to your code:
```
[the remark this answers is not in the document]
JessamyT:
```suggestion
To represent the drinking cup held in your robot's gripper, create a transform with the cup's measurements:
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+We assume the cup is situated against the palm of the gripper's hand.
+This offset is then equal to the radius of the cup (45mm) plus the distance from the gripper's origin (where it meets the arm) to its palm (110mm in this example).
+If your gripper has different dimensions, change the offset accordingly.
+
+Now that you have created the table and tissue box obstacles and the cup transform, create a [WorldState](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.WorldState) that includes all of them:
+
+```python {class="line-numbers linkable-line-numbers"}
+# Create a WorldState that includes the table and tissue box obstacles, and the cup transform
+world_state = WorldState(obstacles=[obstacles_in_frame], transforms = transforms)
+```
```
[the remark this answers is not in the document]
JessamyT: I'm leaving it in the full tutorial code though, so I think I'll leave it in here so that it's matchy-matchy and people can refer back and forth more easily as they build out there full code file. For me, when going through the second tutorial, I found that to be useful.

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: That's a better `why` than mine, nice!

docs/tutorials/services/constrain-motion.md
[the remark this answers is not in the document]
andf-viam: Makes sense to me!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+
+## Define start and end poses
+
+You need to tell the robot where to pick up the cup, and where to put it down.
```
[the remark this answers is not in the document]
JessamyT: Could go either way but personally I like a new big heading being sort of a fresh set of steps. That way, they know that there isn't a missing first step of defining start and end poses? I think if I had a full-on thing like "Now that the worldState and obstacles are defined...." that would be different, but I don't have any such thing to say here that would be relevant and additive as opposed to verbose, I don't think.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,432 @@
...
+The previous tutorial [introduced the concept of defining poses](../plan-motion-with-arm-gripper/#command-an-arm-to-move-with-the-motion-service).
+For this tutorial, you'll define a pose for the gripper that is just above the table on one side of the tissue box, and another pose on the other side of the tissue box.
+
+You'll also want to define some waypoints.
```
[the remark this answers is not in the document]
JessamyT: Fair question, but it seems we're not so strict about that especially when it comes to introducing bits of tutorials. Here, I agree I think it makes more sense in future tense. It's not an imperative because I don't want them to actually do anything just yet; I want to explain the concept first and then we'll get imperative later. Unless @npentrel has other guidance?

docs/tutorials/services/constrain-motion.md
JessamyT: (Sidebar: now that I realize "info" is a thing, if this remains as an alert I think "info" makes more sense than "tip").

I'm not entirely sure what to do here for a couple of reasons:

Some of it is specific to my particular setup. For example, theta=0 doesn't magically make the gripper orient a certain way--it totally depends on someone's frames/hardware etc.

The reason I included the other theta options is because I think it helps understand what theta does, and isn't so much about changing theta during motion and/or spilling tea as it is about understanding the concept of theta in general.

All of this information could theoretically be deduced by the reader without it being provided here, which is why I don't necessarily think it belongs in the body text. But, I added it here, because I think orientation vectors are really confusing and wanted to provide another example with explanation in case someone finds it useful.

Also, it's not necessary that all orientation vectors be the exact same, so I don't want to lead with that per se.

docs/tutorials/services/constrain-motion.md
JessamyT:
```suggestion
When you tell the robot to move the cup from one upright position to another, a linear constraint forces the gripper to move linearly and to maintain the upright orientation of the cup throughout the planned path.
```

docs/tutorials/services/constrain-motion.md
andf-viam: Gotcha! If portions are implementation-specific, that makes a good case for being in a tip | info, agreed. I think I'd mostly just like to see the concept & link for "orientation vector" make it into the body content, if possible.

JessamyT: Hmm we do talk about frames a lot in this tutorial and the previous tutorial, with links to the frame system page from both tutorials. The frame system page contains multiple mentions and links to orientation vector, so it might be okay?

docs/tutorials/services/constrain-motion.md
npentrel: the preview is beautiful!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+Be careful when instructing robot arms to move.
+Before running any code, ensure your robotic arm has enough space and that there are no obstacles.
+Also pay attention to your surroundings, double-check your code for correctness, and make sure anyone nearby is aware and alert before issuing commands to your robot.
+{{< /alert >}}
+
+{{<gif webm_src="../../videos/motion_constraints.webm" mp4_src="../../videos/motion_constraints.mp4" poster="../../videos/motion_constraints.jpg" alt="An arm moving a cup from one side of a tissue box to the other, across a table. The cup stays upright." class="alignright" max-width="250px">}}
+
+Say you want your robot to pass you a cup of tea, but you don't want it to spill the water or bump into other objects on the table.
+
+If you followed along in [part 2 of the Motion Service tutorial series](../plan-motion-with-arm-gripper/), you used the [Motion Service](/services/motion/) to move a robot arm and end effector to desired positions.
```
npentrel, line 25:
```suggestion
If you followed along with the [Plan Motion with an Arm tutorial](../plan-motion-with-arm-gripper/), you used the [Motion Service](/services/motion/) to move a robot arm and end effector to desired positions.
```

^ This adds more useful info. We don't otherwise refer to this as a series currently - so I think this is better until we decide on broader changes.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+
+All example code below assumes +Z points from the base of the gripper to the point where its jaws close, +Y points towards the ground when the gripper is holding a cup from the side, and the X axis is the axis along which the jaws close, following the [right-hand rule](https://en.wikipedia.org/wiki/Right-hand_rule).
+The following diagram shows this, as well as the global coordinate system.
+
+![A gripper mounted on an arm. The Z axis of the gripper points from the base of the gripper to the end of its jaws. The X axis points up through the gripper. The Y axis points in the direction along which the jaws open and close (following the right-hand rule). The diagram also shows the global coordinate system with Z pointing up, X down the length of the horizontal gripper, and Y pointing horizontally in the opposite direction of the gripper's Y.](../../img/constrain-motion/gripper-diagram.png)
+
+If you are using a `fake` gripper, there is no real hardware to calibrate and you can [continue this tutorial](#use-a-transform-to-represent-a-drinking-cup), imagining that your fake gripper corresponds to the diagram above.
```
npentrel, line 195:
```suggestion
If you are using a `fake` gripper, there is no real hardware to calibrate and you can [continue to the next section](#use-a-transform-to-represent-a-drinking-cup), imagining that your fake gripper corresponds to the diagram above.
```

You're already in a tutorial so this makes it a bit more specific and reads better.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+
+If you are using a real arm and gripper, use the **Control** tab in the [Viam app](https://app.viam.com/) to move the gripper, look at its reported orientations, and map them to its orientation in the real world.
```
npentrel, line 197:
```suggestion
If you are using a real arm and gripper, use the [**Control** tab](/manage/fleet/robots/#control) in the [Viam app](https://app.viam.com/) to move the gripper, look at its reported orientations, and map them to its orientation in the real world.
```

npentrel (approved):
A few minor comments but otherwise this looks brilliant! Thanks for updating the preview!

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+
+The following example shows the two obstacles defined with the expanded configuration.
+Adjust the dimensions and positions of the obstacles to describe your own scenario:
+
+```python {class="line-numbers linkable-line-numbers"}
+# Use this offset to set your z to calibrate based on where your table actually is
+z_offset = 10
+
+# Create a table obstacle
+table_origin = Pose(x=0.0, y=0.0, z=-19.0+z_offset)
```
biotinker, line 172: It would be nice to add some sort of comment in the code here to say *why* we're putting this at -19

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+table_dims = Vector3(x=2000.0, y=2000.0, z=38.0)
+table_object = Geometry(center=table_origin, box=RectangularPrism(dims_mm=table_dims))
+
+# Create a tissue box obstacle
+box_origin = Pose(x=400, y=0, z=50+z_offset)
+box_dims = Vector3(x=120.0, y=80.0, z=100.0)
```
biotinker, line 178: Similarly here. You and I know that "centering the box at Z=50 places the obstacle such that a 100mm box will have its lower face at z=0 and its top face at z=100" but someone reading this tutorial probably appreciates every additional piece of code explanation.

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+  This makes it easier to calibrate your motion plans based on how high or low your arm is mounted compared to the table.
+
+The following example shows the two obstacles defined with the expanded configuration.
+Adjust the dimensions and positions of the obstacles to describe your own scenario:
+
+```python {class="line-numbers linkable-line-numbers"}
+# Use this offset to set your z to calibrate based on where your table actually is
+z_offset = 10
+
+# Create a table obstacle
```
JessamyT, line 171:
```suggestion
# Create a table obstacle with its top surface at z=0
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+## Modify your robot's working environment
+
+In the previous tutorial, you [defined a table obstacle](../plan-motion-with-arm-gripper/#describe-the-robots-working-environment) to prevent the arm and gripper from hitting the table upon which the arm is mounted.
+In this tutorial, you'll expand on the code that describes your robot's working environment in two ways:
+
+- Define an additional obstacle: a tissue box sitting on the table.
+- Add a `z_offset` parameter to the `table_origin` and to the new `box_origin`.
+  This makes it easier to calibrate your motion plans based on how high or low your arm is mounted compared to the table.
+
+The following example shows the two obstacles defined with the expanded configuration.
```
JessamyT, line 164:
```suggestion
The following example shows the two obstacles defined such that the table's top surface is at z=0 and the tissue box sits on top of the table.
```

docs/tutorials/services/constrain-motion.md
```diff
@@ -0,0 +1,433 @@
...
+```python {class="line-numbers linkable-line-numbers"}
+# Use this offset to set your z to calibrate based on where your table actually is
+z_offset = 10
+
+# Create a table obstacle
+table_origin = Pose(x=0.0, y=0.0, z=-19.0+z_offset)
+table_dims = Vector3(x=2000.0, y=2000.0, z=38.0)
+table_object = Geometry(center=table_origin, box=RectangularPrism(dims_mm=table_dims))
+
+# Create a tissue box obstacle
```
JessamyT, line 176:
```suggestion
# Create a tissue box obstacle. Setting the box's origin to 50mm above the table
# positions the 100mm tall box on top of the table.
```

docs/tutorials/services/constrain-motion.md
JessamyT: True this is better

docs/tutorials/services/constrain-motion.md
JessamyT: It says more in the body text in the previous tutorial which this links to, but I've added to the comment!

biotinker (approved):
LGTM

PR #1418, opened 2023-06-23
DOCS-884: Explore options to hide TOC

andf-viam:
Add a button to hide the TOC:
- Appears next to GH Edit button, uses same styling
- Not shown if TOC is not shown.

Current issues:
- While it hides the TOC, it does not reclaim the space previously used by the TOC, the primary goal of this effort.
- I am no web designer, I'm sure there's a prettier way to do this.

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -136,6 +136,16 @@ function toTop() {
     window.scroll({top: 0, left: 0, behavior: 'smooth'});
 }
 
+// script for the hideToc button
+const tdToc = document.getElementById("wholeToc");
+function hideToc() {
+    if (tdToc.style.visibility === "visible") {
+        tdToc.style.visibility = "hidden";
+    } else {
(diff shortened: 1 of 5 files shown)


npentrel:
Interesting - I think we could change the CSS on this but my main question is whether it should be the side navigation that should be hidden instead. The toc disappears on <1200px screens. Between 760px and 1200px the navigation takes up potential extra space - thoughts?

andf-viam:
Ah I see now that your original comments in the conversation that inspired this were indeed referring to the left-nav. My mistake. No, you're right -- the left-nav would be the place to achieve toggle-able space savings. I'll close this then -- don't think we need this feature for right-TOC then.

andf-viam:
Closing PR in favor of left-nav hiding instead.

andf-viam:
Closing PR in favor of hiding left-nav instead.

PR #1422, opened 2023-06-26
DOCS-870: Fix GetPose Go information

npentrel:


JessamyT (commented):
Seems good but I'll defer to Peter with respect to accuracy.
While we're at it, I don't think it's explicitly clear why someone would use a supplemental transform or what they actually do. Why is it a list instead of a single transform? If you put in a list of transforms, does GetPose use them as a chain off the end of the robot component or does it return poses with respect to each of them? Does each transform need to be defined with respect to the one before it etc., with the "base" one being with respect to the end of the arm/gripper/whatever?
This may all be outside the scope of this PR but commenting in case we can address now.

npentrel:
If Peter can answer this and it's quick to address, we can try to address here, otherwise this is indeed out of scope and should have its own ticket :)

biotinker (approved):
LGTM

biotinker:
To answer the above questions:
> Why is it a list instead of a single transform?

A transform is loosely defined as "thing that can move with the robot rather than being fixed for all time", e.g. a cup that a robot is holding. There's no need to limit this to only one thing. Many things could be moving, each with many things attached to them.

> If you put in a list of transforms, does GetPose use them as a chain off the end of the robot component or does it return poses with respect to each of them?

I don't understand this question, could you clarify?

> Does each transform need to be defined with respect to the one before it etc., with the "base" one being with respect to the end of the arm/gripper/whatever?

The list of transforms is unordered and any given transform in the list has no relationship whatsoever to any other transform in the list, except when the author of the transform list specifies that the parent of one transform is another transform.

JessamyT (approved):
Ahh cool. I was confused because we describe supplemental transforms a bit differently from how we describe transforms in WorldState in `Move`, but now I understand that they're the same thing. Additional changes would be outside this PR's scope so LGTM.

PR #1425, opened 2023-06-26
Remove excess article

andf-viam:
Removed excess `to` article -- LMK if anyone thinks that the other `to` should be removed!


JessamyT (approved):
Good catch, though should it actually keep the other "to"? Fine with either

PR #1428, opened 2023-06-27
Add aliases for extend and program

npentrel:


andf-viam (approved):
LGTM, but if you like we could also standardize indent count: `4` seems to be the most common currently used in the docs:

```
aliases:
    - "/previous/pagename/"
```

npentrel:
Yes but not in this PR. I want this to merge

PR #1431, opened 2023-06-28
DOCS-874: Change tutorial previews

npentrel:


andf-viam (commented):
This is amazing! I noticed the following small things that could make this amazing PR even more amazing:

- Because both tutorial title and tag count can be arbitrary lengths, this makes the cost indicator render with different alignment per tutorial card. Ideally, it'd be desirable to have this final line universally right-bottom aligned.
- What do you think about including the front matter `description` field within each card? I think this info helps the user choose between like tutorials with more insight, and helps give, at a glance, an overview of all tuts (when viewed on this page). `Move a robot arm` is nice and indicative of contained guidance, but `Mock Robot` and `Detect Color` could really use `Create a mock robot using just your personal computer to try using Viam without any robotic hardware.` and `Use the Vision Service in the Viam app to detect a color with the Viam Rover.` Aware: massively changes sizing considerations for each card. Just a thought to consider / try out.
- `Detect color` and  `tutorials/get-started/try-viam-sdk` focuses on the `Try Viam` experience, potentially useful as its own, additional tag? Helpful to guide users to the tuts that can be done with no hardware using our Try Viam approach.
- What do you think of a bit of text up top on `/tutorials/_index.md` describing why we split tut categories, with links to each section? Users might not scroll all the way down to get to the most involved, arguably most interesting ones all the way at the bottom (`Projects`) without us making its existence a bit more clear.

Otherwise I really like what you've done here, and think it really helps to convey the wide variety of tutorials we have -- and the wide variety of cool uses for Viam software! Certainly filtering by tag / cost / level / language / date published (down the line) is much nearer now that you've provided the front matter for each -- thank you!

npentrel:
1. Done
2. Yeah it's a choice to make. I think adding the description makes it too long - so I think it's an either or. We could add the description on hover or something though. thoughts?
3. Currently all tags are components or services. We can do that but I wonder if that would be confusing? I guess we can add a third type of tags in a third color. I think we need to agree to colors then though to help visually differentiate.
4. That split is going away in favour of a [filtering mechanism](https://docs.google.com/document/d/1X90x5Nwj9ondBxINSNy-SOOAyANJKddpZxEVEE7THqs/edit#). We'll still need text but let's discuss that in the coming weeks.

docs/tutorials/services/webcam-line-follower-robot.md
```diff
@@ -10,6 +10,12 @@ mp4Src: "/tutorials/img/webcam-line-follower/lf-tape-follow3.mp4"
 videoAlt: "The green line the camera sees as the rover moves along it."
 aliases:
     - "/tutorials/webcam-line-follower-robot/"
+authors: []
```
JessamyT, line 13:
```suggestion
authors: [ Jessamy Taylor ]
```
[opt nit :)]

docs/tutorials/configure/scuttlebot.md
```diff
@@ -11,6 +11,12 @@ tags: ["base", "camera", "raspberry pi", "scuttle"]
 aliases:
   - "/tutorials/scuttlebot"
   - "/tutorials/scuttlebot/scuttlebot"
+authors: []
```
JessamyT, line 14: Hazal I think?

docs/tutorials/control/scuttle-gamepad.md
```diff
@@ -12,6 +12,12 @@ aliases:
     - "/tutorials/scuttle-gamepad/"
     - "/tutorials/scuttlebot/scuttle-gamepad/"
 tags: ["base", "scuttle", "gamepad"]
+authors: []
```
JessamyT, line 15: Hazal as well I think?

docs/tutorials/services/accessing-and-moving-robot-arm.md
```diff
@@ -11,7 +11,12 @@ tags: ["arm", "motion", "services"]
 aliases:
   - "/tutorials/motion/accessing-and-moving-robot-arm"
   - "/tutorials/motion/"
-# SMEs: William S.
+authors: []
+languages: [ "python", "go" ]
+viamresources: [ "arm" ]
+level: "Intermediate"
```
JessamyT, line 17:
```suggestion
level: "Beginner"
```
It's the first of the series and doesn't assume prior knowledge. Might well be easier than several of the "intermediate" tutorials here.

JessamyT (commented):
I agree that in some cases it'd be nice to have more info, either by lengthening some titles or including all descriptions but trying to shorten them a bit or something. I do like the idea of displaying that info on hover, though I assume that wouldn't work nicely on mobile?

I think right now, the tags are basically the most colorful thing on the docs site. Maybe they could be less intense somehow? Maybe shrink them slightly, or make the tutorial titles bigger, so that the titles are still the main thing and one can scan titles more quickly without one's eye being drawn so hard to all the colorful tags? Or, once the filtering mechanism is in place, maybe the tags don't need to be colorful or even visible on the cards at all?

npentrel:
I think I'd prefer to lengthen titles to ensure they provide the necessary info. No it probably wouldn't really work nicely on mobile, so better if we can solve if without that. 

I think AI Companion Robot is the only one where the description adds info that isn't otherwise immediately obvious - are there any other?

I can't use any other colours without Cristina's involvement. Cristina will need to review this anyway but I'd like to do that when the rest of the page is done. I can make the font-size a bit smaller (0.833 rem).

I think tags will continue to need to be visible to not confuse readers - and they are there to replace the info that is in the description.

npentrel:
Updates:
- hidden cost information
- reduced label size
- improved mobile experience

npentrel:
Nvm I've found how Cristina does labels - I'll copy that - without the hover effect though. 
Screenshot 2023-06-30 at 13 19 16

docs/tutorials/template.md
```diff
@@ -13,7 +13,15 @@ description: "Short description of your tutorial (1 sentence)."
 # videoAlt: "ALT text for the video"
 tags: ["tutorial"]
 draft: true # Change this when you're ready
-# Author: YOUR NAME
+authors: [] # Your Name
+languages: []
+viamresources: []
```
andf-viam, line 18:
```suggestion
viamresources: [] # Specific components or services used in this tutorial
```

docs/tutorials/template.md
```diff
@@ -13,7 +13,15 @@ description: "Short description of your tutorial (1 sentence)."
 # videoAlt: "ALT text for the video"
 tags: ["tutorial"]
 draft: true # Change this when you're ready
-# Author: YOUR NAME
+authors: [] # Your Name
+languages: []
```
andf-viam, line 17:
```suggestion
languages: [] # Viam SDK programming languages used, if any
```

andf-viam (approved):
LGTM! Added two comments to the template.

I think the coloring of the tags looks great now -- nice work!

JessamyT (approved):
I like this color scheme for the tags much better. Is it expected that on mobile the GIFs on the cards have play buttons on them that aren’t clickable? If you click play it takes you to the tutorial instead of playing the video. 
IMG_8227

npentrel:
That probably is the case currently as well. I think that happens if you're in low battery or low data mode. The only way to avoid that is to not have the hyperlink on the image. I'll file a bug to look into it.

PR #1439, opened 2023-06-30
DOCS-901: Convert glossary links to tooltips

andf-viam:
- Converted a few direct links to glossary to use tooltip shortcode instead.


npentrel (approved):
Awesome - thank you!

PR #1450, opened 2023-07-06
DOCS-765: Update robot dog tutorial to use modular resources

JessamyT:
Also covers DOCS-392 (fix processes tab part of this tute)


docs/tutorials/custom/custom-base-dog.md
```diff
@@ -22,12 +22,9 @@ cost: 190
 
 The [base component type](/components/base/) is useful for controlling mobile robots because it gives users intuitive steering controls to use in code as well as from the [Viam app](https://app.viam.com/) remote control interface.
 
-Viam natively supports a wheeled base model, but if you have a quadruped or other form of rover that isn't compatible with the wheeled model, you have two options:
+Viam natively supports a wheeled base model, but if you have a quadruped or other form of rover that isn't compatible with the wheeled model, you can [create a custom component as a modular resource](/extend/modular-resources/).
```
mcvella, line 25:
```suggestion
Viam natively supports a wheeled base model, but if you have a quadruped or other form of base that requires a different underlying implementation, you can [create a custom component as a modular resource](/extend/modular-resources/).

```

As it could be boat, drone, etc

docs/tutorials/custom/custom-base-dog.md
```diff
@@ -167,7 +164,7 @@ Save the file.
 
 Go to the [robot page](https://app.viam.com/robots) for your robot dog that you created when installing `viam-server` on the Pi.
 
-In the banner towards the top of the page, the IP address of the robot dog Pi is displayed under **ips**.
+In the banner towards the top of the page, the IP address of the robot dog Pi is displayed under **IPs**.
```
mcvella, line 167: could you just use 127.0.0.1 (localhost) since this test script is running on the same machine as the freenove server?

docs/tutorials/custom/custom-base-dog.md
```diff
@@ -39,14 +36,14 @@ You’ll learn to implement a custom component type in Viam, and you’ll be equ
 
 ## Code used in this tutorial
 
+- [Module Generator](https://github.com/viam-labs/generator-viam-module)
 - [Robot Dog Base Code on GitHub](https://github.com/viam-labs/robot-dog-base)
```
mcvella, line 40: Shouldn't this code be updated to match the layout created by the module generator?

Also if the following can't point to 127.0.0.1, make it configurable via config attributes?
        self.client_socket.connect(("PASTE DOG IP ADDRESS HERE", 5001))

mcvella (changes requested):
Overall looks great except the linked code seems to not be updated and match the tutorial!

docs/extend/custom-components-remotes.md
```diff
@@ -69,9 +69,8 @@ Otherwise, the class won’t instantiate.
 
 {{% /alert %}}
 
-The following tutorials also explain how to add custom components as remotes:
+The following tutorial also explains how to add a custom components as a remote:
```
JessamyT, line 72:
```suggestion
The following tutorial also explains how to add a custom component as a remote:
```

docs/tutorials/custom/custom-base-dog.md
JessamyT: Yeah--it is already configurable via config attributes in robotdog.py, but I will make an entirely new repo and link to that

docs/tutorials/custom/custom-base-dog.md
```diff
@@ -209,19 +206,110 @@ You can also try turning the robot off and on again, and then retrying the proce
...
-To create a custom base model, you need to subclass the base component type.
-In other words, you need a script that defines what each base component method (for example `set_power`) makes the robot dog do.
+To create a custom base model, you need a script that defines what each base component method (for example `set_power`) makes the robot dog do.
+
+Open your newly created <file>robotdog.py</file> file.
+It contains stubs of all the [base API methods](/components/base/#api), but you need to modify these method definitions to actually send commands to the robot dog.
 
-Take a look at [<file>my_robot_dog.py</file>](https://github.com/viam-labs/robot-dog-base/blob/main/my_robot_dog.py).
-It creates a "RobotDog" model of the base component type, and defines the `set_power`, `stop`, `is_moving`, and `do` methods by specifying which corresponding commands to send to the Freenove dog server when each of these methods is called.
+Take a look at [<file>robotdog.py</file>](https://github.com/viam-labs/robot-dog-module/blob/main/src/robotdog.py).
```
JessamyT, line 311:
```suggestion
Take a look at [<file>robotdog.py</file>](https://github.com/viam-labs/robot-dog-module/blob/main/robotdog/src/robotdog.py).

```

docs/tutorials/custom/custom-base-dog.md
JessamyT: Not here, because dog_test.py is being run on a different  computer

JessamyT: I could rewrite this to have them run the script on the local machine instead of the laptop but I don't think it matters so I'd rather leave it as-is.
Maybe it could be changed in the configured component (like maybe I don't even need that to be a configurable attribute, but could just hardcode it as localhost), but as-written, and as-tested, this way works. Also, then maybe it gives them some flexibility with where they're running things??

npentrel:
Hey  @JessamyT - not diving in to review but just to ask to please keep the instructions for a custom component because we do still support them for edge cases and it would be wasteful to not keep the docs on it since they're already written. I think there are two ways of going about this: 1) have two tutorials and put the custom component one in the custom component directory or 2) use tabs.

npentrel:
Actually let's hold off on doing either of the solutions to show both until I've spoken more with Shawn about it

npentrel:
Ok let's stick with just keeping this as a modular resources. Matt pointed out this is a bad use case for a custom component example - so I'll have a poke around for a good example for that

docs/tutorials/custom/custom-base-dog.md
mcvella: so if you don't want it to run on the same machine, can't you use robotdog.local instead of the IP address?

mcvella (approved):

PR #1464, opened 2023-07-07
Improve section lists

npentrel:


JessamyT (commented):
Since "Manage" and "SDKs" are shorter lists, maybe put them next to each other so that on narrower screens there is less white space
image

npentrel:
done @JessamyT

PR #1473, opened 2023-07-10
RSDK-3613 update motion tutorial code to not use a singularity

raybjork:
I tested this new position works on a UR arm.

### DIFF
--- a/docs/tutorials/services/accessing-and-moving-robot-arm.md
+++ b/docs/tutorials/services/accessing-and-moving-robot-arm.md
@@ -252,7 +252,7 @@ if err != nil {
 {{% /tab %}}
 {{< /tabs >}}
 
-If you execute the sample joint move statement above, your arm should move the last joint a small amount (15 degrees).
+If you execute the sample joint move statement above, the third joint of your arm should move a small amount (30 degrees).
 Feel free to experiment further with joint position commands by changing the values for each joint and re-sending the commands.
 
 When you are ready to move on, the next section will show you how to use **pose commands**.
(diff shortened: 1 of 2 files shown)


JessamyT (commented):
This is fine but this code is carried over from https://github.com/viamrobotics/docs/blob/main/docs/tutorials/services/accessing-and-moving-robot-arm.md, and there's some accompanying body text. Also the code comments here and there should be updated. Mind adding that to this PR?

Also also, this will induce a bigger motion right? Since it's both a lower joint and a bigger angle change? Should we not be worried that people might smack themselves/their surroundings with such a move?

raybjork:
> This is fine but this code is carried over from https://github.com/viamrobotics/docs/blob/main/docs/tutorials/services/accessing-and-moving-robot-arm.md, and there's some accompanying body text. Also the code comments here and there should be updated. Mind adding that to this PR?
> 
> Also also, this will induce a bigger motion right? Since it's both a lower joint and a bigger angle change? Should we not be worried that people might smack themselves/their surroundings with such a move?

I'll make the fixes to the other doc.  I'm not worried about your other question though.  We run through this tutorial weekly and never have had concerns like that.  This movement is followed up by an 100 mm move up in tool-space so there should be no danger posed by this change.

raybjork:
I also want to make a note that I tested this new position and that both it and the subsequent position in the tutorial are reachable from either the xarm or the UR arm.  The motion team QA tests using this tutorial on both arms, so this is an improvement for us

docs/tutorials/services/accessing-and-moving-robot-arm.md
JessamyT, line 1: Line 255 references this change so it should be changed to something like:

If you execute the sample joint move statement above, the third joint of your arm should move a small amount (30 degrees).

JessamyT (approved):
Line 255 still needs changing. Otherwise looks good; thanks for updating.

PR #1478, opened 2023-07-10
Update tipsy.md

JessamyT:
Real quick nit fixes


JessamyT:
Moved onto existing PR. Just happened to be looking and noticed these.

PR #1512, opened 2023-07-14
Update IP address info on robot dog tutorial

JessamyT:
In my testing, it seems like this works more reliably. Including the note about other IP addresses because I was having some trouble with the server connecting and this might help folks who run into the same issue.

### DIFF
--- a/docs/tutorials/custom/custom-base-dog.md
+++ b/docs/tutorials/custom/custom-base-dog.md
@@ -363,17 +363,17 @@ Click **Create Component**.
 ![Screenshot of the Viam app CONFIG tab with the Config subtab open, showing quadruped configured.](../../img/custom-base-dog/config-modular-component.png)
 
 You need to add the `ip_address` and `port` attributes to your base config.
-In the attributes field, paste the following:
+In the attributes field, paste the following, replacing `<HOSTNAME>` with your Pi's hostname (for example, `"ip_address": "robotdog.local"`):
 
 ```json
 {
-  "ip_address": "localhost",
+  "ip_address": "<HOSTNAME>.local",
   "port": 5001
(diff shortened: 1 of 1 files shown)


docs/tutorials/custom/custom-base-dog.md
```diff
@@ -363,17 +363,17 @@ Click **Create Component**.
...
 {
-  "ip_address": "localhost",
+  "ip_address": "<HOSTNAME>.local",
   "port": 5001
 }
 ```
 
-Use `localhost` as the local IP address because the Freenove server is running on the same machine as `viam-server`.
-If for some reason you are running the module on a separate machine (instead of on the Pi), you need to replace it with the IP address of the machine where the module is running.
+This ".local" address works because the Freenove server is running on the same machine as `viam-server`.
```
JessamyT, line 375:
```suggestion
This ".local" address works because the Freenove server is running on the same network as `viam-server`.
```
Or could just remove:
```suggestion
```

docs/tutorials/custom/custom-base-dog.md
```diff
@@ -363,17 +363,17 @@ Click **Create Component**.
...
+If for some reason you are running the module on a separate machine (instead of on the Pi), you need to replace it with the IP address of the machine where the module is running, for example, `"ip_address": "10.0.0.123"`.
```
JessamyT, line 376:
```suggestion
If this doesn't work, you can instead try using the IP address of the machine where the module is running, for example, `"ip_address": "10.0.0.123"`.
```
I found this to be true and necessary, but I'm still not sure why, so should I do this or take out the sentence entirely?

mcvella (approved):

PR #1513, opened 2023-07-14
Add Type to glossary

JessamyT:
Realized this isn't in there, though model is, and it seems increasingly necessary as we talk about defining new APIs and namespaces and such.

### DIFF
--- a/docs/appendix/glossary/model.md
+++ b/docs/appendix/glossary/model.md
@@ -6,5 +6,5 @@ short_description: A particular implementation of a component type. For example,
 aka:
 ---
 
-A particular implementation of a {{< glossary_tooltip term_id="component" text="component" >}} type.
-For example, UR5e is a model of the arm component type.
+A particular implementation of a {{< glossary_tooltip term_id="type" text="component or service type" >}}.
+For example, UR5e is a model of the arm {{< glossary_tooltip term_id="component" text="component" >}} type.
--- a/docs/appendix/glossary/resource.md
+++ b/docs/appendix/glossary/resource.md
@@ -8,7 +8,7 @@ aka:
 
 Resources are individual, addressable elements of a robot.
 
-{{< glossary_tooltip term_id="part" text="Parts" >}} operate multiple types of resources:
+{{< glossary_tooltip term_id="part" text="Parts" >}} operate multiple kinds of resources:
 
 - physical {{< glossary_tooltip term_id="component" text="components" >}}
 - software {{< glossary_tooltip term_id="service" text="services" >}}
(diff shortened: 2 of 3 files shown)


sguequierre (changes requested):
This looks like a great update-- I think we need SME approval here for accuracy though-- while reviewing I'm realizing it is further complicated by motor etc actually being sub-type and "type" refers specifically to that component or service label (see https://docs-test.viam.dev/3ed908fedc3d0960530b9fc0c5e78ab0e0fbe969/public/appendix/glossary/ the first triplet definition). I'm going to work to update the DOCS-915 PR to make that clear (and format that glossary better) 😆

JessamyT:
Yeah I realize there are the two uses of "type" but this one is far more common than the other one, at least in the docs and most canonical use I've seen. I put "resource type" on there in parentheses as an attempt at disambiguation, but I realize the other usage of type also has to do just as much with resources, so idk what to call each of them. Motor is both a type and a sub-type, depending on which level of "type" we're talking about ;( But we use this usage of "type" (not sub-type) throughout the rest of the docs and glossary right now, so I feel like we need to add something here to define it! Would love ideas as to how to distinguish.

sguequierre:
> Yeah I realize there are the two uses of "type" but this one is far more common than the other one, at least in the docs and most canonical use I've seen. I put "resource type" on there in parentheses as an attempt at disambiguation, but I realize the other usage of type also has to do just as much with resources, so idk what to call each of them. Motor is both a type and a sub-type, depending on which level of "type" we're talking about ;( But we use this usage of "type" (not sub-type) throughout the rest of the docs and glossary right now, so I feel like we need to add something here to define it! Would love ideas as to how to distinguish.

as Motor in API namespace triplet is subtype and I'm not sure what you mean by the second usage in the technical sense (we have def referred to it as type of resource informally I'm just not sure where)--  I would recommend changing type in this entire entry to subtype and trying to change the other usage where you see it, even though it's super intuitive to just call it type. It would help SUBTYPE to stick in ppls heads if it was **very obvious**

JessamyT:
Well, the biggest example of where we use "type" this way is on the app, and in config files. There is the "Type" field when creating a new component, that maps to the `type` element in the raw JSON.

sguequierre:
> Well, the biggest example of where we use "type" this way is on the app, and in config files. There is the "Type" field when creating a new component, that maps to the `type` element in the raw JSON.

Yeah that's true, and could provoke some confusion. Would you want to discuss this in devrel or docs sync?

JessamyT:
Closing because now we have https://github.com/viamrobotics/docs/pull/1544

PR #1524, opened 2023-07-14
Add a specific directive to the Cartographer configuration to SLAM page

sguequierre:
thought this would be a good improvement to make  now that the API methods are on that page!

* Adds a specific directive with a link to the Cartographer configuration page.


JessamyT (commented):
Seems redundant with the model name link, and would be different from all component pages. What if instead, you change line 29 to "Integrated SLAM libraries include the following.
Click the model name for configuration instructions." or similar?

sguequierre:
@je

> Seems redundant with the model name link, and would be different from all component pages. What if instead, you change line 29 to "Integrated SLAM libraries include the following. Click the model name for configuration instructions." or similar?

Thanks, have changed!

JessamyT (approved):
LGTM!

PR #1547, opened 2023-07-20
DOCS-865: Change to get_detections_from_camera

JessamyT:


docs/services/vision/detection.md
```diff
@@ -223,7 +223,7 @@ Wait for the robot to reload, and then go to the **Control** tab to test the str
 
 ## Code
 
-The following code gets the robot’s vision service and then runs a color detector vision model on an image from the robot's camera `"camera_1"`:
+The following code gets the robot’s Vision Service and then runs a color detector vision model on output from the robot's camera `"cam1"`:
```
npentrel, line 226:
```suggestion
The following code gets the robot’s vision service and then runs a color detector vision model on output from the robot's camera `"cam1"`:
```

I think we settled on capitalizing if it's the abstract service and not if it's an instance of it

docs/services/vision/detection.md
```diff
@@ -232,13 +232,18 @@ The following code gets the robot’s vision service and then runs a color detec
...
 cam1 = Camera.from_robot(robot, "cam1")
-# grab Viam's vision service for the detector
+# Grab Viam's vision service for the detector
 my_detector = VisionClient.from_robot(robot, "my_detector")
 
+detections = await my_detector.get_detections_from_camera(cam1)
+
+# If you need to get an image first and then run detections on it,
+# you can do it this way (generally slower but useful if you need to,
+# for instance, crop the image before running the detector on it):
```
npentrel, line 244:
```suggestion
# you can do it this way (generally slower but useful if you need to use the image
# afterwards):
```
^ cropping you could define a transform camera for - you only need to do this if you want to save the image after finding something on it for example . Like if it's a bird camera and you detected a bird you'd want to save the image. You can only do that if you still have the image otherwise you'd need to take another image and you don't know if the bird would still be there

JessamyT:
```suggestion
# you can do it this way (generally slower but useful if you need to
# use the image afterwards
```

PR #1568, opened 2023-07-25
DOCS-946: Deal with camera SDK differences

JessamyT:
The camera API varies much more widely between languages than most of the APIs. I think we should address that in some way. I'm not sure whether this is the way--very open to other ideas but had a thought and wanted to make a suggestion!

I'll make a ticket to discuss.

### DIFF
--- a/docs/components/camera/_index.md
+++ b/docs/components/camera/_index.md
@@ -145,6 +145,14 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 {{% /tab %}}
 {{% tab name="Go" %}}
 
+{{% alert title="Info" color="info" %}}
+
+Unlike most Viam [component APIs](/program/apis/#component-apis), the methods of the Go camera client do not map exactly to the Python SDK camera methods.
+`Stream` works somewhat differently from the Python `get_image` method.
+In the example below, note how a stream is constructed before the next image is gotten from the stream.
+
+{{% /alert %}}
+
 **Parameters:**
 
 - `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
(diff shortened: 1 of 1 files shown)


JessamyT:
But the API list of methods, as well as the heading here, says "GetImage," which is pretty different from "Stream"

docs/components/camera/_index.md
```diff
@@ -145,6 +145,14 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 {{% /tab %}}
 {{% tab name="Go" %}}
 
+{{% alert title="Info" color="info" %}}
+
+Unlike most Viam [component APIs](/program/apis/#component-apis), the methods of the Go camera client do not map exactly to the Python SDK camera methods.
```
npentrel, line 150:
```suggestion
Unlike most Viam [component APIs](/program/apis/#component-apis), the methods of the Go camera client do not map exactly to the names of the other SDK's camera methods.
```

docs/components/camera/_index.md
```diff
@@ -145,6 +145,14 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
...
+`Stream` works somewhat differently from the Python `get_image` method.
+In the example below, note how a stream is constructed before the next image is gotten from the stream.
```
npentrel, line 152: I'd prefer to not use Python as a reference:
```suggestion
To get an image in the Go SDK, you first need to construct a `Stream` and then you can get the next image from that stream.
```

npentrel (commented):
Added two suggestions to make the note more general - let me know what you think

JessamyT:
I like those suggestions. Do you think I should just close this PR though? I still think it could clear up confusion for some users, but I don't feel super strongly about that now. Not sure where we netted out.

npentrel (approved):

PR #1574, opened 2023-07-26
Add RPLidar rule

npentrel:


andf-viam (approved):
LGTM, thanks! (ignore my double commit, it was unnecessary in retro)

PR #1575, opened 2023-07-26
Dad 283 tutorial viam on vm

npentrel:
Moving from https://github.com/viamrobotics/docs/pull/590


sguequierre:
sorry to comment and not be of much help-- This should duplicate a ton of the work from extend if I have the right idea

PR #1583, opened 2023-07-27
DOCS-731: Update list in table

npentrel:
The way to make these lists inside a table is in HTML - we might be able to fix this some other way but for now that's the way


sguequierre (approved):
LGTM if "buttons" -> "button" to keep the subsection title from looking awkward ("Each buttons has the following field") 

Thank you for figuring this out, it looks a lot better

PR #1586, opened 2023-07-27
DOCS-933: Add image of detection bounding boxes to detections page

JessamyT:


npentrel:
There is good but I was also hoping for one at the bottom before the code section where we instruct people to go to the control tab. And we probably want the same for classifications

sguequierre (approved):
LGTM (oops, once naomi's comments are addressed)!

JessamyT:
Looks like we already have one on the classifications page. It'd be nice to have one for segmentation, but I don't have a 3D camera set up. Do we have any good images around?

npentrel:
Please ask the vision team if they have something for the segmenter page

JessamyT:
Realized there's not actually a good place to put the image for segmentation right now, so created https://viam.atlassian.net/browse/DOCS-975, and maybe adding that image can be part of that ticket instead of this one?

npentrel (approved):

PR #1587, opened 2023-07-27
DOCS-941: Fix table rendering in includes

andf-viam:
Fix table rendering in includes by removing wrapping `<div>` elements from `readfile.html`.

- Also using opportunity to expand `note.md` to showcase HTML and MD tables in includes, useful for testing this.

### DIFF
--- a/docs/note.md
+++ b/docs/note.md
@@ -123,7 +123,7 @@ This is **some markdown.**
 
 ### Alerts Shortcodes
 
-{{< alert title="Note" color="note" >}}
+{{< alert title="Important" color="note" >}}
 It can even contain shortcodes.
 {{< /alert >}}
 
@@ -416,7 +416,7 @@ If you add an MP4 file, test it on an iPhone with the deployed link on the PR.
 
 {{< /alert >}}
 
-{{< alert title="Note" color="note" >}}
+{{< alert title="Important" color="note" >}}
 The gif can and should be used only in the frontmater `images` variable and only if close in size to 1MB.
 The `images` variable sets it to be the preview image on social platforms (for example it will be the preview when you share a link on slack).
 Link previews do not support `webm` and `mp4` but they do support gifs.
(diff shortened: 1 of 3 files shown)


docs/note.md
```diff
@@ -267,7 +267,7 @@ Figure styles the Attribution text as body text.
 
 {{< readfile "/static/include/sample.md" >}}
 
-Section content before this line is contained in an included file: /static/include/sample.md
+Section content before this line is contained in an included file: <file>/static/include/sample.md</file>
```
JessamyT, line 270: Why this change?

static/include/sample.md
```diff
@@ -1,24 +1,94 @@
-
 ### This is a Heading in the Sample File
-This entire section was added to the current topic using the readfile shortcode.
+
+This entire section was added to the current topic using the `readfile` shortcode.
 
 {{% alert="Note" color="note" %}}
```
JessamyT, line 5:
```suggestion
{{% alert="Important" color="note" %}}
```
[out of diff nit] While we're here...realizing we don't use "note" anymore so should be consistent through this doc :)

static/include/sample.md
```diff
@@ -1,24 +1,94 @@
...
-Changes to included files do not appear in local mode (hugo server -D). You'll need to halt and restart the server to view the changes.
+Changes to included files do not appear in local mode (`hugo server -D`). You'll need to halt and restart the server to view the changes.
```
JessamyT, line 6:
```suggestion
Changes to included files do not appear in local mode (`hugo server -D`).
You'll need to halt and restart the server to view the changes.
```
[out of diff opt nit]

JessamyT (approved):
Looks good function-wise, though I still don't know why the fence is there :) so maybe I shouldn't be reviewing. Threw some handy out-of-diff nits in there too; can of course ignore. Not sure why the change on line 270 of note.md though

docs/note.md
andf-viam: Why not this change? It's a file path!

PR #1588, opened 2023-07-27
DOCS-960: Update subtype links

JessamyT:


andf-viam (approved):
LGTM, but looks like Naomi also captured this in https://github.com/viamrobotics/docs/pull/1589

PR #1595, opened 2023-07-28
DOCS-928: Move API overviews into includes

JessamyT:
This does change the API page in some ways:
- Adds links to each component doc in the text under the heading
- Adds movement sensor model specifications (which were on the movement sensor page but not the APIs page)
  - This isn't necessarily a good or bad thing, was just out of necessity, and I'm just calling out since with includes it's harder to see in the diff, in case anyone has opinions about this change
- Adds DoCommand to the tables for every component on the API page (except sensor and board because there's nothing to link for them yet; made https://viam.atlassian.net/browse/DOCS-964). As above, this isn't necessarily an improvement, just a change because it's needed on the component pages so it's in the included files.
- Shortens vision service descriptions (makes sense with recent addition of vision service API documentation update)

### DIFF
--- a/docs/program/apis/_index.md
+++ b/docs/program/apis/_index.md
@@ -229,72 +229,44 @@ The [encoder component](/components/encoder/) supports the following methods:
 
 The [gantry component](/components/gantry/) supports the following methods:
 
-| Method Name | Description |
-| ----------- | ----------- |
-[Position](/components/gantry/#position) | Get the current positions of the axes of the gantry in mm. |
-[MoveToPosition](/components/gantry/#movetoposition) | Move the axes of the gantry to the desired positions. |
-[Lengths](/components/gantry/#lengths) | Get the lengths of the axes of the gantry in mm. |
-[Stop](/components/gantry/#stop) | Stop the gantry from moving. |
-[IsMoving](/components/gantry/#ismoving) | Get if the gantry is currently moving. |
+{{< readfile "/static/include/components/apis/gantry.md" >}}
 
 ### Gripper
 
-Method Name | Description
------------ | -----------
-[`Open`](/components/gripper/#open) | Open the gripper.
-[`Grab`](/components/gripper/#grab) | Close the gripper until it grabs something or closes completely.
-[`Stop`](/components/gripper/#stop) | Stop the gripper's movement.
-[`IsMoving`](/components/gripper/#ismoving) | Report whether the gripper is currently moving.
+The [gripper component](/components/gripper/) supports the following methods:
+
+{{< readfile "/static/include/components/apis/gripper.md" >}}
 
 ### Input Controller
 
-| Method Name | Description |
-| ----------- | ----------- |
-| [Controls](/components/input-controller/#controls) | Get a list of input `Controls` that this Controller provides. |
-| [Events](/components/input-controller/#events) | Get the current state of the Controller as a map of the most recent [Event](/components/input-controller/#event-object) for each [Control](/components/input-controller/#control-field). |
-| [RegisterControlCallback](/components/input-controller/#registercontrolcallback) | Define a callback function to execute whenever one of the [`EventTypes`](/components/input-controller/#eventtype-field) selected occurs on the given [Control](/components/input-controller/#control-field). |
+The [input controller component](/components/input-controller/) supports the following methods:
+
+{{< readfile "/static/include/components/apis/input-controller.md" >}}
 
 ### Motor
 
-Method Name | Description
------------ | -----------
-[SetPower](/components/motor/#setpower) | Set the power to send to the motor as a portion of max power.
-[GoFor](/components/motor/#gofor) | Spin the motor the specified number of revolutions at specified RPM.
-[GoTo](/components/motor/#goto) | Send the motor to a specified position (in terms of revolutions from home) at a specified speed.
-[ResetZeroPosition](/components/motor/#resetzeroposition) | Set the current position to be the new zero (home) position.
-[GetPosition](/components/motor/#getposition) | Report the position of the motor based on its encoder. Not supported on all motors.
-[GetProperties](/components/motor/#getproperties) | Return whether or not the motor supports certain optional features.
-[Stop](/components/motor/#stop) | Cut power to the motor off immediately, without any gradual step down.
-[IsPowered](/components/motor/#ispowered) | Return whether or not the motor is currently on, and the amount of power to it.
-[IsMoving](/components/motor/#ismoving) | Return whether the motor is moving or not.
+The [motor component](/components/motor/) supports the following methods:
+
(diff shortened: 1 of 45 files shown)


docs/program/apis/_index.md
```diff
@@ -332,105 +276,71 @@ Documentation on using these methods in your SDK code is found on [service pages
...
 ### Vision
 
-Method Name | Description |
------------ | ----------- |
-`DetectionsFromCamera` | Get a list of detections in the next image given a camera and a detector. |
-`Detections`| Get a list of detections in the given image using the specified detector. |
-`ClassificationsFromCamera` | Get a list of classifications in the next image given a camera and a classifier. |
-`Classifications` | Get a list of detections in the given image using the specified detector. |
-`ObjectPointClouds`| Returns a list of the 3D point cloud objects and associated metadata in the latest picture obtained from the specified 3D camera (using the specified segmenter). |
+Different [Vision Service](/services/vision/) models support different methods:
```
sguequierre, line 321:
```suggestion
Different models of the [Vision Service](/services/vision/) support different methods:
```
or 
```suggestion
Different API methods are available for each model of [Vision Service](/services/vision/):
``` 
/ endpoints 

(opt. for user comprehension!)

docs/services/sensors.md
```diff
@@ -29,7 +26,7 @@ Go to your robot's **Code sample** tab on the [Viam app](https://app.viam.com) f
 
 {{% /alert %}}
 
-### Sensors
+### GetSensors
```
sguequierre, line 29:
```suggestion
### Sensors
```
I think this needs to stay `Sensors`, as least as long as it's defined so in the interface here https://pkg.go.dev/go.viam.com/rdk/services/sensors#Service and client file. Same for `Readings`!

sguequierre, line 29:
```suggestion
### Sensors
```
I think this needs to stay `Sensors`, as least as long as it's defined so in the interface here https://pkg.go.dev/go.viam.com/rdk/services/sensors#Service and client file. Same for `Readings`!

static/include/components/apis/gripper.md
```diff
@@ -0,0 +1,7 @@
+Method Name | Description
+----------- | -----------
+[`Open`](/components/gripper/#open) | Open the gripper.
```
sguequierre, line 3:
```suggestion
[Open](/components/gripper/#open) | Open the gripper.
```
Needs to be not code here through line 6

static/include/services/apis/sensors.md
```diff
@@ -0,0 +1,4 @@
+Method Name | Description
+----------- | -----------
+[`GetSensors`](/services/sensors/#getsensors) | Returns a list of names of the available sensors.
```
sguequierre, line 3:
```suggestion
[Sensors](/services/sensors/#sensors) | Returns a list of names of the available sensors.
```
Think not opt. -- See comment above about https://pkg.go.dev/go.viam.com/rdk/services/sensors#Service!

static/include/services/apis/base-rc.md
```diff
@@ -0,0 +1,4 @@
+Method Name | Description
```
sguequierre, line 1: With code comment above, are you intending to make the service client method names code format separately from attribute? fine w/ me either way but thought i'd clarify

sguequierre (changes requested):
Thanks so much for this Jessamy, this will make updates so much less of a hassle in the future! 
Some opt comments around style and fixup method changes

docs/services/sensors.md
JessamyT: So, I realize that in the Go SDK/RDK it's Sensors and Readings, but in the proto it's GetSensors and GetReadings. And in the Python it's get_sensors and get_readings. This is the same exact thing that we see across other APIs and we made the decision with those to use what's in the proto.

docs/program/apis/_index.md
JessamyT: Mm I think since the Vision Service is one thing, the current wording makes the most sense? I might be missing what you're trying to clarify though

static/include/services/apis/base-rc.md
JessamyT: Naw I wasn't changing anything with that--I just copy pasted from what was already there. But you're right, it is inconsistent! Will make it consistent now.

JessamyT: Right now on https://docs.viam.com/program/apis/ it's about 50/50 haha

static/include/services/apis/sensors.md
JessamyT: See my comment on the other one--we use the proto definition, not the Go SDK one, for other APIs, so this should follow that too.

docs/services/sensors.md
JessamyT: https://github.com/viamrobotics/api/blob/main/service/sensors/v1/sensors.pb.go

static/include/services/apis/base-rc.md
JessamyT: (Will address once Slack convo comes to a consensus)

docs/program/apis/_index.md
npentrel: I think any of these three work 🤷‍♀️

npentrel:
PR looks great - once we decide on code vs no code :)

static/include/services/apis/base-rc.md
andf-viam: A consensus has been reached! The people have spoken!! 😄

JessamyT: Let there be code! :D

docs/services/sensors.md
sguequierre: This is the one thing I would say still blocks-- I really don't think we should change for the Go Client's sake and also for modular resource creaters. Can discuss with everyone but I think it needs to be consistent with https://github.com/viamrobotics/rdk/blob/main/components/base/client.go because that is the actual client and if the method's names are off users will be confused

sguequierre (changes requested):
LGTM but don't want changes to be made to sensors service names-- if it was ok with you and @npentrel @andf-viam maybe we can keep `Readings` --> `GetReadings` etc. out of here and discuss in a meet or a bit later this week as a group?

JessamyT:
As discussed, was just making the sensors service consistent with most of the other API documentation.

Hadn't realized that there are a couple of other APIs (sensor component and gantry) that also aren't using the protobuf method names for the docs headings right now. I think we should make them all consistent, but will leave those out of this PR, and I think this can be merged as-is and yeah let's discuss.

sguequierre (approved):
LGTM see comment update!! 
Basically as long as Go code is not made inaccurate, which this change isn't, lgtm

PR #1605, opened 2023-08-02
Update single-line-check

npentrel:
You were right, we can just check for a lower case letter with a period and then a space. That will catch sentences starting with a lower case letter on the same line.

### DIFF
--- a/.github/vale/bin/single-line-check
+++ b/.github/vale/bin/single-line-check
@@ -4,14 +4,14 @@
 
 RC=0
 
-NEWLINE_CHECK=$(grep -v 'alt=' $(git diff --diff-filter=d --name-only $1 | grep '\.md$') | grep -v 'md:#' | grep -v '!\[' | grep -v '//' | grep -v '|' | grep -v '#' | grep '[a-z]\. [A-Z]')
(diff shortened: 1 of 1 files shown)


.github/vale/bin/single-line-check
```diff
@@ -4,14 +4,14 @@
 
 RC=0
 
-NEWLINE_CHECK=$(grep -v 'alt=' $(git diff --diff-filter=d --name-only $1 | grep '\.md$') | grep -v 'md:#' | grep -v '!\[' | grep -v '//' | grep -v '|' | grep -v '#' | grep '[a-z]\. [A-Z]')
+NEWLINE_CHECK=$(grep -v 'alt=' $(git diff --diff-filter=d --name-only $1 | grep '\.md$') | grep -v 'md:#' | grep -v '!\[' | grep -v '//' | grep -v '|' | grep -v '#' | grep '[a-z]\. ')
```
JessamyT, line 7:
```suggestion
NEWLINE_CHECK=$(grep -v 'alt=' $(git diff --diff-filter=d --name-only $1 | grep '\.md$') | grep -v 'md:#' | grep -v '!\[' | grep -v '//' | grep -v '|' | grep -v '#' | grep '\. ')
```
[opt] I don't think the sentence needs to end in a lower case letter does it? What if it ends in a symbol (like in a link or something) or abbreviation or some such?

JessamyT (approved):
Looks good! An optional additional suggestion to further catch more things

npentrel:
Will merge and if there are any other symbols we can add them separately :)

PR #1609, opened 2023-08-02
Fix base get properties

JessamyT:


sguequierre (approved):
LGTM thanks for catching this!

PR #1617, opened 2023-08-03
DOCS-948: Add PulseAudio troubleshooting

andf-viam:
Add brief troubleshooting item addressing `PulseAudio` / `jackd` errors with best effort fix:
- Fix is specific to Linux OS / sound library combo, but included RPI example command to install missing deps for that specific platform / sound library combo.


dannenberg (approved):
lgtm. it may be worth mentioning that "this error can safely be ignored if you do not intend to use audio on your robot"

dannenberg (approved):
wonderful!

PR #1639, opened 2023-08-09
DOCS-986: Update Viam CLI installation

andf-viam:
Update Viam CLI tool installations instructions to use new compiled binary executable downloads.
- Retained "Build from Source" as fifth tab: build in Go for any supported platform.
- Using `stable` in place of `latest` in the URL links per discussion.

Question:
- Sending the downloaded binary to `/usr/local/bin/viam`, so requiring `sudo` twice. Same way we have users [download our pre-built modules](https://docs.viam.com/extend/modular-resources/examples/rplidar/#requirements). Look good?

### DIFF
--- a/docs/manage/CLI.md
+++ b/docs/manage/CLI.md
@@ -34,9 +34,9 @@ You can use the `uname -m` command to determine your system architecture.
 {{% /alert %}}
 
 {{< tabs >}}
-{{% tab name="Linux arm64" %}}
+{{% tab name="Linux aarch64" %}}
 
-To download the Viam CLI on a Linux computer with the `arm64` architecture, run the following commands:
+To download the Viam CLI on a Linux computer with the `aarch64` architecture, run the following commands:
 
 ```{class="command-line" data-prompt="$"}
 sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.com/apps/viam-cli/viam-cli-stable-linux-arm64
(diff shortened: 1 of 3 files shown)


npentrel:
Not really related but I noticed a text shadow on the code elements in the tabs - I've remove that in https://github.com/viamrobotics/docs/pull/1641

abe-winter (approved):
ARM 64 is aarch64 on linux, but arm64 on mac (see pi output below)
```
# I ran this on a pi
pi@aw-pi-modreg:~ $ uname -m
aarch64
```

I copied 'latest' to 'stable' + tested the linux amd64 command, it works.

if useful in the future, we can tweak the filenames so that you can download https://whatever/$(uname -m) and have it work without any thought

andf-viam:
Thanks @abe-winter ! You're right, I've updated the tab title.

It'd be great to use `uname -m` in the download command! If you're amenable to changing the filenames down the line ( `viam-cli-stable-linux-amd64` --> `viam-cli-stable-linux-aarch64` ), you're right: we could just do this:

```
# Linux users:
sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.com/apps/viam-cli/viam-cli-stable-linux-$(uname -m)
# macOS users:
sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.com/apps/viam-cli/viam-cli-stable-darwin-$(uname -m)
```

For now, I've just updated the tab title to match what `uname -m` returns: `aarch64` for Linux. Thanks!

PR #1632, opened 2023-08-09
DOCS-806: Update GPS RTK model docs

JessamyT:

PR #1657, opened 2023-08-10
DOCS-990: Fix gps-nmea board attribute and prune includes file

JessamyT:
As of #1632, the connection attributes includes file is only used by the gps-nmea doc so it doesn't need to be an includes anymore. This PR moves its content into the one place it is used, and edits its contents because "board" is listed in two different places and it belongs only in i2c attributes. Taking out the extra JSON examples too since all of that is captured in the main JSON example tabs.


andf-viam (approved):
LGTM, thanks!

andf-viam (approved):
LGTM, thanks!

PR #1670, opened 2023-08-14
Fix base API code examples

JessamyT:
If you try to move a base at 1mm/s, or 1 deg/sec, not only would that motion theoretically be almost imperceptible to a human, but it won't actually physically work because most motors won't get enough power to break the static friction, so it will stall/not move at all. Changing these to values that would reasonably be expected to work for many hobbyist bases.

Also, I think the Go examples all need "nil" as the extra argument instead of nothing?


npentrel (approved):
Good catch

PR #1672, opened 2023-08-16
Fix classification typo

JessamyT:
Also noticed there's not a good reason to tell them to head to the components tab since they might test with existing images and not need to configure a transform camera.


npentrel (approved):
true you can remove the instruction to go to the components tab. It's probably on the detection page too(?)

JessamyT:
I believe we recently removed it from there? Removing from segmentation as well in a different PR

PR #1674, opened 2023-08-17
DOCS-1012: Fix power supply notes on Jetson Nano 

sguequierre:
* Add different spec for power jack connector for Orin Nano as requested
* Add note about polarity into troubleshooting to make users extra aware 
* Edit instructions to make clear micro-usb powering is a no-go


sguequierre:
@npentrel closing after receiving separate/slack confirmation from chris that this looks good and is correct!

PR #1676, opened 2023-08-17
DOCS-1018: Add duplicate component button

andf-viam:
Adds brief notice of **Duplicate component** and **Trash** buttons in Viam app > Config > Components UI.

- Seems too simple. Is there anywhere else in the docs we talk about configuring components in the app that could use this notice?
- Drive-by alert indentation fix while in here.

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -174,11 +174,10 @@ Components of the same model are supported using the same low-level code.
   Many built-in components have convenient implicit dependencies, in which case `depends_on` can be left blank.
   For example, a [`gpio` motor](/components/motor/gpio/) depends on the `board` to which it is wired, but it has a dedicated `board` attribute and `viam-server` will automatically initialize that board before it looks for the motor.
 
-If you are configuring several similar components, you can use the **Duplicate component** button in the upper-right of a component's configuration pane to duplicate the existing configuration for that component.
-A new component with identical configuration settings will be created beneath your existing one, named `<existing-component>-copy`.
+If you are configuring several similar components, you can use the **Duplicate component** button in the upper-right of a component's configuration pane to create a new identical component beneath your existing one.
 Be sure to edit the duplicated component to change any parameters that are unique to the new component, such as its name.
 
-To delete a component, click the **trash** icon in the upper-right of the component configuration pane.
+To delete a component, click the trash can icon in the upper-right of the component configuration pane.
 
 For specific information on how to configure each supported component type, see the [components documentation](/components/).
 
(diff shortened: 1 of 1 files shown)


docs/manage/configuration.md
```diff
@@ -163,17 +163,23 @@ Components of the same model are supported using the same low-level code.
...
 
 - `depends_on`: Any components that a given component relies upon, and that must be initialized on boot before this component is initialized.
   Many built-in components have convenient implicit dependencies, in which case `depends_on` can be left blank.
   For example, a [`gpio` motor](/components/motor/gpio/) depends on the `board` to which it is wired, but it has a dedicated `board` attribute and `viam-server` will automatically initialize that board before it looks for the motor.
 
+If you are configuring several similar components, you can use the **Duplicate component** button in the upper-right of a component's configuration pane to duplicate the existing configuration for that component.
+A new component with identical configuration settings will be created beneath your existing one, named `<existing-component>-copy`.
+Be sure to edit the duplicated component to change any parameters that are unique to the new component, such as its name.
+
+To delete a component, click the **trash** icon in the upper-right of the component configuration pane.
```
JessamyT, line 181:
```suggestion
To delete a component, click the trash can icon in the upper-right of the component configuration pane.
```
Mine doesn't have text unless you hover, and then the text is **Delete component**, so could alternately change to that to match the above **Duplicate component** which is also just tooltip text

docs/manage/configuration.md
```diff
@@ -163,17 +163,23 @@ Components of the same model are supported using the same low-level code.
...
 
-{{% /alert %}}
+   {{% /alert %}}
 
 - `depends_on`: Any components that a given component relies upon, and that must be initialized on boot before this component is initialized.
   Many built-in components have convenient implicit dependencies, in which case `depends_on` can be left blank.
   For example, a [`gpio` motor](/components/motor/gpio/) depends on the `board` to which it is wired, but it has a dedicated `board` attribute and `viam-server` will automatically initialize that board before it looks for the motor.
 
+If you are configuring several similar components, you can use the **Duplicate component** button in the upper-right of a component's configuration pane to duplicate the existing configuration for that component.
+A new component with identical configuration settings will be created beneath your existing one, named `<existing-component>-copy`.
```
JessamyT, line 178:
```suggestion
If you are configuring several similar components, you can use the **Duplicate component** button in the upper-right of a component's configuration pane to create a new identical component beneath your existing one.
```
[opt] I think since this is fairly self-explanatory (and folks will use the app and quickly see what happens), can make this more concise.

JessamyT (approved):
LGTM % button name

npentrel (approved):

PR #1679, opened 2023-08-18
DOCS-1019: Add control view for servo

npentrel:


JessamyT (commented):
Here you have the "Test your servo" heading outside of the include, such that it appears in the right nav. In my motor/encoder PR I was following the camera way of doing it with "Test your..." in the include, and it doesn't show up in the right nav. Which way should we go?

npentrel:
should show up in the nav

PR #1680, opened 2023-08-18
DOCS-740: Document the Session Management API and SessionsClient

sguequierre:
* A start on the request above. Lots of documentation in RDK that was probably written a long time ago, and as far as I have been able to find the `Session Management API` is not fully implemented in the Python Client SDK (might be confused)


docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+When controlling a robot or fleet with the Viam platform, you want some kind of protocol or system-wide means to understand the presence of the clients that are talking and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis).
+
+The presence of these clients at one point in time can be referred to as a "session".
+A session technically is any presence mechanism at the application layer (i.e. RDK, not TCP) maintained by a client (e.g. SDK) with a server (e.g. RDK).
+Session management, as explained here, allows for safer robotic operation when you are dealing with moving parts on the robot (the "actuating controls").
+
+Specifically, without session management, controls that would be "sticky" (e.g. SetPower of a base) based on the last input of a **client**, meaning the API request from the client sets the flow of electricity on a part of the robot and then does not time out, could lead to **having a robot trying to continue doing what it was told to do forever.**
+
+You do not want that to happen.
```
sguequierre, line 27: this is kind of a weird sentence but want to stress that that particular scenario is not ideal?

sguequierre: also with the weird bold that's diff from normal docs usage, just kinda want to stress those parts-- but can remove

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Purpose of Session Management
+
+When controlling a robot or fleet with the Viam platform, you want some kind of protocol or system-wide means to understand the presence of the clients that are talking and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
+
+The presence of these clients at one point in time can be referred to as a "session".
+A session technically is any presence mechanism at the RDK application layer maintained by any client with `viam-server`.
+Session management, as explained here, allows for safer robotic operation when you are dealing with moving parts on the robot (the "actuating controls").
+
+Specifically, without session management, controls that would be "sticky" (e.g. SetPower of a base) based on the last input of a **client**, meaning the API request from the client sets the flow of electricity on a part of the robot and then does not time out, could lead to **having a robot trying to continue doing what it was told to do forever.**
```
sguequierre, line 25:
```suggestion
Specifically, without session management, controls that would be "sticky" (for example [`SetPower()` of a base](/components/base/#setpower)) based on the last input of a **client**, meaning the API request from the client sets the flow of electricity on a part of the robot and then does not time out, could lead to **having a robot trying to continue doing what it was told to do forever.**
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+A client has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's Session Management API is your built-in solution to this.
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the heartbeat window.
+
+Your client must send at least one session heartbeat within this window.
```
sguequierre, line 40: need to add more info about how to send a session heartbeat likely if this stays here on line 40

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+When controlling a robot or fleet with the Viam platform, you want some kind of protocol or system-wide means to understand the presence of the clients that are talking and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
+
+The presence of these clients at one point in time can be referred to as a "session".
+A session technically is any presence mechanism at the RDK application layer maintained by any client with `viam-server`.
+Session management, as explained here, allows for safer robotic operation when you are dealing with moving parts on the robot (the "actuating controls").
+
+Specifically, without session management, controls that would be "sticky" (for example [`SetPower()` of a base](/components/base/#setpower)) based on the last input of a **client**, meaning the API request from the client sets the flow of electricity on a part of the robot and then does not time out, could lead to **having a robot trying to continue doing what it was told to do forever.**
+
+You do not want that to happen.
```
JessamyT, line 27: Suggestion:
```suggestion
For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower)) command as the last input from a client before the connection to the robot is interrupted.
Without session management, the API request from the client sets the flow of electricity to the motors of the robot and then does not time out, causing the robot to continue driving forever, colliding with objects and people.

With session management, if the robot does not receive a signal from the client at regular intervals, it will safely stop until the connection is reestablished.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Clients and Viam's Client SDKs
+
+A client has multiple meanings for a Viam robot.
```
JessamyT, line 31:
```suggestion
*Client* has multiple meanings for a Viam robot.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+The presence of these clients at one point in time can be referred to as a "session".
+A session technically is any presence mechanism at the RDK application layer maintained by any client with `viam-server`.
+Session management, as explained here, allows for safer robotic operation when you are dealing with moving parts on the robot (the "actuating controls").
+
+Specifically, without session management, controls that would be "sticky" (for example [`SetPower()` of a base](/components/base/#setpower)) based on the last input of a **client**, meaning the API request from the client sets the flow of electricity on a part of the robot and then does not time out, could lead to **having a robot trying to continue doing what it was told to do forever.**
+
+You do not want that to happen.
+
+#### Clients and Viam's Client SDKs
```
JessamyT, line 29:
```suggestion
#### Clients and Viam's client SDKs
```
default to sentence case

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the heartbeat window.
+
+Your client must send at least one session heartbeat within this window.
+As soon as the window lapses/expires, the server will safely stop all resources that are marked for safety monitoring that have been last used by that session, and no others; a lapsed client will attempt to establish a new session immediately prior to the next operation it performs.
+
+{{< /alert >}}
+
+## Usage
+
+Usage of the Sessions Management API differs across [Viam's SDKS](/program/).
```
JessamyT, line 47:
```suggestion
Usage of the session management API differs across [Viam's SDKS](/program/).
```
be consistent with session or sessions

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "Sessions Management with Viam's Client SDKs"
```
JessamyT, line 2:
```suggestion
title: "Session Management with Viam's Client SDKs"
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+linkTitle: "Session Management API"
+weight: 20
+type: "docs"
+description: "How to use the Sessions Management API with Viam's Client SDKs."
```
JessamyT, line 6:
```suggestion
description: "How to use the session management API with Viam's Client SDKs."
```
Pick session or sessions and make all throughout here the same. Also headings -> sentence case, and lower case session management (or sessions management if you go that route)

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Clients and Viam's Client SDKs
+
+A client has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's Session Management API is your built-in solution to this.
```
JessamyT, line 37:
```suggestion
Viam's session management API is your built-in solution to this.
```
To what though? This sentence sounds like session management solves the problem of Viam having different meanings for "client," and I don't think that's the intent, but I don't know that I have a good enough understanding of the goal of this section to make a good suggestion.

Maybe this paragraph (just these two sentences) can be removed?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+tags: ["client", "sdk", "viam-server", "networking", "apis", "robot api", "session", "sessions", "session management"]
+---
+
+The Session Management API provides support for robot session management.
+
+A Session allows a **client** to express that it is actively connected or authenticated to `viam-server` on your robot and
+supports stopping actuating components when the connection is no longer active.
+
+{{< alert title="Info" color="info" >}}
+
```
JessamyT, line 16:
```suggestion
```
This doesn't need to be in an alert box; can just be an intro section.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's Session Management API is your built-in solution to this.
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the heartbeat window.
+
+Your client must send at least one session heartbeat within this window.
+As soon as the window lapses/expires, the server will safely stop all resources that are marked for safety monitoring that have been last used by that session, and no others; a lapsed client will attempt to establish a new session immediately prior to the next operation it performs.
+
+{{< /alert >}}
+
```
JessamyT, line 44:
```suggestion
```

docs/program/apis/sessions.md
JessamyT: Okay; I think there's a way to transition from the part about what happens without session management into this part nicely, but will wait to make suggestions until after this is done.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+You do not want that to happen.
+
+#### Clients and Viam's Client SDKs
+
+A client has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
```
JessamyT, line 35: I think this part about what a client is could go into an "Info" alert box if you want. It feels like a separate interjection for clarification, whereas everything else in this intro feels like intro body text I think.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+Your client must send at least one session heartbeat within this window.
+As soon as the window lapses/expires, the server will safely stop all resources that are marked for safety monitoring that have been last used by that session, and no others; a lapsed client will attempt to establish a new session immediately prior to the next operation it performs.
+
+{{< /alert >}}
+
+## Usage
+
+Usage of the Sessions Management API differs across [Viam's SDKS](/program/).
+
```
JessamyT, line 48:
```suggestion

### Manage on-robot sessions

```
Why not put your tabs in here, since it looks like this is the only part that differs between Go and Python, and then you'll get a handy heading in the right nav?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+{{< tabs >}}
+{{% tab name="Go" %}}
+
+### On-Robot Session Management
+
```
JessamyT, line 53:
```suggestion
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+To use the [Session Management API](https://pkg.go.dev/go.viam.com/rdk/session) the Go Client SDK provides, use your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API) to instantiate a `SessionsManager` with `SessionManager()`.
+
+### Sessions for RobotClients
+
+First, use your [`RobotClient()`](/program/apis/#robot-api) instance to access the [`SessionsClient`](https://pkg.go.dev/go.viam.com/rdk/session) within your Go Client SDK program.
+This is a [gRPC](https://grpc.io/) client that `viam-server` instantiates at robot runtime.
+Find `SessionsClient` defined on [Github](https://github.com/viamrobotics/rdk/blob/main/robot/client/client.go).
+
+{{% /tab %}}
```
JessamyT, line 62: I recommend putting just the on-robot session management content into the tabs, and taking the Sessions for robot clients part out of tabs. Agnosticize the sessions for robot clients part by changing "Go Client SDK" to just "client SDK"

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+{{% tab name="Python" %}}
+
+### On-Robot Session Management
+
+The `Session Management API` is not currently provided in the Python SDK.
+Use the Go Client SDK instead.
+
+### Sessions for RobotClients
```
JessamyT, line 70:
```suggestion
### Manage sessions for robot clients
```
or something more active

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+First, use your [`RobotClient()`](/program/apis/#robot-api) instance to access the [`SessionsClient`](https://python.viam.dev/autoapi/viam/sessions_client/index.html#viam.sessions_client.SessionsClient) within your Python Client SDK program.
+This is a [gRPC](https://grpc.io/) client that `viam-server` instantiates at robot runtime.
+Find `SessionsClient` defined on [Github](https://github.com/viamrobotics/rdk/blob/main/robot/client/client.go).
```
JessamyT, line 74:
```suggestion
Find `SessionsClient` defined on [GitHub](https://github.com/viamrobotics/rdk/blob/main/robot/client/client.go).
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+{{% /tab %}}
+{{% /tabs %}}
+
+### Disable the Sessions Client
```
JessamyT, line 79:
```suggestion
### Disable the sessions client
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `SessionsClient` is automatically enabled on your robot.
+If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's Client SDKs, access your `SessionsClient` instance with the `disabled` parameter set to false, as shown in the following Python code snippet:
```
JessamyT, line 82:
```suggestion
If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's client SDKs, access your `SessionsClient` instance with the `disabled` parameter set to `True`, as shown in the following Python code snippet:
```
True or false? The snippet says True

JessamyT (changes requested):
Thanks for taking this on!

docs/program/apis/sessions.md
sguequierre: Sounds good to me 😆, thank you!

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "Sessions Management with Viam's Client SDKs"
+linkTitle: "Session Management API"
+weight: 20
+type: "docs"
+description: "How to use the Sessions Management API with Viam's Client SDKs."
+tags: ["client", "sdk", "viam-server", "networking", "apis", "robot api", "session", "sessions", "session management"]
+---
+
+The Session Management API provides support for robot session management.
```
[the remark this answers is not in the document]
sguequierre: I think you're right, adding

docs/program/apis/sessions.md
sguequierre: done

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+A Session allows a **client** to express that it is actively connected or authenticated to `viam-server` on your robot and
+supports stopping actuating components when the connection is no longer active.
```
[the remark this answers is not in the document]
sguequierre: I think the point is that you can vs that it automatically does that? I could be confused but I think first one is better- thank you for the suggestion!

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+{{< alert title="Info" color="info" >}}
+
+### Purpose of Session Management
+
+When controlling a robot or fleet with the Viam platform, you want some kind of protocol or system-wide means to understand the presence of the clients that are talking and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
```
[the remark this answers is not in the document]
sguequierre: I agree, first reword might be better to show off fleet management

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+
+The presence of these clients at one point in time can be referred to as a "session".
+A session technically is any presence mechanism at the RDK application layer maintained by any client with `viam-server`.
+Session management, as explained here, allows for safer robotic operation when you are dealing with moving parts on the robot (the "actuating controls").
+
```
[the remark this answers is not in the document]
sguequierre: done

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+A Session allows a **client** to express that it is actively connected or authenticated to `viam-server` on your robot and
+supports stopping actuating components when the connection is no longer active.
+
+{{< alert title="Info" color="info" >}}
+
+### Purpose of Session Management
+
+When controlling a robot or fleet with the Viam platform, you want some kind of protocol or system-wide means to understand the presence of the clients that are talking and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
+
+The presence of these clients at one point in time can be referred to as a "session".
```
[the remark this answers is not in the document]
sguequierre: Adding! I think it's hard to get conceptually and hard to phrase. Using _is_ is definitely better, though

docs/program/apis/sessions.md
sguequierre: Done

docs/program/apis/sessions.md
sguequierre: done, thank you

docs/program/apis/sessions.md
sguequierre: done, thank you

docs/program/apis/sessions.md
sguequierre: done

docs/program/apis/sessions.md
sguequierre: done

docs/program/apis/sessions.md
sguequierre: done

docs/program/apis/sessions.md
sguequierre: done/remove info along with top

docs/program/apis/sessions.md
sguequierre: good idea!!

docs/program/apis/sessions.md
sguequierre: done

sguequierre: Unsure about removing `RobotClient` but I think "robot clients" is probably fine

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,86 @@
...
+## Usage
+
+Usage of the session management API differs across [Viam's SDKS](/program/).
+
+### Manage on-robot sessions
+
+{{< tabs >}}
+{{% tab name="Go" %}}
+
+To use the [Session Management API](https://pkg.go.dev/go.viam.com/rdk/session) the Go Client SDK provides, use your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API) to instantiate a `SessionsManager` with `SessionManager()`.
```
maximpertsov, line 52: No need to instantiate a separate `SessionsManager` client - one will be instantiated when you start a robot client. Instead, you should disable sessions by passing the following option to the robot client:

```go
robot, err := client.New(ctx, "my-robot-address", logger, client.WithDisableSessions(), ...)
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,86 @@
...
+#### Clients and Viam's client SDKs
+
+*Client* has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API is your built-in solution to this.
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the heartbeat window.
```
maximpertsov, line 38: You can adjust the heartbeat window via the robot config (needs to be done through raw JSON at the moment):

```json
  ...
  "network": {
    "sessions": {
      "heartbeat_window": "30s"
    }
  },
  ...
```

The default heartbeat window is 2 seconds if this configuration is omitted.

docs/program/apis/sessions.md
maximpertsov: Heartbeats are sent automatically from our clients (golang, python, and typescript) unless disabled via a client option or session management is not implemented by the server. 

As of now, heartbeats are sent at an interval that is 1/5th of the heartbeat window (e.g. if the heartbeat window is 5 seconds, clients will send a heartbeat every 1 second).

maximpertsov (dismissed):
Made some changes to how users should disable sessions.

We should also add a section for typescript client if it doesn't exist already. Sessions are enabled by default as with the other clients, and can be disabled via:


```ts
  const robot = await VIAM.createRobotClient({
    // ...
    disableSessions: true
    // ...
  });
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,86 @@
...
+### Disable the sessions client
+
+The `SessionsClient` is automatically enabled on your robot.
+If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's client SDKs, access your `SessionsClient` instance with the `disabled` parameter set to `True`, as shown in the following Python code snippet:
+
+```python {class="line-numbers linkable-line-numbers"}
+async def main():
+    robot = await connect()
+    sessions = SessionsClient.(channel: Channel, disabled: bool = True)
+```
```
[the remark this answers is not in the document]
sguequierre: Awesome, thank you Max! Will add and comment once addressed

docs/program/apis/sessions.md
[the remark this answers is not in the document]
JessamyT: Yes--I think the first one makes it sound like you can but it doesn't automatically do that, so it's a little ambiguous but maybe that's correct?

docs/program/apis/sessions.md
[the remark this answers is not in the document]
JessamyT: "When controlling a robot or fleet with Viam" is kinda implied/obvious though so should probably be deleted

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,86 @@
...
+A session:
+
+- allows a client to express that it is actively connected or authenticated to `viam-server` on your robot
+- supports stopping moving components when the connection is no longer active.
+
+### Purpose of session management
+
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
+
+The presence of these clients at one point in time is called a *session*.
```
JessamyT, line 21: Per my comment on the last go 'round, is this right? I think it's actually a period of time, not the presence of something at a point in time, right?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,86 @@
...
+A session technically is any presence mechanism at the RDK application layer maintained by any client with `viam-server`.
```
JessamyT, line 22: Same here--I think it's true that session management has a whole lot to do with mechanisms for monitoring presence, but session =/= presence mechanism.

docs/program/apis/sessions.md
JessamyT: @sguequierre thoughts on this rework?

docs/program/apis/sessions.md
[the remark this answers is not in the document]
JessamyT: Yeah it's tricky, and I Googled it for a bit before making that comment^^ to see if there are other ways of conceptualizing and/or phrasing it. Everyone seems to have a hard time putting it into words :D But I think it can be improved. In this case, is it indeed more of a period of time?

docs/program/apis/sessions.md
JessamyT: @sguequierre This comment is still unaddressed

docs/program/apis/sessions.md
JessamyT: Either way

docs/program/apis/sessions.md
sguequierre, line 21:
```suggestion
The period of time during which these clients are present is called a *session*.
```

sguequierre: Great point-- this was weird phrasing I copied-- LMK if you think updated phrasing is more sensible

docs/program/apis/sessions.md
sguequierre: This way of phrasing it is really good/a lot easier to understand, sorry @ delay -- I was leaving this up to read through properly during last pass 😆 will implement now

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,87 @@
...
+{{% /tabs %}}
+
+### Disable the sessions client
+
+The `SessionsClient` is automatically enabled on your robot.
+If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's client SDKs, access your `SessionsClient` instance with the `disabled` parameter set to `True`, as shown in the following Python code snippet:
+
+```python {class="line-numbers linkable-line-numbers"}
+async def main():
+    robot = await connect()
```
sguequierre, line 85:
```suggestion
    opts = RobotClient.Options(disable_sessions=True, ...)
    await RobotClient.at_address("my-robot-address", opts)
    robot = await connect()
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,87 @@
...
+    sessions = SessionsClient.(channel: Channel, disabled: bool = True)
```
sguequierre, line 86:
```suggestion
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,87 @@
...
+## Usage
+
+Usage of the session management API differs across [Viam's SDKS](/program/).
+
+### Manage on-robot sessions
+
+{{< tabs >}}
+{{% tab name="Go" %}}
+
+To use the [Session Management API](https://pkg.go.dev/go.viam.com/rdk/session) the Go Client SDK provides, use your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API) to instantiate a `SessionsManager` with `SessionManager()`.
```
sguequierre, line 53:
```suggestion
To use the [Session Management API](https://pkg.go.dev/go.viam.com/rdk/session) the Go Client SDK provides, use your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API).
Pass the following option to the robot client:
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,87 @@
...
+### Disable the sessions client
+
+The `SessionsClient` is automatically enabled on your robot.
+If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's client SDKs, access your `SessionsClient` instance with the `disabled` parameter set to `True`, as shown in the following Python code snippet:
+
+```python {class="line-numbers linkable-line-numbers"}
+async def main():
+    robot = await connect()
+    sessions = SessionsClient.(channel: Channel, disabled: bool = True)
+```
```
sguequierre, line 87:
```suggestion
```

```

File: docs/program/apis/sessions.md [hunk 35]
```diff
@@ -0,0 +1,87 @@
...
+First, use your [`RobotClient()`](/program/apis/#robot-api) instance to access the [`SessionsClient`](https://python.viam.dev/autoapi/viam/sessions_client/index.html#viam.sessions_client.SessionsClient) within your Python Client SDK program.
+This is a [gRPC](https://grpc.io/) client that `viam-server` instantiates at robot runtime.
+Find `SessionsClient` defined on [GitHub](https://github.com/viamrobotics/rdk/blob/main/robot/client/client.go).
+
+{{% /tab %}}
+{{% /tabs %}}
+
+### Disable the sessions client
+
+The `SessionsClient` is automatically enabled on your robot.
```
sguequierre, line 80:
```suggestion
The `SessionsClient` is automatically enabled on your robot.
It is instantiated as part of your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API).
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,88 @@
...
+It is instantiated as part of your [`RobotClient`](/program/apis/#robot-api) instance (client of the Robot API).
+If you want to disable it to keep any additional clients from authenticating to your robot's session while running a control program with Viam's client SDKs, you can pass the option to your robot, as shown in the following code snippets:
+
+```python {class="line-numbers linkable-line-numbers"}
+async def main():
+    opts = RobotClient.Options(disable_sessions=True, ...)
+    await RobotClient.at_address("my-robot-address", opts)
+    robot = await connect()
```
sguequierre, line 88:
```suggestion
    robot = await connect()
```

```go {class="line-numbers linkable-line-numbers"}

```
```

File: docs/program/apis/sessions.md [hunk 37]
```diff
@@ -0,0 +1,104 @@
...
+```
+
+{{% /tab %}}
+{{% tab name="Go" %}}
+
+```go {class="line-numbers linkable-line-numbers"}
```
sguequierre, line 97:
```suggestion
```go {class="line-numbers linkable-line-numbers"}
robot, err := client.New(ctx, "my-robot-address", logger, client.WithDisableSessions(), ...)
```

docs/program/apis/sessions.md
sguequierre: I had tried to adapt this part but the old docs wording is wild sometimes 😆 i think it's a byproduct of having engineers write docs bc what's technically true might not be particularly easy to understand

JessamyT: Yeah I was cross referencing that doc with this as I made my first review pass. We are under no obligation to keep any wording that doesn't make sense :D

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,105 @@
...
+The session management API provides support for robot session management.
+
+A session:
+
+- allows a client to express that it is actively connected or authenticated to `viam-server` on your robot
+- supports stopping moving components when the connection is no longer active.
+
+### Purpose of session management
+
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server` through Viam's various [resource APIs](/program/apis/).
```
JessamyT, line 19:
```suggestion
When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
```
I wonder would this actually be more clear?

docs/program/apis/sessions.md
JessamyT: This sentence is ambiguous with respect to whether it's maintained between a client and `viam-server`, or whether it's maintained, by a client that has `viam-server`.

How about
```suggestion
*Session management* is a presence mechanism between a client and `viam-server`.
```

docs/program/apis/sessions.md
sguequierre: Yeah, I think either one is fine technically but this change is simpler so lgtm-- implementing

docs/program/apis/sessions.md
sguequierre: LGTM adding

docs/program/apis/sessions.md
sguequierre: @JessamyT What would you think about me making this part a separate page/a second-half ticket of this? wondering if could merge as is to get this info up and then add a lot more about heartbeats, but maybe i should just do it now lol

docs/program/apis/sessions.md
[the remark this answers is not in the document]
sguequierre: done

docs/program/apis/sessions.md
sguequierre: done/changed

docs/program/apis/sessions.md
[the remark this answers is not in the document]
sguequierre: Retroactive oops, I have not added yet for typescript but I think your comment with that may be connected to this comment thread? Adding now

docs/program/apis/sessions.md
JessamyT: I don't think it needs to be a whole second page? Maybe just incorporate it now? I think some of this stuff can be kinda combined/made smooth and succinct, and the basics of heartbeats are sorta part of how the whole thing works so it's important to at least have high level info about that.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,117 @@
...
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
+
+The period of time during which these clients are present is called a *session*.
+*Session management* is a presence mechanism between a client and `viam-server`.
+
+Session management allows for safer operation of robots that physically move.
+For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower) command as the last input from a client before the connection to the robot is interrupted.
+Without session management, the API request from the client sets the flow of electricity to the motors of the robot and then does not time out, causing the robot to continue driving forever, colliding with objects and people.
+
+With session management, if the robot does not receive a signal from the client at regular intervals, it will safely stop until the connection is reestablished.
```
JessamyT, line 28:
```suggestion
With session management, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.
```
I think this passive voice was my fault to begin with but catching it now haha

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Clients and Viam's client SDKs
```
JessamyT, line 30: I find this title kinda odd because a client SDK can act as a client right? So it's kinda like "Rectangles and squares"? And then it goes on to talk about heartbeats too?

What would you think of titling this section "How session management works" or "Clients and heartbeats" or something?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,117 @@
...
+
+*Client* has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API is your built-in solution to this.
```
JessamyT, line 38: As I said on a previous review, "this" is ambiguous. Solution to what? Grammatically, it is saying that it's the solution to the fact that "client" means multiple things. Again, I still think this sentence should be removed/replaced.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,117 @@
...
+
+With session management, if the robot does not receive a signal from the client at regular intervals, it will safely stop until the connection is reestablished.
+
+#### Clients and Viam's client SDKs
+
+*Client* has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
```
JessamyT, line 36:
```suggestion
For example, if you [add modular resources to your robot](/extend/modular-resources/), the clients of your robot include the "module servers" you instantiate on your robot for individual resources.
```
Can shorten. I'm less sure about module versus model, but it seems like module makes more sense?

docs/program/apis/sessions.md
JessamyT: Would it be correct to say
```suggestion
Each client maintains a session, telling the `viam-server` instance that it is still present every so often, or staying within the heartbeat window.
```
The page just talked about how there are multiple meanings of client. So, do all of them maintain sessions or just one?

docs/program/apis/sessions.md
JessamyT: But if you think there need to be lots of details added, sure, can be a separate PR to modify this page.

docs/program/apis/sessions.md
maximpertsov: Just one

e.g. the python sdk maintains one session - it does *not* create a new session for each component/service client it manages

docs/program/apis/sessions.md
sguequierre: yeah i just want to get this part up first tbh but i can find time to add this in 🤷‍♀️ That makes sense that it's fundamental

sguequierre: Will comment here when done

sguequierre: Ok @maximpertsov @JessamyT I think I have added the necessary info for users for client heartbeats now, sorry for delay here

docs/program/apis/sessions.md
sguequierre: A little confused here-- why do you recommend?

docs/program/apis/sessions.md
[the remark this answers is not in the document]
sguequierre: done!

docs/program/apis/sessions.md
sguequierre: ^ I think this is consistent with current proposed docs so no changed needed?

docs/program/apis/sessions.md
sguequierre: re. this I added a `heartbeats` section~~ see thread closer to top for more ref. info!

docs/program/apis/sessions.md
sguequierre: This should be kept model to be consistent with the other docs. I know it might seem like module makes more sense but the module refers to the mechanism and the model refers to the actual plans for design-- think of architecture, the docking mechanism vs the dock that is built.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,143 @@
...
+
+Viam's session management API is your built-in solution to manage this.
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the **heartbeat** window.
+
+#### Heartbeats
+
+A **heartbeat** is a signal that indicates robot connectivity.
+Essentially, "heartbeats" are a Viam robot's way of letting a user reading data from it know the different parts of it are "alive."
+
+Heartbeats are sent automatically from our clients (golang, python, and typescript) unless disabled via a client option or session management is not implemented by the server.
```
sguequierre, line 46:
```suggestion
Heartbeats are sent automatically from our clients (golang, python, and typescript) unless you disable this with the session management API, or session management is not implemented by the server.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,143 @@
...
+
+As of now, heartbeats are sent at an interval that is 1/5th of the heartbeat window.
```
sguequierre, line 48:
```suggestion
As of now, heartbeats are sent at an interval that is one fifth of the heartbeat window.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,143 @@
...
+For example, if the heartbeat window is 5 seconds, clients will each send a heartbeat every 1 second.
+
+You can adjust the heartbeat window via the robot config.
```
sguequierre, line 51:
```suggestion
You can adjust the heartbeat window through the configuration of your robot.
```

docs/program/apis/sessions.md
JessamyT: How about 
```suggestion
For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "modular resource servers" you instantiate on your robot for individual resources.
```
then?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Clients and Viam's client SDKs
+
+*Client* has multiple meanings for a Viam robot.
+Essentially, it's anything that is receiving the information served by `viam-server` running on the robot.
+
+A **client** can be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API is your built-in solution to manage this.
```
JessamyT, line 38:
```suggestion
```
I realize you've added the word "manage" but "this" is still ambiguous and I still don't think this is a useful sentence. What is it meant to convey?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,143 @@
...
+Your client maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the **heartbeat** window.
```
JessamyT, line 39: As mentioned previously, you just talked about how a client can be lots of different things, so is it the case that each client maintains a separate session with `viam-server` or just one client does, even in the case of a robot with many parts? That needs to be clearer here.

docs/program/apis/sessions.md
JessamyT: The "Manage sessions for robot clients" and "Sessions for RobotClients" content for Python and Go is identical except for the mention of Go vs Python. So instead of having them as tabbed content, take that out of the tabs. Move it to one "Manage sessions for robot clients" heading with no tabs.

JessamyT: Keep the "Manage sessions on-robot" heading and put tabs for python and go under that since that part *is* different between the two languages (since it's unsupported for Python). LMK if you wanna jump in a video call if still confused about what I mean.

JessamyT: Keep the "Manage sessions on-robot" heading and put tabs for python and go under that since that part *is* different between the two languages (since it's unsupported for Python). LMK if you wanna jump in a video call if still confused about what I mean.

docs/program/apis/sessions.md
sguequierre: "this" is the problem i'm talking about beforehand!

JessamyT: The problem being that client can mean lots of different things? That's what this is currently implying, but I don't think that's what it's meant to mean.

sguequierre: client can mean lots of different things, yes, but there is only one session in question for the session management api. The session is what's being discussed, not the client

sguequierre: I'll try and address your feedback but I'm sorry bc I too am getting confused 😆

sguequierre: @JessamyT  would you suggest the specific noun be maybe something like "this dissonance in your session"?

JessamyT: Then what is the actual problem this solves? This section (under the client heading) just talks about how "client" can mean various things. Then it says "here's the solution to this problem," but what is the problem?

JessamyT: I don't think I understand what "dissonance in your session" actually means, but yeah something more specific than "this" is what I'm advocating for! Could talk synchronously about this maybe :D I'll slack you

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+A session:
+
+- allows a client to express that it is actively connected or authenticated to `viam-server` on your robot
+- supports stopping moving components when the connection is no longer active.
+
+### Purpose of session management
+
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
+The period of time during which these clients are present is called a *session*.
+Imagine a session as analogous to the lifetime of a robot, in regards to its network connection.
```
JessamyT, line 22:
```suggestion
```
I think this is confusing and doesn't add, because when I think of a "lifetime of a robot" I think of its entire lifetime, not just a session.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A **client** within the ecosystem of Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API is your built-in solution to manage this.
+Your client of the session management API can maintain the session, telling the `viam-server` instance that it is still present every so often, or staying within the **heartbeat** window.
+
+#### Heartbeats
+
+A **heartbeat** is a signal that indicates robot connectivity.
```
JessamyT, line 39:
```suggestion
A *heartbeat* is a signal that indicates robot connectivity.
```
Use italics for new terms

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+*Session management* is a presence mechanism between a client and `viam-server`, or, a way to manage the lifetime of the robot yourself.
+
+Session management allows for safer operation of robots that physically move.
+For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower) command as the last input from a client before the connection to the robot is interrupted.
+Without session management, the API request from the client sets the flow of electricity to the motors of the robot and then does not time out, causing the robot to continue driving forever, colliding with objects and people.
+
+With session management, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A **client** within the ecosystem of Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
```
JessamyT, line 31:
```suggestion
A *client* of a Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
+The period of time during which these clients are present is called a *session*.
+Imagine a session as analogous to the lifetime of a robot, in regards to its network connection.
+*Session management* is a presence mechanism between a client and `viam-server`, or, a way to manage the lifetime of the robot yourself.
+
+Session management allows for safer operation of robots that physically move.
+For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower) command as the last input from a client before the connection to the robot is interrupted.
+Without session management, the API request from the client sets the flow of electricity to the motors of the robot and then does not time out, causing the robot to continue driving forever, colliding with objects and people.
+
+With session management, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.
```
JessamyT, line 29:
```suggestion
With session management, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.

```
This is a separate idea from the next few sentences because those are talking about how client can be various things but this isn't really about that even though it has the word "client" in it.

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A **client** within the ecosystem of Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot in its lifetime will include the "model servers" you instantiate on your robot for individual resources.
```
JessamyT, line 32:
```suggestion
For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot will include the "model servers" you instantiate on your robot for individual resources.
```
Edit like ^ if you think this really must stay in. But I think just delete this, because the sentence before provides examples.
```suggestion
```

docs/program/apis/sessions.md
JessamyT:
```suggestion
Viam's session management API provides this functionality.
```
Okay, so, this will make sense if you move the "what is a client" type stuff (cut down to 2 sentences I think) into an info box before the purpose stuff. Then the "why" will flow directly into the "how".

docs/program/apis/sessions.md
sguequierre: it'll give me a fail on the check, already tried that unfortunately. could we just keep this bold like is the case for client?

docs/program/apis/sessions.md
sguequierre: I think this is helpful-- implementing first suggestion but not removing

docs/program/apis/sessions.md
sguequierre: Ok cool done!

docs/program/apis/sessions.md
sguequierre: I think it makes sense as-is, is that ok? the this in question is the abundance of clients i'm discussing, then i go into the how of how it works.

sguequierre: as-is after implementing your other suggestions!! @JessamyT

docs/program/apis/sessions.md
JessamyT: In what way is it failing checks? Is it saying it's inconsistent? in which case either try underscores instead or make sure there aren't any underscores-as-italics elsewhere in here

JessamyT: And client shouldn't be bold either :)

docs/program/apis/sessions.md
JessamyT: But again, I thought the problem is *not* the abundance of clients, but rather the potential for disconnection without timeout?

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+With session management, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.
+
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A *client* of a Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API is your built-in solution to manage this.
+Your client of the session management API maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the **heartbeat** window.
```
JessamyT, line 36:
```suggestion
Your client of the session management API maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the *heartbeat* window.
```

docs/program/apis/sessions.md
sguequierre: I am thinking of https://docs.viam.com/extend/modular-resources/create/ info and other suggestions recently-- I have had both Naomi and Andrew suggest I use bold for this purpose. Is that not ok?

sguequierre: Just confused. I know we've gone back and forth a lot on what we are using bold and italic for in terms of emphasis-- If we aren't doing **bold** for importance at all then I suppose more docs need to be changed

sguequierre: but I would really prefer to just keep it and land this one so i can expand the usage section

docs/program/apis/sessions.md
sguequierre: The problem that's being solved here is wanting to turn off Automated Session Management (eric's session management client), ao that all calls to/through that API you'll now make yourself instead of viam/cloud run doing that for you. it's really an advanced use case

docs/program/apis/sessions.md
sguequierre: You might imagine the lifetime of a robot differently, and that makes sense, but I still find it helpful to imagine a session in that way! is it ok if I leave this up or do you find it completely nonsensical? I think this whole thing will be a lot easier for users to understand w the clients ecosystem diagram thing

docs/program/apis/sessions.md
sguequierre: ok done! not worth arguing about

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+The session management API provides support for robot session management.
+With default recourse configuration, sessions are automatically managed for you with Viam's `SessionsClient`, but if you want to manage sessions yourself, you can with Viam's sessions management API.
+
+### The `SessionsClient`
+
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A *client* of a Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API's `SessionsClient` is a built-in solution that manages the connection between your robot's clients and your robot.
```
sguequierre, line 33: @JessamyT I updated this part now to try and make the function of the (eric's) automated sessions client easier to understand! LMK what you think

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+linkTitle: "Session Management API"
+weight: 20
+type: "docs"
+description: "How to use the session management API with Viam's Client SDKs."
+tags: ["client", "sdk", "viam-server", "networking", "apis", "robot api", "session", "sessions", "session management"]
+---
+
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
+The period of time during which these clients are present is called a *session*.
+Imagine a session as analogous to the lifetime of a robot in regards to its network connection.
```
JessamyT, line 12:
```suggestion
```
see other thread on last go 'round

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+
+A session:
```
JessamyT, line 14:
```suggestion
Session management:
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+
+- allows a client to express that it is actively connected or authenticated to `viam-server` on your robot
+- supports stopping moving components when the connection is no longer active.
+
+*Session management* is a presence mechanism between a client and `viam-server`, or, a way to manage the lifetime of the robot yourself.
```
JessamyT, line 19:
```suggestion
*Session management* is a presence mechanism between a client and `viam-server`.
```

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+Session management allows for safer operation of robots that physically move.
+For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower) command as the last input from a client before the connection to the robot is interrupted.
+Without session management, the API request from the client sets the flow of electricity to the motors of the robot and then does not time out, causing the robot to continue driving forever, colliding with objects and people.
+
+The session management API provides support for robot session management.
+With default recourse configuration, sessions are automatically managed for you with Viam's `SessionsClient`, but if you want to manage sessions yourself, you can with Viam's sessions management API.
```
JessamyT, line 25:
```suggestion
With default configuration, sessions are automatically managed for you with Viam's `SessionsClient`.
If you want to manage sessions yourself, use Viam's sessions management API.
```

docs/program/apis/sessions.md
JessamyT: Yeah I think this helps a lot

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,133 @@
...
+
+### The `SessionsClient`
+
+A Viam robot has many clients because a client is anything that is receiving the information served by `viam-server` running on the robot, which includes the primary part, sub-parts, client SDKs, and more.
+A *client* of a Viam robot could be an SDK script controlling the robot, an input controller, or just the different resources on the robot talking amongst themselves.
+For example, if you use Viam's module registry to [add modular resources to your robot](/extend/modular-resources/), the clients of your robot will include the "model servers" you instantiate on your robot for individual resources.
+
+Viam's session management API's `SessionsClient` is a built-in solution that manages the connection between your robot's clients and your robot.
+With the `SessionsClient`, if the robot does not receive a signal from the client at regular intervals, it safely stops until the connection is reestablished.
+Your client of the session management API maintains the session, telling the `viam-server` instance that it is still present every so often, or staying within the *heartbeat* window.
```
JessamyT, line 35:
```suggestion
Your client of the session management API maintains the session, telling the `viam-server` instance that it is still present every so often; in other words, staying within the *heartbeat* window.
```
"or" makes it sound like an either or such that it's ambiguous as to whether "telling the ...." is equivalent to or an option that is different from "staying within...."

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,132 @@
...
+weight: 20
+type: "docs"
+description: "How to use the session management API with Viam's Client SDKs."
+tags: ["client", "sdk", "viam-server", "networking", "apis", "robot api", "session", "sessions", "session management"]
+---
+
+When controlling a robot or fleet with Viam, you want a way to understand the presence of the clients that are communicating with and authenticated to each robot's `viam-server`.
+The period of time during which these clients are present is called a *session*.
+
+Session management:
```
sguequierre, line 13: @JessamyT Makes way more sense subtitling this as session management, thank you

docs/program/apis/sessions.md
```diff
@@ -0,0 +1,132 @@
...
+```ts {class="line-numbers linkable-line-numbers"}
+const robot = await VIAM.createRobotClient({
+// ...
+disableSessions: true
+// ...
+});
+```
+
+{{% /tab %}}
+
```
sguequierre, line 127:
```suggestion
```

docs/program/apis/sessions.md
JessamyT: Per offline convo yeah idk about emphasis either I just know about new terms!

JessamyT (approved):
I think this is good! Yay, thanks for figuring it out with me :D And I think once the diagram is sorted out it'll make sense to add back in sentences about what a session is analogous to, that relate it to the diagram.

PR #1681, opened 2023-08-18
DOCS-1009: Remove `camera_mono` movement sensor experimental model

sguequierre:
* Taking this model out of RDK as it is being removed 

keeping page with `draft: true` for now in case we want that info in the future for `mono-camera` module? can remove


andf-viam (approved):
LGTM, thanks!

randhid (approved):

PR #1693, opened 2023-08-22
DOCS-876: Use instantsearch and typesense for tutorials page

npentrel:
- Add filtering options to tutorials page
- Add GitHub action to upsert tutorials collection on typesense on push
- Improve user experience for folks that don't use javascript
- Order tutorials by featured value (1-8) and then by date (newest first)

https://docs-test.viam.dev/21dbe0b2507868d3e2054a9a9639ae22a7751641/public/tutorials/


layouts/partials/sidebar-tree.html
```diff
@@ -51,6 +51,8 @@
 {{ $manualLink := cond (isset $s.Params "manuallink") $s.Params.manualLink ( cond (isset $s.Params "manuallinkrelref") (relref $s $s.Params.manualLinkRelref) $s.RelPermalink) -}}
 {{ $manualLinkTitle := cond (isset $s.Params "manuallinktitle") $s.Params.manualLinkTitle $s.Title -}}
 {{ $empty_node := $s.Params.empty_node -}}
+
+{{ warnf "%s %s %s" $s $s.RelPermalink $s.Permalink }}
```
npentrel, line 55:
```suggestion
```

layouts/partials/sidebar-tree.html
```diff
@@ -80,7 +82,7 @@
       </a>
   {{ end }}
   {{- end }}
-  {{- if $withChild }}
+  {{- if (and $withChild (ne $manualLink "/tutorials/" ) ) }}
```
npentrel, line 85:
```suggestion
  {{- if (and $withChild (ne $s.Permalink "/tutorials/" ) ) }}
```

layouts/partials/sidebar-tree.html
```diff
@@ -66,12 +67,12 @@
         <a href="{{ $manualLink }}"{{ if ne $s.LinkTitle $manualLinkTitle }} title="{{ $manualLinkTitle }}"{{ end }}{{ with $s.Params.manualLinkTarget }} target="{{ . }}"{{ if eq . "_blank" }} rel="noopener"{{ end }}{{ end }} class="{{ if $active}} active{{ end }}{{ if $treeRoot }} tree-root{{ end }}">{{ $s.LinkTitle }}
         </a>
       {{- end }}
-      {{- if (and $withChild (not $treeRoot) ) }}
+      {{- if (and (and $withChild (not $treeRoot) (ne $s.Permalink "/tutorials/" ) ) ) }}
```
npentrel, line 70:
```suggestion
{{ warnf "%s %s %s" .Section $s.Permalink $s.RelPermalink }}
      {{- if (and (and $withChild (not $treeRoot) (ne $s.Permalink "/tutorials/" ) ) ) }}
```

erh (approved):

mcvella (approved):
LGTM

docs/tutorials/services/color-detection-scuttle.md
```diff
@@ -16,8 +16,8 @@ authors: [ "Hazal Mestci" ]
 languages: [ "python" ]
 viamresources: [ "base", "vision", "camera" ]
 level: "Beginner"
-date: "18 August 2022"
-updated: "11 August 2023"
+date: "2022-08-18"
```
sguequierre, line 19: (opt) I don't want to block bc I don't think it matters too much but I will say american ppl might get confused by this date format 
```suggestion
date: "08-18-2022"
```

docs/tutorials/services/data-management-tutorial.md
```diff
@@ -16,9 +16,11 @@ authors: []
 languages: []
 viamresources: [ "data_manager", "camera" ]
 level: "Beginner"
-date: "8 February 2023"
+date: "2023-02-08"
+# updated: ""
 cost: "0"
 no_list: true
+featured: 4
```
sguequierre, line 23: This is how much we've featured it on socials?

docs/tutorials/template.md
```diff
@@ -15,13 +15,16 @@ tags: ["tutorial"]
 draft: true # Change this when you're ready
 authors: [] # Your Name
```
sguequierre, line 16: Would add "featured" to template here

static/js/tutorials.js
```diff
@@ -0,0 +1,118 @@
...
+      {
+        host: "cgnvrk0xwyj9576lp-1.a1.typesense.net",
+        port: "443",
+        protocol: "https",
+      },
+    ],
+    cacheSearchResultsForSeconds: 2 * 60, // Cache search results from server. Defaults to 2 minutes. Set to 0 to disable caching.
+  },
+  // The following parameters are directly passed to Typesense's search API endpoint.
+  //  So you can pass any parameters supported by the search endpoint below.
```
sguequierre, line 17: This seems useful but I would suggest having more info here if this tip will be important for understanding search functionality in the future - is this endpoint what people are using/querying in the search bar of docs? and code here is for that?

assets/scss/_styles_project.scss
```diff
@@ -1130,3 +1131,105 @@ ul.sectionlist > li:hover {
...
+}
+
+#hits .ais-Hits-list {
+  list-style-type: none;
+  display: flex;
+  flex-wrap: wrap;
+  width: 100%;
+}
+
+.ais-Pagination-item {
```
sguequierre, line 1152: (opt) If you know how to toggle button size on square buttons, it seems like the buttons might be smaller than they appear. 

I don't know if this is the right object (here where I'm commenting) but I did notice at least in staging a lot of the buttons in the tutorials index/nav bar don't work for me if I try to click them, at first, they only have a small surface area. Tiny issue and non blocking, but if you could make the buttons _slightly bigger_ it would be completely unnoticeable. 

 Only the buttons that look like `<<` and `>>` are easily clickable, the rest I have to hover for a bit. Not sure how to fix, but might be a browser related issue on my end? 
 
 This bar is what I'm talking about:
Screenshot 2023-08-31 at 10 32 44 AM

sguequierre (commented):
LGTM no issues -- a couple comments with suggestions in case you want to implement

docs/tutorials/services/color-detection-scuttle.md
npentrel: I think it's not just american people, but in this case this is how Go expects the date to be formatted, so it is what it is

docs/tutorials/template.md
npentrel: done

static/js/tutorials.js
npentrel: the search bar works similar but with a different api key and different parameters. It's two different collections that the data for searching is stored in. This code dynamically adds the search sidebar and the tutorial cards to the page by accessing the collection and searching for tutorials that fit the search criteria and then adding them to the page

JessamyT:
- I agree that left nav should still be available, so I like the idea of filters up top instead.
- I worry that, since almost all the tutorials are currently categorized as beginner level, folks won't actually find super simple starter tutorials as easily as they would with the old format.
  - Maybe we could have a super obvious "getting started" area somewhere, either here or elsewhere, to make up for it?
- The cost bars spill into the page on my default screen width: 
image
(which I suppose would be rectified if filters moved to the top)

PR #1699, opened 2023-08-25
DOCS-1017: Update encoder docs with new fancy config

JessamyT:


npentrel (approved):
Remember to make the window size a bit smaller where possible before taking the screenshots. These are fine but and not worth redoing but could've been slightly less wise to make it more readable on mobile :)

JessamyT:
Hmm so, often we show these kinds of screenshots with the tabs included. Though we don't show the tabs here, I thought it would be good to keep the width consistent with other pages, so I chose basically the minimum width such that the tabs (such as control, config etc) don't have to wrap. I guess that type of consistency doesn't matter in this context though so I can make them narrower next time.

PR #1704, opened 2023-08-25
[DOCS-1047][DOCS-DOCS-1053]Odrive Renaming and Movement sensor units

randhid:
# Description

Sorry if this is a bit out of order, but wanted to get these changes in since they were simple and the readmes were updated:
- Our API contract for angular velocity of movement sensors is degrees/second
- The odrive naming triplet has changed.

cc @jeremyrhyde because I couldn't find your docs ticket.


sguequierre (approved):
thought i'd approve, change looks great, thank you so much for picking this up Rand!

martha-johnston (approved):

PR #1711, opened 2023-08-28
Update scuttlebot.md

mestcihazal:
renamed your-wheeled-base to match with the photo, but most photos need updating since they don't match the component names we give them to, thus creating misalignment between photo versus raw json

changed the order of wheel circumference and width since they are misordered according to how it is in the app

will do more reviews as i add more things and test

### DIFF
--- a/docs/tutorials/configure/scuttlebot.md
+++ b/docs/tutorials/configure/scuttlebot.md
@@ -13,7 +13,7 @@ aliases:
   - "/tutorials/scuttlebot/scuttlebot"
 authors: []
 languages: [ "python", "go" ]
-viamresources: [ "board", "arm", "motor" ]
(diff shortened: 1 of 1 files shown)


mestcihazal:
Also took out arm from the components because we don't use it, and took out camera from the hardware list because it comes with the scuttle robot similar to motors, motor driver, etc

docs/tutorials/configure/scuttlebot.md
```diff
@@ -13,7 +13,7 @@ aliases:
   - "/tutorials/scuttlebot/scuttlebot"
 authors: []
 languages: [ "python", "go" ]
-viamresources: [ "board", "arm", "motor" ]
+viamresources: [ "board", "motor", "camera" ]
```
npentrel, line 16: Thank you!

PR #1705, opened 2023-08-25
DOCS-1044, DOCS-1050: Provide model implementation guidance

andf-viam:
Add guidance to selecting a model name to the Modular Resources Key Concepts page
- Add links to this guidance from two places on existing `create` workflow page.
- Folded in separate guidance from DOCS-1035 as well (`to_lower()` pls!)
- Folded in additional guidance from DOCS-1050 as well (valid APIs to implement as custom models)
- Removing subtype content and API naming details per EH & FA feedback. Dedicated "New API subtype" page will move elsewhere in a separate PR.
	- Updated links elsewhere to use new headers.

## Staging

- [Modular Resources: Key Concepts](https://docs-test.viam.dev/0d6b450bee3620c3a98b715595977b9ca222723f/public/extend/modular-resources/key-concepts/#namespace-1)

PR #1714, opened 2023-08-29
DOCS-1035: Add naming tip for uppercase models bug with module registry

sguequierre:
* Sol-eng request: Add in a note gently urging them to use lowercase letters

### DIFF
--- a/docs/extend/modular-resources/create/_index.md
+++ b/docs/extend/modular-resources/create/_index.md
@@ -84,6 +84,13 @@ The following example module registers a modular resource implementing Viam's bu
 
 The Go code for the custom model (<file>mybase.go</file>) and module entry point file (<file>main.go</file>) is adapted from the full demo modules available on the [Viam GitHub](https://github.com/viamrobotics/rdk/blob/main/examples/customresources).
 
+{{% alert title="Naming module models" color="tip" %}}
+
+Name your model with all lowercase letters for optimal performance with Viam's SDKs.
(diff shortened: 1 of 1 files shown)


npentrel (approved):

docs/extend/modular-resources/create/_index.md
```diff
@@ -84,6 +84,13 @@ The following example module registers a modular resource implementing Viam's bu
 
 The Go code for the custom model (<file>mybase.go</file>) and module entry point file (<file>main.go</file>) is adapted from the full demo modules available on the [Viam GitHub](https://github.com/viamrobotics/rdk/blob/main/examples/customresources).
 
+{{% alert title="Naming module models" color="tip" %}}
+
+Name your model with all lowercase letters for optimal performance with Viam's SDKs.
```
andf-viam, line 89: Oh! I did not know about this bug!

I have [this PR](https://github.com/viamrobotics/docs/pull/1705) for model naming guidance, which adds some context to this section, and points to a new section on the Concepts page.

I think this guidance (lowercase) would better live there alongside the other related guidance I've added there, like triplet form, namespace matching, matching model `family` to API `subtype`, etc.

Do you think providing guidance in the above-referenced PR is the place for this? If you still feel strongly that we also need notice here, given the changes in my PR, I'm happy to review this on its own. Otherwise, I suggest that I take this ticket's ask and fold it into my existing naming PR. Whatcha think? Happy either way!

PR #1718, opened 2023-08-29
Update scuttlebot.md

mestcihazal:
replaced the right motor pins so it drives forward, fixed phrasing on direction flip according to this change, updated the raw JSON for the base (which is think is still missing some info like:         "encoder": "renc", "ticks_per_rotation": 2, "ramp_rate_rpm_per_sec": "") and i think we should add it, but in case i am missing something let me know

### DIFF
--- a/docs/tutorials/configure/scuttlebot.md
+++ b/docs/tutorials/configure/scuttlebot.md
@@ -202,8 +202,8 @@ Add the following JSON objects to the `components` array:
   "model": "gpio",
   "attributes": {
     "pins": {
-      "a": "15",
-      "b": "16",
+      "a": "16",
+      "b": "15",
       "pwm": "",
       "dir": ""
     },
@@ -258,7 +258,7 @@ If you hear a "whining" sound from the motor, the power level is not high enough
 If this happens, increase the power level by 10% increments until it starts to turn.
 {{< /alert >}}
 
-If you want your wheel to turn in reverse, you can add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
+If your wheel turns in reverse when it should turn forward, add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
 
 ## Configure the base
 
@@ -297,8 +297,8 @@ Click **Create component**.
         "board": "local",
         "pins": {
           "pwm": "",
-          "a": "15",
-          "b": "16",
+          "a": "16",
+          "b": "15",
         }
       },
       "model": "gpio",
--- a/docs/tutorials/configure/scuttlebot.md
+++ b/docs/tutorials/configure/scuttlebot.md
@@ -166,8 +166,8 @@ Select `Encoded` in the **Encoder** section and select `renc` as the **encoder**
 Next, describe how the motor is wired to the Pi:
 
 1. Switch the Component Pin Assignment Type to `In1/In2`.
-2. Set **A/In1** to `16`.
-3. Set **B/In2** to `15`.
+2. Set **A/In1** to `15`.
+3. Set **B/In2** to `16`.
 4. Leave the `pwm` (pulse-width modulation) pin blank, because this specific motor driver's configuration does not require a separate PWM pin.
 
 ![The motor config panel.](/tutorials/scuttlebot/pi-wheel.png)
@@ -202,8 +202,8 @@ Add the following JSON objects to the `components` array:
   "model": "gpio",
   "attributes": {
     "pins": {
-      "a": "16",
-      "b": "15",
+      "a": "15",
+      "b": "16",
       "pwm": "",
       "dir": ""
     },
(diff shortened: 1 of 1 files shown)


npentrel (approved):

docs/tutorials/configure/scuttlebot.md
```diff
@@ -258,7 +258,7 @@ If you hear a "whining" sound from the motor, the power level is not high enough
 If this happens, increase the power level by 10% increments until it starts to turn.
 {{< /alert >}}
 
-If your wheel turns in reverse when it should turn forward, add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
+If you want your wheel to turn in reverse, you can add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
```
npentrel, line 261: I think it makes sense to call this out as something people will encounter as opposite to what they're expercting
```suggestion
If you wheel turns in the opposite in reverse when it should turn forward, you can add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
```

mestcihazal: i was just not sure since we give them the exact a and b inputs, it wouldn't turn reverse?

npentrel: I think @JessamyT said it is possible to flip these accidentally? I think? Possible I am wrong. Let's wait for Jessamy to reply

mestcihazal: @JessamyT

JessamyT: I think that since DC motors don’t have an inherent “forward”, they might be flipped?? I think this sentence shouldn’t change from how it was before, because people wouldn’t want it to turn in reverse—it’s just if it is turning in reverse and they need to fix it.

JessamyT: As mentioned above, my SCUTTLE is configured in the way this page said before this PR's changes (with A 16, B 15), no dir flip, and it drives forwards properly. @mestcihazal do you have SCUTTLEs that have 15, 16 and no dir flip, and drive forward? At one point I know some of the NYC office scuttles that people were learning on were driving backwards (with the big wheels in the back) and had the cameras by the caster wheels.

@npentrel your suggestion seems to have a typo but what do you think of the original wording?

mestcihazal: the one I just configured has 15, 16, and no dir flip, and it drives forward (same with old tipsy)

docs/tutorials/configure/scuttlebot.md
```diff
@@ -166,8 +166,8 @@ Select `Encoded` in the **Encoder** section and select `renc` as the **encoder**
 Next, describe how the motor is wired to the Pi:
 
 1. Switch the Component Pin Assignment Type to `In1/In2`.
-2. Set **A/In1** to `16`.
-3. Set **B/In2** to `15`.
+2. Set **A/In1** to `15`.
```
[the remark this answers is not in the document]
mestcihazal: reverse for me

docs/tutorials/configure/scuttlebot.md
JessamyT: Just looked in the SCUTTLE wiring guide. When they talk about motor driver power wires, they say "Apply +5V to the red wire and select terminal which gives CCW rotation." So it's kinda up to the user to assemble it one way or the other, and they do specify, but people could totally do it the other way on accident. Also mine came pre-assembled by scuttle the way that it is--maybe yours did too.

npentrel: original wording is fine by me

docs/tutorials/configure/scuttlebot.md
```diff
@@ -166,8 +166,8 @@ Select `Encoded` in the **Encoder** section and select `renc` as the **encoder**
...
+3. Set **B/In2** to `16`.
```
npentrel, line 170:
```suggestion
2. Set **A/In1** to `16`.
3. Set **B/In2** to `15`.
```

docs/tutorials/configure/scuttlebot.md
```diff
@@ -202,8 +202,8 @@ Add the following JSON objects to the `components` array:
   "model": "gpio",
   "attributes": {
     "pins": {
-      "a": "16",
-      "b": "15",
+      "a": "15",
+      "b": "16",
```
npentrel, line 206:
```suggestion
      "a": "16",
      "b": "15",
```

docs/tutorials/configure/scuttlebot.md
npentrel:
```suggestion
If your wheel turns in reverse when it should turn forward, add the `dir_flip` attribute to the motor's configuration, by clicking **Show more** and setting the attribute to "true."
```

docs/tutorials/configure/scuttlebot.md
```diff
@@ -296,8 +296,9 @@ Click **Create component**.
       "attributes": {
         "board": "local",
         "pins": {
-          "dir": "16",
-          "pwm": "15"
+          "pwm": "",
+          "a": "15",
+          "b": "16",
```
npentrel, line 301:
```suggestion
          "a": "16",
          "b": "15",
```

npentrel:
Merging as there have been no futher changes.

PR #1722, opened 2023-08-30
Update RDK.md remove code ticks that don't render well in previews 

sguequierre:
* Just noticed this description was rendering without changing to code-- we could probably update something on site theme css to make it render as code if dev content overall doesn't want to make this change


andf-viam (approved):
Thanks, good catch!! LGTM

PR #1724, opened 2023-08-30
remove config builder tab from rplidar docs

EshaMaharishi:
Per this thread (https://viaminc.slack.com/archives/C02KFSY4H89/p1693405628296269), we're going to remove the Config Builder tab so that the docs don't have non-working instructions. The docs should be updated again once rplidar-module is uploaded to the modular registry.


andf-viam (approved):
Ah, I see this UI change is live now in the App. You're right, I can't see a way around this until the RPLidar is available on the Registry. Thanks for this, LGTM!

PR #1728, opened 2023-09-01
Update image on servo test sections

npentrel:


npentrel:
Hi @skyleilani, I noticed that we didn't restrict the image width and the screenshot feels a bit big currently. Te way we generally do that is with a short code: `{{<imgproc src="IMAGE" resize="400x" declaredimensions=true alt="ALT">}}` instead of the `ALT` syntax. Have a look at the servo pages on the staged link (see above) and if it looks good to you, you can approve this PR - thanks!

skyleilani (approved):

PR #1706, opened 2023-08-26
DOCS-908: Document part 1 of the cloud app API and move robot API

JessamyT:

PR #1731, opened 2023-09-01
Update _index.md

npentrel:
The card that is at the bottom of the page for the subpage is a bit random currently


JessamyT:
In https://github.com/viamrobotics/docs/pull/1706 I am making this no_list: true per your rec. I don't think it makes sense for these cards to be here, since they are just a few of the APIs, and there are links to these pages higher in the page.

PR #1764, opened 2023-09-07
DOCS-1070: Update arm and base builder UI instructions

JessamyT:
Part 1 of many


sguequierre (approved):
lgtm!! 💯  Recommend also getting naomi/andrew input but I'm good w this wording

npentrel (approved):

PR #1767, opened 2023-09-07
Fix JSON indentation on modular config doc

JessamyT:
Also applies a small body text change that was applied to the similar words in the other part of the doc.

### DIFF
--- a/docs/extend/modular-resources/configure.md
+++ b/docs/extend/modular-resources/configure.md
@@ -84,7 +84,7 @@ If, for example, the module provides a motor component, and the motor is running
 
 ### Configure a modular resource from a Registry module
 
-Once you have configured a module from the Viam Registry, you can add any number of the resources that module makes available to your robot by adding new components or services configured with your modular resources' new subtype or [model](/extend/modular-resources/key-concepts/#models).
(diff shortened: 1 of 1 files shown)


docs/extend/modular-resources/configure.md
```diff
@@ -84,7 +84,7 @@ If, for example, the module provides a motor component, and the motor is running
 
 ### Configure a modular resource from a Registry module
 
-Once you have configured a module from the Viam Registry, you can add any number of the resources that module makes available to your robot by adding new components or services configured with your modular resources' new subtype or [model](/extend/modular-resources/key-concepts/#models).
+Once you have configured a module from the Viam Registry, you can add any number of the resources that the module makes available to your robot by adding new components or services configured with your modular resources' [model](/extend/modular-resources/key-concepts/#models).
```
andf-viam, line 87: Thank you I must have brained `incoming | current` wrong 👍

andf-viam (approved):
LGTM, thank you for making this change.

docs/extend/modular-resources/configure.md
JessamyT: Haha I don't even know any more--this text was (and is) in two places so I think we both only caught one of them before :D

PR #1786, opened 2023-09-10
DOCS-963: Enable scroll on search modals

npentrel:


andf-viam (approved):
Yesss works well! Thanks so much -- LGTM!

PR #1793, opened 2023-09-11
Add templates to readme so contributors can find them more quickly

sguequierre:
Internal feedback from WTD conference


andf-viam (approved):
Thank you!

PR #1795, opened 2023-09-11
DOCS-1090: Note platform visibility in registry

andf-viam:
Add brief notice that a module's supported platforms can be viewed on that module's registry page.

- Will add additional related copy (selecting a platform) to the `configure` page for "add a module" steps as part of the DOCS - 1059 and DOCS - 1086 effort as soon as the eng tickets they are blocked on merge.


sguequierre (approved):
LGTM!

PR #1807, opened 2023-09-12
Update robots.md for better instructions to delete robot

sguequierre:
Was confused on how to delete a robot myself from the instructions from AI chatbot and App UI, think this rephrase could be an improvement

### DIFF
--- a/docs/manage/fleet/robots.md
+++ b/docs/manage/fleet/robots.md
@@ -111,6 +111,6 @@ Click on the **Generate Key** button to generate a new key.
 
 ## Delete a robot
 
-You can delete a robot by navigating to its **Config** page in [the Viam app](https://app.viam.com) and selecting **Sure?** and **Delete robot** in the lower left corner of the page.
+You can delete a robot by navigating to its page in [the Viam app](https://app.viam.com) and selecting **Sure?** and **Delete robot** in the lower left corner of the page.
 
 {{< imgproc alt="The DELETE ROBOT button and the confirmation checkbox (Sure?) next to it." src="/manage/app-usage/delete.png" resize="300x" >}}
--- a/docs/manage/fleet/robots.md
+++ b/docs/manage/fleet/robots.md
@@ -111,6 +111,6 @@ Click on the **Generate Key** button to generate a new key.
 
 ## Delete a robot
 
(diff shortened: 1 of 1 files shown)


docs/manage/fleet/robots.md
```diff
@@ -111,6 +111,6 @@ Click on the **Generate Key** button to generate a new key.
 
 ## Delete a robot
 
-You can delete a robot by checking the **Sure?** box in the lower left of the robot page and clicking **Delete robot**.
+You can delete a robot by navigating to its **Config** page in [the Viam app](https://app.viam.com) and selecting **Sure?** and **Delete robot** in the lower left corner of the page.
```
JessamyT, line 114:
```suggestion
You can delete a robot by navigating to its page in [the Viam app](https://app.viam.com) and selecting **Sure?** and **Delete robot** in the lower left corner of the page.
```
You can do this from any of the tabs, not just from the config tab.

JessamyT (changes requested):

docs/manage/fleet/robots.md
sguequierre: @JessamyT True!! However, you cannot do this from setup tab, which is where they land when opening the robot's page. I figured `navigate to config tab` would be simpler than saying Config, Security, etc-- is that ok with you/does that make sense?

JessamyT: Actually I saw it on the setup tab 10 minutes ago, but right now it looks like there's an app bug such that you can't do it from the config tab! The scrolling seems to be bugging. So I think you can actually do it from setup but I can't prove it right now because bug :D

JessamyT: I've filed an app bug ticket

JessamyT: Turns out ticket was a duplicate :D but did confirm with eng that there should be a delete robot box on the setup tab.

JessamyT (changes requested):
See comment thread

docs/manage/fleet/robots.md
sguequierre: Ok, cool. Would you recommend I just close this pr, or should we merge it for now and update once bug is closed? @JessamyT up to you

PR #1816, opened 2023-09-13
DOCS-1109: Clarify difference between `max_rotation_deg` and `max` attribute configuration for a `pi` servo

sguequierre:
* Add clarification here that `max` is user's choice for maximum rotation (along with `min`) and `max_rotation_deg` should be based off of what your hardware can possibly support, for PWM safety protection

### DIFF
--- a/docs/components/servo/pi.md
+++ b/docs/components/servo/pi.md
@@ -90,7 +90,7 @@ The following attributes are available for `pi` servos:
 | `max` | float | Optional | Sets a software limit on the maximum angle in degrees your servo can rotate to. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
 | `starting_position_degs` | float | Optional | Starting position of the servo in degrees. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
 | `hold_position` | boolean | Optional | If `false`, power down a servo if it has tried and failed to go to a position for a duration of 500 milliseconds. <br> Default = `true` |
-| `max_rotation_deg` | int | Optional | The maximum angle that you know your servo can possibly rotate to, according to its hardware. Refer to your servo's data sheet for clarification. Must be in between the values you choose for `min` and `max`. <br> Default = `180` |
+| `max_rotation_deg` | int | Optional | The maximum angle that you know your servo can possibly rotate to, according to its hardware. Refer to your servo's data sheet for clarification. Must be greater than or equal to the value you set for `max`. <br> Default = `180` |
 
 {{% alert title="Tip" color="tip" %}}
 
--- a/docs/components/servo/pi.md
+++ b/docs/components/servo/pi.md
@@ -86,11 +86,11 @@ The following attributes are available for `pi` servos:
 | ---- | ---- | --------- | ----------- |
 | `pin` | string | **Required** | The {{< glossary_tooltip term_id="pin-number" text="pin number" >}} of the pin the servo's control wire is wired to on the [board](/components/board/). |
 | `board` | string | **Required** | `name` of the [board](/components/board/) the servo is wired to. |
-| `min` | float | Optional | The minimum angle in degrees that the servo can reach. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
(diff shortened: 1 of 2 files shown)


docs/components/servo/pi.md
```diff
@@ -86,11 +86,11 @@ The following attributes are available for `pi` servos:
...
-| `max` | float | Optional | The maximum angle in degrees that the servo can reach. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
+| `min` | float | Optional | The minimum angle in degrees you want your servo to reach while rotating. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
+| `max` | float | Optional | The maximum angle in degrees you want your servo to reach while rotating. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
 | `starting_position_degs` | float | Optional | Starting position of the servo in degrees. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
 | `hold_position` | boolean | Optional | If `false`, power down a servo if it has tried and failed to go to a position for a duration of 500 milliseconds. <br> Default = `true` |
-| `max_rotation_deg` | int | Optional | The maximum angle the servo can rotate. Must be in between `min` and `max`. <br> Default = `180` |
+| `max_rotation_deg` | int | Optional | The maximum angle that you know your servo can possibly rotate to, according to its hardware. Refer to your servo's data sheet for clarification. Must be in between the values you choose for `min` and `max`. <br> Default = `180` |
```
JessamyT, line 93: I wonder whether the "Must be in between the values you choose for min and max" is actually accurate. I know it was here before but that part doesn't make sense to me, since min and max could be inside of that range.
Based on https://github.com/viamrobotics/rdk/blob/main/components/board/pi/impl/servo.go I think we can say:
```suggestion
| `max_rotation_deg` | int | Optional | The maximum angle that you know your servo can possibly rotate to, according to its hardware. Refer to your servo's data sheet for clarification. Must be greater than or equal to the value you set for `max`. <br> Default = `180` |
```

docs/components/servo/pi.md
```diff
@@ -86,11 +86,11 @@ The following attributes are available for `pi` servos:
 | ---- | ---- | --------- | ----------- |
 | `pin` | string | **Required** | The {{< glossary_tooltip term_id="pin-number" text="pin number" >}} of the pin the servo's control wire is wired to on the [board](/components/board/). |
 | `board` | string | **Required** | `name` of the [board](/components/board/) the servo is wired to. |
-| `min` | float | Optional | The minimum angle in degrees that the servo can reach. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
-| `max` | float | Optional | The maximum angle in degrees that the servo can reach. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
+| `min` | float | Optional | The minimum angle in degrees you want your servo to reach while rotating. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
```
JessamyT, line 89: I think describing it as a software limit will help users understand. What do you think of something like
```suggestion
| `min` | float | Optional | Sets a software limit on the minimum angle in degrees your servo can rotate to. <br> Default = `0.0` <br> Range = [`0.0`, `180.0`] |
```

docs/components/servo/pi.md
```diff
@@ -86,11 +86,11 @@ The following attributes are available for `pi` servos:
...
+| `max` | float | Optional | The maximum angle in degrees you want your servo to reach while rotating. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
```
JessamyT, line 90:
```suggestion
| `max` | float | Optional | Sets a software limit on the maximum angle in degrees your servo can rotate to. <br> Default = `180.0` <br> Range = [`0.0`, `180.0`] |
```

JessamyT (changes requested):

docs/components/servo/pi.md
sguequierre: Ahh, this does seem right. It seems like it's actually checking << `min` and `max` , though, so I will edit to be that

docs/components/servo/pi.md
sguequierre: great suggestion, sounds good to me!

JessamyT (approved):

PR #1818, opened 2023-09-13
Update configure.md

zaporter-work:
# Description

Type: local is not strictly required yet, but it will be in the coming months. Might as well tell people about the flag so there are fewer configs that are silently migrated. 

Not sure if this is the right way to indicate this to the user. 

Feel free to change this up. Just wanted to mention it.


andf-viam (approved):
LGTM! Thanks, sorry for the delay!

PR #1819, opened 2023-09-13
DOCS-1113: Add clarification for running code remotely to client SDK Run Code page

sguequierre:
* Trying to explain the "why" of this use case-- internal docs feedback from @mcvella and team

### DIFF
--- a/docs/program/run.md
+++ b/docs/program/run.md
@@ -28,7 +28,7 @@ You can find it at the top of the robot's **Control** tab.
 
 Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
 The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
-You can remotely control your robot from anywhere in the world.
+You can remotely control your robot with any application you implement from anywhere in the world.
 After editing your code to include your robot's [authentication credentials](#authentication), run a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed:
 
 {{< tabs >}}
--- a/docs/program/run.md
+++ b/docs/program/run.md
@@ -10,13 +10,27 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
 
 After saving your [code sample](/program/#hello-world-the-code-sample-tab) and adding control logic with [Viam's SDKs](/program/apis/), run your program to control your Viam-connected robot.
 
+### Authentication
+
+You must reference a robot's location secret to authenticate yourself to the robot.
+However, the app hides the robot location secret from the sample by default for your security.
+
+To copy the robot location secret, select **Include Secret** on the **Code sample** tab of your robot's page on the [Viam app](https://app.viam.com).
+Paste it into your SDK code as directed by the code sample.
+
+You must also include the robot's remote address, like `12345.somerobot-main.viam.cloud`, as an external or public address to connect to your robot.
+The code sample should include this address at default.
+You can find it at the top of the robot's **Control** tab.
+
+{{% snippet "secret-share.md" %}}
+
 ## Run Code Remotely
 
-As long as you are connecting to your robot through an external signaling address, you can remotely control your robot without editing the app-generated connection code in your SDK program beyond including the necessary [authentication credentials](#authentication).
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not even have to connected to the same WAN/LAN to issue control commands, as your robot receives operation requests from your client program over the cloud.
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
-
-For example:
+As long as you are connecting to your robot through an external signaling address, you can remotely control your robot from anywhere in the world.
+After editing your code to include your robot's [authentication credentials](#authentication), run a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed:
 
 {{< tabs >}}
 {{% tab name="Python" %}}
@@ -53,20 +67,6 @@ flutter run <DART_FILE>
 {{< /tabs >}}
 
 This is useful because as long as that computer is able to establish a network connection with the robot's computer, your control logic will be executed on the robot.
-
-### Authentication
-
-You must reference a robot's location secret to authenticate yourself to the robot.
-However, the app hides the robot location secret from the sample by default for your security.
-
-To copy the robot location secret, select **Include Secret** on the **Code sample** tab of your robot's page on the [Viam app](https://app.viam.com).
-Paste it into your SDK code as directed by the code sample.
-
-You must also include the robot's remote address, like `12345.somerobot-main.viam.cloud`, as an external or public address to connect to your robot.
-The code sample should include this address at default.
-You can find it at the top of the robot's **Control** tab.
-
-{{% snippet "secret-share.md" %}}
 
 ## Run Code On-Robot
 
--- a/docs/program/run.md
+++ b/docs/program/run.md
@@ -29,7 +29,7 @@ You can find it at the top of the robot's **Control** tab.
 Most of the time, as long as both you and your robot are connected to the internet, you will want to run code to control your robot remotely.
 The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
 You can remotely control your robot with any application you implement from anywhere in the world.
-For example, you can run code on your personal computer, creating a client [session](/program/apis/sessions/), where the code running on that computer sends instructions to your robot's computer's `viam-server` instance over the internet.
+For example, you can run code on your personal computer, creating a client [session](/program/apis/sessions/), where the code running on that computer sends instructions to your robot's `viam-server` instance over the internet.
 
 After editing your code to include your robot's [authentication credentials](#authentication), run a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed:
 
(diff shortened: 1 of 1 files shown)


docs/program/run.md
```diff
@@ -10,13 +10,27 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+The code sample should include this address at default.
+You can find it at the top of the robot's **Control** tab.
+
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not even have to connected to the same WAN/LAN to issue control commands, as your robot receives operation requests from your client program over the cloud.
```
sguequierre, line 30:
```suggestion
The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands, as your robot receives operation requests from your client program over the cloud.
```

docs/program/run.md
```diff
@@ -10,13 +10,27 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+The code sample should include this address at default.
+You can find it at the top of the robot's **Control** tab.
+
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands, as your robot receives operation requests from your client program over the cloud.
```
sguequierre, line 30:
```suggestion
The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
```

docs/program/run.md
```diff
@@ -10,13 +10,27 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+The code sample should include this address at default.
+You can find it at the top of the robot's **Control** tab.
+
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
```
sguequierre, line 31:
```suggestion
```

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-For example:
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
+As long as you are connecting to your robot through an external signaling address, you can remotely control your robot from anywhere in the world.
```
sguequierre, line 31:
```suggestion
You can remotely control your robot from anywhere in the world.
```

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-For example:
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
+The advantage of this method is that your robot and your computer do not have to be connected to the same WAN/LAN to issue control commands.
+You can remotely control your robot from anywhere in the world.
```
sguequierre, line 31:
```suggestion
You can remotely control your robot with any application you implement from anywhere in the world.
```

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+The code sample should include this address at default.
+You can find it at the top of the robot's **Control** tab.
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-For example:
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
```
npentrel, line 29: This doesn't make sense to have in a section that has the heading "run code remotely" because you're not describing that - you're describing moving the code to the robot and running it there instead.

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
 
 After saving your [code sample](/program/#hello-world-the-code-sample-tab) and adding control logic with [Viam's SDKs](/program/apis/), run your program to control your Viam-connected robot.
 
-## Run Code Remotely
+### Authentication
+
+You must reference a robot's location secret to authenticate yourself to the robot.
```
npentrel, line 15: I don't think a user would know what they are supposed to do to "reference a robot's location secret", maybe include or add would be better?

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+However, the app hides the robot location secret from the sample by default for your security.
+
+To copy the robot location secret, select **Include Secret** on the **Code sample** tab of your robot's page on the [Viam app](https://app.viam.com).
+Paste it into your SDK code as directed by the code sample.
 
-As long as you are connecting to your robot through an external signaling address, you can remotely control your robot without editing the app-generated connection code in your SDK program beyond including the necessary [authentication credentials](#authentication).
+You must also include the robot's remote address, like `12345.somerobot-main.viam.cloud`, as an external or public address to connect to your robot.
+The code sample should include this address at default.
```
npentrel, line 22:
```suggestion
The code sample includes this address at default.
```

(let's trust our product)

docs/program/run.md
npentrel: Oh I see now you were just movign this up... We can fix it separetely

docs/program/run.md
sguequierre: Hahaha yes sounds good, done

docs/program/run.md
sguequierre: Hmmm-- this is the ask itself. And I can change the heading but you are running code remotely, like remotely from the computer where the robot is located, then it "moves" the code to the robot. Vs on-robot, you are on the same computer. I believe remote is a fairly standard term for this but I am slightly head-scratched by this as well so I can see why it's confusing. What would you think about `Run Code Remotely from the Robot`? @npentrel

npentrel: But if you are running the code from the robot you are by definition not remote? I think this conflates both concepts. Here's my understanding with my guardian robot as the example. I can:
1. Run the code that makes it move and follow my dog from my macbook. Then I am running the code remotely. The python code does not get "moved" to the robot. A client session is created and the code running on my macbook will send instructions to the viam-server instance. If the connection drops because my mac loses wifi, the robot stops moving because the code is not actually on the robot now, it was just receiving the commands resulting from the code.
2. Run the code locally on the robot. Then I am putting the python file for my robot on the raspberry pi that is the heart of the robot. This is what I actually do and then I have the script [automatically start when viam-server is running](https://docs.viam.com/tutorials/projects/guardian/#run-the-program-automatically).

Now, it occurs to me that it could also be that by "remote" we might mean not on the same WAN/LAN. I'm not sure that's clear. But also then - no code is uploaded.

^ Does this make it clearer? Explaining it with an example may help the reader's understanding here

docs/program/run.md
```diff
@@ -10,13 +10,26 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+You must also include the robot's remote address, like `12345.somerobot-main.viam.cloud`, as an external or public address to connect to your robot.
+The code sample includes this address at default.
+You can find it at the top of the robot's **Control** tab.
+
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-For example:
+Most of the time, as long as both you and your robot are connected to the internet, you will want to upload code to your robot remotely.
```
sguequierre, line 29:
```suggestion
Most of the time, as long as both you and your robot are connected to the internet, you will want to run code on your robot remotely.
```

docs/program/run.md
sguequierre: Yes, that is super helpful, thank you!! I was confused thinking about how it worked/didn't realize "uploading code to the robot" was inaccurate. Will edit to have the use case with the macbook to make that more clear.

docs/program/run.md
```diff
@@ -10,13 +10,28 @@ tags: ["client", "sdk", "application", "sdk", "fleet", "program"]
...
+The code sample includes this address at default.
+You can find it at the top of the robot's **Control** tab.
 
-As long as you are connecting to your robot through an external signaling address, you can remotely control your robot without editing the app-generated connection code in your SDK program beyond including the necessary [authentication credentials](#authentication).
+{{% snippet "secret-share.md" %}}
+
+## Run Code Remotely
 
-Do this by running a command to execute the program in the terminal of a machine with the appropriate programming language and Viam SDK installed.
+Most of the time, as long as both you and your robot are connected to the internet, you will want to run code on your robot remotely.
```
sguequierre, line 29:
```suggestion
Most of the time, as long as both you and your robot are connected to the internet, you will want to run code to control your robot remotely.
```

npentrel: run code to control the robot remotely

PR #1820, opened 2023-09-13
DOCS-1116: Update viam CLI brew install

andf-viam:
Removed both macOS arch direct download steps in favor of new `brew` install.
- Also updated "To later update ..." sentence with macOS (brew)-specific instruction.
- Left Linux instructions as-are (but simplified `uname -m` tip, since it now only applies to a subset of installation paths)

## Staging
- [/manage/cli/#install](https://docs-test.viam.dev/08459e33eca493b9327147cabd30ee23e22262d8/public/manage/cli/#install)

### DIFF
--- a/docs/manage/CLI.md
+++ b/docs/manage/CLI.md
@@ -55,23 +55,13 @@ sudo chmod a+rx /usr/local/bin/viam
 ```
 
 {{% /tab %}}
-{{% tab name="macOS arm64" %}}
+{{% tab name="macOS" %}}
 
-To download the Viam CLI on a macOS computer with the `arm64` (Apple Silicon) architecture, run the following commands:
-
-```{class="command-line" data-prompt="$"}
-sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.com/apps/viam-cli/viam-cli-stable-darwin-arm64
-sudo chmod a+rx /usr/local/bin/viam
-```
-
-{{% /tab %}}
-{{% tab name="macOS x86_64" %}}
-
-To download the Viam CLI on a macOS computer with the `amd64` (Intel `x86_64`) architecture, run the following commands:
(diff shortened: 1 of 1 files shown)


docs/manage/CLI.md
```diff
@@ -29,10 +29,7 @@ viam.component.servo.v1.ServoService.MoveRequest
...
 You can download the Viam CLI executable using one of the options below.
 Select the tab for your platform and architecture.
-
-{{% alert title="Tip" color="tip" %}}
-You can use the `uname -m` command to determine your system architecture.
-{{% /alert %}}
+If you are on Linux, you can use the `uname -m` command to determine your system architecture.
 
 {{< tabs >}}
 {{% tab name="Linux aarch64" %}}
```
abe-winter, line 35: out of scope for PR, but should macos be first here? (on the assumption it's the most common case)

docs/manage/CLI.md
```diff
@@ -95,7 +82,8 @@ echo 'export PATH="$HOME/go/bin:$PATH"' >> ~/.bashrc
 {{% /tab %}}
 {{< /tabs >}}
 
-To later update the Viam CLI tool, you can use the steps above to reinstall the latest version.
+To later update the Viam CLI tool on Linux, use the steps above to reinstall the latest version.
+to later update the Viam CLI tool on macOS, run `brew upgrade viamrobotics/brews/viam`.
```
abe-winter, line 86: if they went through `brew tap` above, can also do `brew upgrade viam`, but this works too

abe-winter (approved):
lgtm

docs/manage/CLI.md
andf-viam: You're right, works for me!

docs/manage/CLI.md
andf-viam: Good point!

PR #1821, opened 2023-09-13
Add linuxbrew install instructions

zaporter-work:
# Description

tested on linux x86_64.

Might as well include instructions so that people get updates.

@abe-winter Is there a reason we didn't include instructions for linuxbrew?

### DIFF
--- a/docs/manage/CLI.md
+++ b/docs/manage/CLI.md
@@ -51,13 +51,6 @@ sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.co
 sudo chmod a+rx /usr/local/bin/viam
 ```
 
-You can also install the Viam CLI using [brew](https://brew.sh/):
-
-```{class="command-line" data-prompt="$"}
-brew tap viamrobotics/brews
-brew install viam
-```
-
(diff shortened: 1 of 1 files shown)


zaporter-work:
@andf-viam 

I just threw this pr up to cause discussion.


If we decide to let customers know that it works on linux, might be better to throw these instructions in the tab for `Linux x86` rather than creating a new tab.

docs/manage/CLI.md
```diff
@@ -51,6 +51,13 @@ sudo curl -o /usr/local/bin/viam https://storage.googleapis.com/packages.viam.co
 sudo chmod a+rx /usr/local/bin/viam
 ```
 
+You can also install the Viam CLI using [brew](https://brew.sh/):
+
+```{class="command-line" data-prompt="$"}
+brew tap viamrobotics/brews
+brew install viam
+```
+
```
zaporter-work, line 60:
```suggestion
```

brew is linux/x86_64 only.

zaporter-work (commented):
LGTM!

andf-viam (approved):
LGTM, thanks!

PR #1825, opened 2023-09-14
DOCS-792: Persist tab selection to local storage

npentrel:


andf-viam (approved):
LGTM! This is some ace coding! I'll write up a walkthrough of setting up named tabsets in the note.md file, once this builds and I test around. Amazing job!

npentrel:
So this actaully just works by the tab name. So as long as we use the same "Go" and "JSON Template" etc that's all it needs. Not that we can't change that if you think something else would be better

PR #1839, opened 2023-09-15
Update CLI.md from organization to organizations

dannenberg:
plural

# Description


dannenberg:
maybe not.... apparently both exist now...

dannenberg:
i'ma re-open. these two commands are now one (along with org). i think we should either use longest name or shortest name consistently. this is a silly nit of a PR now. feel free to throw it out.

dannenberg:
citing my source https://viaminc.slack.com/archives/C02KFSY4H89/p1694799255900919?thread_ts=1694796568.220349&cid=C02KFSY4H89

npentrel (approved):

PR #1852, opened 2023-09-16
DOCS-493: Clear images for new headlines

npentrel:
On wide monitors this will create a gap instead (there's some margin magic going on for linking to headers that interferes with this) - I think that's probably fine though


andf-viam (approved):
LGTM, let's see this in action!

PR #1869, opened 2023-09-19
Update ramp_rate attribute on encoded motor

JessamyT:
Seems like this is the actual JSON attribute name as of https://github.com/viamrobotics/rdk/pull/2905


martha-johnston (approved):

andf-viam (approved):
Thanks for catching!

PR #1870, opened 2023-09-20
Update connectivity.md

npentrel:


npentrel:
@dannenberg - could you read through the changes to the connectivity page and ensure that that answers all remaining questions

and @cheukt - apologies - could you confirm this is all still correct?

@sguequierre - please read over it for correctness as well

cheukt (approved):
looks accurate

docs/program/connectivity.md
```diff
@@ -3,25 +3,22 @@ title: "Client Sessions and Robot Network Connectivity"
...
 
-## What happens to clients connected to the robot when your robot loses its network connection?
+When a robot loses its connection to LAN or WAN, all client sessions will timeout and end by default.
 
-When your client cannot connect to your robot's `viam-server` instance, all client operations will timeout automatically and halt.
-Any active commands will be cancelled, stopping any moving parts, and no new commands will be able to reach the robot until the connection is restored.
+## Client session timeout and end
 
-### How do Viam's client SDKs work if your robot loses its network connection?
+When your client cannot connect to your robot's `viam-server` instance, `viam-server` will end any current client [*sessions*](/program/apis/sessions/) on this robot and all client operations will [timeout automatically]() and halt: any active commands will be cancelled, stopping any moving parts, and no new commands will be able to reach the robot until the connection is restored.
```
npentrel, line 22:
```suggestion
When your client cannot connect to your robot's `viam-server` instance, `viam-server` will end any current client [*sessions*](/program/apis/sessions/) on this robot and all client operations will [timeout automatically](/program/apis/sessions/#heartbeats) and halt: any active commands will be cancelled, stopping any moving parts, and no new commands will be able to reach the robot until the connection is restored.
```

dannenberg (commented):
the only question i'm not seeing an answer to here is "can a client establish a connection to the viam-server over LAN without an internet connection?" and possibly implicit in that "if we were connected via a route using the internet, but that gets severed, will the client be able to reconnect over their shared LAN?" (could this happen. that is, would webrtc ever choose a route out of LAN when LAN is available?)

otherwise this is looking great. thank you for your efforts

docs/program/apis/sessions.md
```diff
@@ -17,6 +17,8 @@ This is especially important for robots that physically move.
 For example, imagine a wheeled rover gets a [`SetPower()`](/components/base/#setpower) command as the last input from a client before the connection to the robot is interrupted.
 Without session management, the API request from the client would cause the rover's motors to move, causing the robot to continue driving forever and potentially colliding with objects and people.
 
+For more information on information, see [Client Sessions and Robot Network Connectivity](/program/connectivity/).
```
sguequierre, line 20:
```suggestion
For more information, see [Client Sessions and Robot Network Connectivity](/program/connectivity/).
```
Not sure what intended (replace word) but seems like typo

docs/program/connectivity.md
```diff
@@ -3,25 +3,22 @@ title: "Client Sessions and Robot Network Connectivity"
 linkTitle: "Network Connectivity Issues"
 weight: 20
 type: "docs"
-description: "Whan a robot loses its connection to the internet, all client sessions will timeout and end by default."
+description: "When a robot loses its connection to LAN or WAN, all client sessions will timeout and end by default."
```
sguequierre, line 6:
```suggestion
description: "When a robot loses its connection to a LAN or WAN, all client sessions will timeout and end by default."
```

docs/program/connectivity.md
```diff
@@ -3,25 +3,22 @@ title: "Client Sessions and Robot Network Connectivity"
...
 tags: ["client", "sdk", "viam-server", "networking", "apis", "robot api", "session"]
 ---
 
-What happens when your robot loses its connection to the internet over a LAN or WAN and can no longer communicate with clients through [the Viam app](https://app.viam.com).
+A [client session](/program/apis/sessions/) uses the most efficient route to connect to your robot either through local LAN or WAN or the internet.
 
-## How does `viam-server` work if your robot loses its network connection?
+When a robot loses its connection to the internet but is still connected to a local LAN or WAN:
```
sguequierre, line 12:
```suggestion
When a robot loses its connection to the internet but is still connected to a LAN or WAN:
```
(opt) I might be wrong here but this seems redundant with `local local area network` so would suggest removiign

docs/program/connectivity.md
```diff
@@ -3,25 +3,22 @@ title: "Client Sessions and Robot Network Connectivity"
...
 
-When a robot loses its connection over LAN or WAN, it can no longer communicate with clients through [the Viam app](https://app.viam.com).
-When no heartbeat has been received over a timeout period, `viam-server` will end any current client [*sessions*](/program/apis/sessions/) on this robot.
+- Client sessions connected through the internet will timeout and end.
```
sguequierre, line 14: (opt) For clarity I would suggest moving this down and specifying "different LAN or WAN" or "through the cloud" vs just "through the internet", but idk I think this is technically correct anyways?

sguequierre (approved):

npentrel:
@cheukt please confirm the answers:

1. "can a client establish a connection to the viam-server over LAN without an internet connection?" 

I assume yes and you'd do that the same way as https://docs-test.viam.dev/d4b541ef8633aa931fd9723fc58b82f3d0fd8915/public/program/run/#run-code-on-robot ?

2. "if we were connected via a route using the internet, but that gets severed, will the client be able to reconnect over their shared LAN?" (could this happen. that is, would webrtc ever choose a route out of LAN when LAN is available?)

I believe it would automatially figure out there's a better connection and use LAN, so no reconnection needed?

cheukt:
for 1, yes, and at this point you should be able to connect just using the code sample for all the supported SDKs
for 2, if the client was connected over internet, it will disconnect and reconnect again, but that should get handled automatically. it should favor a connection over LAN though, at which point connection will not get dropped

dannenberg:
for clarity, the steps naomi linked in #1 are no longer necessary. correct cheuk?

cheukt:
I believe so - should definitely work with go and python sdk at least

npentrel:
@dannenberg and @cheukt could you please review the changes in the last commit ("Address feedback") to ensure the changes based on your comments are factually accurate and cover what we want to cover? 

We will need to understand the Typescript limitations as well but we can do that separately.

docs/program/connectivity.md
```diff
@@ -3,26 +3,24 @@ title: "Client Sessions and Robot Network Connectivity"
...
+When connecting to a robot using the connection code from the [code sample tab](/program/#hello-world-the-code-sample-tab), a [client session](/program/apis/sessions/) automatically uses the most efficient route to connect to your robot either through local LAN or WAN or the internet.
 
-## How does `viam-server` work if your robot loses its network connection?
+When a robot loses its connection to the internet but is still connected to a LAN or WAN:
 
-When a robot loses its connection over LAN or WAN, it can no longer communicate with clients through [the Viam app](https://app.viam.com).
-When no heartbeat has been received over a timeout period, `viam-server` will end any current client [_sessions_](/program/apis/sessions/) on this robot.
+- Client sessions connected through the same LAN or WAN will function normally.
+- Client sessions connected through the internet will timeout and end.
+  If the client is on the same LAN or WAN, the client will automatically disconnect and then reconnect over LAN.
```
dannenberg, line 17: this seems to claim we will see a disconnect when this occurs, but i would expect that only happens if the connection was for some reason not direct to bot. in the case where it was not direct, (based on some of cheuks claims), i would expect it cannot connect locally? 

perhpas my expectation of non local connect is wrong, in which case, i'd still expected it wouldnt only disconnect in some cases (those where it was connected via internet/external connections)

npentrel: @cheukt should I remove the last sentence since very likely sessions would already have chosen the most efficient route through LAN/WAN?

cheukt: did some manual testing, I would say it's very likely that sessions (which usually just uses the client connection) would have chosen the most efficient route, which would be direct to robot over the LAN/WAN

cheukt: so last line is probably not needed

npentrel: How about this:
```suggestion
  If the client is on the same LAN or WAN but the route it chose to connect is through the internet, the client will automatically disconnect and then reconnect over LAN.
```

cheukt (approved):

PR #1872, opened 2023-09-21
DOCS-1146: Update outdated code on Go custom base example

sguequierre:
Updated directly from here: https://github.com/viamrobotics/rdk/blob/main/examples/customresources/models/mybase/mybase.go


bhaney (approved):
great! The RDK example works, so this should work too

PR #1873, opened 2023-09-21
DOCS-1111: Make more clear when you need to prepare venv

sguequierre:
Make this page fit in better


docs/program/_index.md
```diff
@@ -293,6 +293,10 @@ This connection must be established for your program to be executed properly on
 
 {{< readfile "/static/include/snippet/secret-share.md" >}}
 
+### Set up your virtual environment
+
+If you are using the Python SDK, [set up a virtual environment](/program/apis/python-venv) to package the SDK inside before running your code, avoiding conflicts with other projects or your system.
```
sguequierre, line 298:
```suggestion
If you are using the Python SDK, [set up a virtual environment](/program/python-venv/) to package the SDK inside before running your code, avoiding conflicts with other projects or your system.
```

docs/program/python-venv.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Prepare your Python Virtual Environment"
-linkTitle: "Prepare Virtualenv"
+linkTitle: "Virtualenv for Python SDK"
```
sguequierre, line 3: open to any other suggestions here but for now, think this might make it more clear these docs are just if you are using python SDK (not just using viam to program a robot at all)

npentrel: Kind of makes me wonder if this page should just be information in an expander on the program page

docs/program/_index.md
```diff
@@ -293,6 +293,10 @@ This connection must be established for your program to be executed properly on
 
 {{< readfile "/static/include/snippet/secret-share.md" >}}
 
+### Set up your virtual environment
```
npentrel, line 296: This doesn't make sense here because following the page flow they'd already have installed the SDK now - let's move this to the requirements section and have it in the python tab where the install command is

sguequierre: oh yes thank you 😵‍💫 , will do and comment here when done

docs/program/python-venv.md
sguequierre: I think it's a lot of info to put in an expander and people may look for it otherwise-- I like directing from requirements. Expander is not a bad idea though, so really up to you (if you want it I'll make the change!!)

docs/program/_index.md
sguequierre: done!

docs/program/_index.md
```diff
@@ -41,9 +41,7 @@ Next, to install your preferred Viam SDK on your Linux or macOS development mach
 {{< tabs >}}
 {{% tab name="Python" %}}
 
-```sh {class="command-line" data-prompt="$"}
```
npentrel, line 44: but we want to keep this command - just put it after the virtual env

sguequierre: ok done!!

npentrel (approved):

PR #1877, opened 2023-09-21
DOCS-1070: Pt6: Update builder UI instructions

JessamyT:


sguequierre (approved):
LGTM thank you Jessamy for going through all this!!

PR #1882, opened 2023-09-25
DOCS-1159: Update VisionClient python examples

npentrel:


bhaney (approved):
good catch on making sure to use the camera name in the arguments, not the camera object

dannenberg (approved):
lgtm thanks for fixes!

PR #1894, opened 2023-09-25
Update confetti-bot.md code for correct board name

sguequierre:
* Noticed that raw JSON template was using `local` after naming board `party`


andf-viam (approved):
LGTM, thanks for catching!

PR #1920, opened 2023-09-27
DOCS-1079: Pt1: Update arm and base config instructions

JessamyT:
- An option for manual fancy mode assistance (in [fake arm](https://docs-test.viam.dev/14cf9bfb65093d3f43ffc365f4123f343d4e4195/public/components/arm/fake/))
- Update various screenshots with latest UI (i.e. no more attribute guide, and some were very old)
- In the JSON examples/templates, standardize order of name, model, type, attributes, depends_on--that's what it currently is in the app.
- Some indentation linting
- Fix ur5e attribute description (per open RDK PR)


docs/components/arm/fake.md
```diff
@@ -57,35 +74,32 @@ If both attributes are specified, an error is thrown stating "can only populate
 
 {{% /alert %}}
 
-Refer to the following JSON examples for differences in configuration between the two attributes available:
+### Attributes examples
```
npentrel, line 77: I'm wondering whether these should be inside the config builder tab - we did want to start adding more guidance on attributes there. I can see why they'd be generally useful but I think treating folks on the json tabs as advanced users who need less examples could be ok? What do you think?

JessamyT: Yeah I was wondering about that too—I agree about the advanced users thing. I just don’t want to introduce these until people know what the attributes are even for (which the table describes)

npentrel: But we already do that with the JSON example the same way. We show an example and underneath it is the table that explains the options a bit more. So I think in the config builder tab adding it at the bottom with the first tab as a template and then a tab for JSON example with the attribute examples is what people might expect

JessamyT: But then config builder users are shown JSON snippets without knowing what to do with them? Do you think that'd be okay if it includes notes/links down to the attribute table?

npentrel: Config builder users need to fill in the attribute field with a json object already. DOCS-1079 is supposed to help users on the config builder have a json object with attribute placeholders to copy paste into the attribute field. The easiest way I see to do that is to have a tab with the general json template for the available attributes and then examples. The context is set with a sentence where you say something like

I think in the config builder tab you'd have something like this:

```
Copy and paste the following attribute template and then fill in the attributes as applicable.
See the JSON Examples tab for attribute example JSON documents: 

{{< tabs >}}
{{% tab name="JSON Template" %}}

```json
{
        "host": "<your-arms-ip-address-on-your-network>",
        "port": <int>,
        "speed_degs_per_sec": <float>,
        "acceleration_degs_per_sec_per_sec": <float>
}
```

{{% /tab %}}
{{% tab name="JSON Examples" %}}

```json
{
        "host": "localhost",
        "port": 8080,
        "speed_degs_per_sec": 1234,
        "acceleration_degs_per_sec_per_sec": 1.2
}
```

{{% /tab %}}
{{< /tabs >}}

```

JessamyT: Okay I was thinking we'd be even more hand-holdy but your suggestion does seem more generalizable. Sounds good. Also didn't even realize I could put tabs inside of tabs--awesome :D

npentrel (approved):
I would like the JSON template attributes snippet on each arm page - but ofc that can happen in a different PR - is that the intention?

JessamyT:
> I would like the JSON template attributes snippet on each arm page - but ofc that can happen in a different PR - is that the intention?

Well, all the other arms have fancy config available so I figured it’s not necessary and would only confuse config builder users. If you think so we can revisit later.

PR #1922, opened 2023-09-28
Update segmentation.md

bhaney:
# Description

obstacles pointcloud had outdated text related to the old "radius clustering segmenter". The new model is now faster, and works a bit different than the radius clustering model.


npentrel (approved):
Thank you!

PR #1925, opened 2023-09-28
Update python-venv.md

mestcihazal:
added the python3-venv installation step


mcvella (approved):

npentrel:
Hi @mcvella, can we please in future wait for a docs reviewer before merging? 

You've added an instruction to a code block that is described in the preceeding paragraph as `In the project directory, create and activate a virtual environment for Python to run in.` and now it also installs venv - and that's not really what we should do. I will fix it but please in future wait unless it's urgent. If it's urgent you're fine to merge but please note that on the PR - thank you!

PR #1927, opened 2023-09-29
DOCS-1079: Make board attribute configuration easier

JessamyT:
Making this draft to show how ideas would apply to board


docs/components/board/_index.md
```diff
@@ -151,7 +153,8 @@ Integrate `digital_interrupts` into your robot in the `attributes` of your board
 "digital_interrupts": [
   {
     "name": "<your-digital-interrupt-name>",
-    "pin": "<pin-number>"
+    "pin": "<pin-number>",
+    "type": "< basic | servo >"
```
sguequierre, line 157:
```suggestion
    "type": "<`basic` or `servo`>"
```
(opt) I like the idea but I think the two options also belongs in the attribute table, the | or also ppl might not automatically know what that means

docs/components/board/_index.md
```diff
@@ -320,7 +323,7 @@ The following properties are available for `i2cs`:
 | Name | Type | Inclusion | Description |
 | ---- | ---- | --------- | ----------- |
 |`name`| string| **Required** | `name` of the I<sup>2</sup>C bus. |
-|`bus_select`| int | **Required** | The index of the I<sup>2</sup>C bus. |
+|`bus`| string | **Required** | The index of the I<sup>2</sup>C bus. |
```
sguequierre, line 326:
```suggestion
|`bus`| string | **Required** | The index of the I<sup>2</sup>C bus. <br> Example: `"1"` |
```
(opt) Would help to put example on here-- assuming this is what it looks like if it's string

static/include/components/board-attr-config.md
```diff
@@ -0,0 +1,60 @@
+Copy and paste the following attribute template into your board's **Attributes** box.
+Then remove and fill in the attributes as applicable to your board, according to the table below.
```
sguequierre, line 2:
```suggestion
Edit and fill in the attributes as applicable to your board, according to the table below.
```
(opt) Just to keep wording consistent

docs/components/board/pi.md
```diff
@@ -36,9 +36,25 @@ Edit and fill in the attributes as applicable.
...
     {
-      "type": "board",
+      "name": "<your-pi-board-name>",
       "model": "pi",
-      "name": "<board_name>"
+      "type": "board",
+      "namespace": "rdk",
+      "attributes": {
+        "analogs": [
+          <...See table below...>
```
sguequierre, line 45:
```suggestion
          <See table below>
```
(opt) I think the ... is a little informal but not a biggie, would suggest removing

static/include/components/board-attr-config.md
```diff
@@ -0,0 +1,60 @@
...
+
+{{< tabs >}}
```
sguequierre, line 4: This seems like a really good add, thank you jessamy!!

docs/components/board/_index.md
JessamyT: They are in the attribute table already, unless I'm misunderstanding you. I like the pipe because it's visually distinctive, but could go either way. Curious what Naomi says.

static/include/components/board-attr-config.md
JessamyT: I changed it for these ones with the non-fancy configs because if we're telling people to copy paste a template that has all attributes, even the optional and mutually-exclusive ones in it, I want to make it explicit that they can and should remove some of them.

docs/components/board/pi.md
JessamyT: I put it in because you (or maybe someone else) used it in various other places like the base configs. I figured this is a way to emphasize that this needs to be expanded significantly and isn't just a one-line edit for the user to make.

docs/components/board/_index.md
JessamyT: I wasn't actually sure what a good example would be--would this example apply to all board models?

docs/components/board/_index.md
npentrel: I think the pipe is how that is commonly done to conserve space, so I'd stick with that way of doing it though I think it's also commonly without extra spaces so:

```suggestion
    "type": "<basic|servo>"
```

npentrel (approved):

docs/components/board/_index.md
sguequierre: I believe so but SME could confirm

docs/components/board/pi.md
sguequierre: I'm not anti using ... at all I just don't think it is needed in this particular context on both sides of the brackets, if you prefer happy to keep it in though!

docs/components/board/_index.md
JessamyT: @randhid Is this example applicable to i2c buses on all boards? TY!

docs/components/board/upboard.md
```diff
@@ -34,7 +34,46 @@ Enter a name for your board and click **Create**.
...
+{{< tabs >}}
+{{% tab name="Attributes template" %}}
+
+```json {class="line-numbers linkable-line-numbers"}
+{
+  "digital_interrupts": [
+    {
+      "name": "<your-digital-interrupt-name>",
+      "pin": "<pin-number>",
+      "type": "< basic | servo >"
```
JessamyT, line 49:
```suggestion
      "type": "<basic|servo>"
```

docs/components/board/upboard.md
```diff
@@ -44,11 +83,23 @@ Edit and fill in the attributes as applicable.
...
+      "type": "board",
+      "namespace": "rdk",
       "attributes": {
-        "i2cs": [],
-        "digital_interrupts": []
+        "digital_interrupts": [
+          {
+            "name": "<your-digital-interrupt-name>",
+            "pin": "<pin-number>",
+            "type": "< basic | servo >"
```
JessamyT, line 94:
```suggestion
            "type": "<basic|servo>"
```

docs/components/board/numato.md
```diff
@@ -23,7 +23,39 @@ Enter a name for your board and click **Create**.
...
+Copy and paste the following attribute template into your board's **Attributes** box.
+Then remove and fill in the attributes as applicable to your board, according to the table below.
+
+{{< tabs >}}
+{{% tab name="Attributes template" %}}
+
+```json {class="line-numbers linkable-line-numbers"}
+{
+  "pins": <int>,
+  "analogs": [
```
randhid, line 35: This is going to break soon and is currently unused.

docs/components/board/pca9685.md
```diff
@@ -30,12 +56,15 @@ Edit and fill in the attributes as applicable.
...
-      "name": "<your-pca9685-board>",
-      "type": "board",
+      "name": "<your-pca9685-board-name>",
       "model": "pca9685",
+      "type": "board",
+      "namespace": "rdk",
       "attributes": {
         "board_name": "<your-board-name>",
-        "i2c_name": "<your-bus-name>"
+        "i2c_name": "<your-bus-name>",
```
randhid, line 65: Will break soon.

docs/components/board/upboard.md
randhid: Only document this attribute (basic versus servo) for the raspberry pi.

static/include/components/board-attr-config.md
```diff
@@ -0,0 +1,60 @@
...
+{{% tab name="Attributes template" %}}
+
+```json {class="line-numbers linkable-line-numbers"}
+{
+  "analogs": [
```
randhid, line 9: These are all going away and we are not having a standard config for the board component. The common config is too tied to the raspberry pi, and we've already broken it for custom-linux.

randhid (changes requested):
So... we're about to break a bunch of board/sensor i2c things, would you be adverse to not mentioning the standard board config on any board other than the raspberry pi? The configs will be board-specific, and i2 and spi will disappear from the new board components.

I would also appreciate just hiding analogs everywhere until the Analog writer project.

Sorry about hat, cc @oliviamiller @penguinland

randhid:
> @randhid Ah okay this PR is not an attempt to change what is or isn't documented; it's just part of a larger project to give users chunks that are easier to copy and paste into non-fancy configs. I'd like to deal with the "will break soon" things in a separate PR unless absolutely necessary to do here. Waiting for GitHub to stop bugging, then I will fix the "type" stuff though since it sounds like that's already only for Pi.

Oky then I'd like the following:
analogs over configuring an mcp3008 chip over spi - pi only
digital interrupt types - pi only
only document how to configure i2c buses, spi buses and non-servo digital interrupts on jetson, beagle, upboard, and ti.

docs/components/board/_index.md
JessamyT: Okay I'm applying this such that it's only on the Pi-specific page per Rand, and taking out the spaces.

docs/components/board/_index.md
JessamyT: Sounds like this doesn't matter since it's all about to change

JessamyT:
Using the includes in some places for now but can easily take out again when more changes come down and things become more board-specific

docs/components/board/jetson.md
```diff
@@ -44,10 +44,21 @@ Edit and fill in the attributes as applicable.
...
+      "name": "<your-jetson-board-name>",
       "model": "jetson",
-      "attributes": {},
+      "type": "board",
+      "namespace": "rdk",
+      "attributes": {
+        "digital_interrupts": [
+          <...See table below...>
+        ],
+        "spis": [
```
randhid, line 55: we can either get ahead of removing spi and i2c attributes or wait until they're merged in for generic linux/custom linux drivers, up to you - cc @penguinland @oliviamiller

randhid (approved):
LGTM

PR #1933, opened 2023-09-29
DOCS-1129: Add clarifying note that color and depth streams can come from same camera in `join_color_depth` model

sguequierre:
* Adds note in attribute information-- not sure if should add to top as well? Can if @bhaney thinks appropriate


bhaney (approved):
LGTM

npentrel (approved):

PR #1939, opened 2023-10-02
DOCS-1201: Update `ultrasonic` sensor docs and document `ultrasonic` camera

sguequierre:
* Make images new 
* Make attributes includes and update reading to not say sensor
* Add `ultrasonic` camera page with usage tips


docs/components/sensor/_index.md
```diff
@@ -37,7 +37,7 @@ For configuration information, click on one of the supported sensor models:
 | Model | Description |
 | ----- | ----------- |
 | [`fake`](fake/) | A model used for testing, with no physical hardware. |
-| [`ultrasonic`](ultrasonic/) | [HC-S204 ultrasonic distance sensor](https://www.sparkfun.com/products/15569) |
+| [`ultrasonic`](ultrasonic/) | An ultrasonic distance sensor like the [HC-S204](https://www.sparkfun.com/products/15569) |
```
sguequierre, line 40: @bhaney Is this accurate? I realized that it may have been misleading to suggest one model of sensor in this place because this is such a common robotic component, but unsure. Can revert if you think more accurate

bhaney (approved):

docs/components/camera/ultrasonic.md
```diff
@@ -0,0 +1,88 @@
...
+description: "Configure an ultrasonic model camera."
+tags: ["camera", "components", "ultrasonic"]
+icon: "/icons/components/camera.svg"
+images: ["/icons/components/camera.svg"]
+# SME: #team-bucket
+---
+
+{{< alert title="Usage" color="tip" >}}
+An ultrasonic distance sensor can also be configured as a [sensor](/components/sensor/) resource.
+[Configure it as such](/components/sensor/ultrasonic/) to utilize the [sensor API](/components/sensor/#api) with your ultrasonic sensor hardware.
```
npentrel, line 15: Removes some duplication and the "as such" which feels a bit roundabout
```suggestion
An ultrasonic distance sensor can also be configured as a [sensor](/components/sensor/) resource, if you need to use the [sensor API](/components/sensor/#api) with your ultrasonic sensor hardware.
```

docs/components/sensor/ultrasonic.md
```diff
@@ -4,13 +4,18 @@ linkTitle: "ultrasonic"
...
+tags: ["sensor", "components", "ultrasonic"]
 icon: "/icons/components/sensor.svg"
 images: ["/icons/components/sensor.svg"]
 # SME: #team-bucket
 ---
 
-Configure an `ultrasonic` sensor to integrate an [HC-S204 ultrasonic distance sensor](https://www.sparkfun.com/products/15569) into your robot:
+{{< alert title="Tip" color="tip" >}}
+An ultrasonic distance sensor can also be configured as a [camera](/components/camera/) resource.
+[Configure it as such](/components/camera/ultrasonic/) to utilize the [camera API](/components/camera/#api) with your ultrasonic sensor hardware, enabling you to use it as an input to a [vision service](/services/vision/) model that returns obstacles.
```
npentrel, line 15: I think this might be slightly better
```suggestion
An ultrasonic distance sensor can also be configured as a [camera](/components/camera/) resource.
When configured as a camera, you can use the [camera API](/components/camera/#api) and you can use the component as an input to a [vision service](/services/vision/) model that returns obstacles.
```

docs/components/camera/ultrasonic.md
```diff
@@ -0,0 +1,88 @@
...
+
+Configure your ultrasonic sensor as a camera if you want to use it as an input to a [vision service](/services/vision/) model that returns obstacles.
+As a camera model, the ultrasonic model implements the camera method [`GetPointCloud()`](/components/camera/#getpointcloud), rather than the sensor method [`GetReadings()`](/components/sensor/#getreadings).
```
npentrel, line 18:
```suggestion
An ultrasonic distance sensor can also be configured as a [sensor](/components/sensor/) resource.
When configured as a sensor, you can use the [sensor API](/components/sensor/#api):

- [`GetReadings()`](/components/sensor/#getreadings)
```

sguequierre: Why not have the input to vision service obstacles part? @npentrel

sguequierre: Actually, I see what you mean-- makes more sense with the suggestion below for sensor page also calling that out. done

docs/components/sensor/ultrasonic.md
sguequierre: Done!

docs/components/sensor/_index.md
sguequierre: @randhid is it accurate to say "like the HC-S204" or should we just recommend HC-S204?

randhid: Ah yes our famous camera. So this technically refers to anything that uses an echo pin as an interrupt and a trigger pin interrupt. You can document that, there are people that will  read between the lines and grab another echo/trigger ultrasonic to use this module.  Calling out the HC-S204 is the most helpful thing for most people.

Also sorry for the name - maybe one day we can change this to its proper chip name. 

This model won't work with a single ultrasonic sensor without this chip - like the TCT40-16R, but might with another model that has both a trigger and echo pin - like the JSN-SR04T.

npentrel (approved):

PR #1946, opened 2023-10-03
DOCS-1079: Update camera config instructions

JessamyT:
For all camera config docs:
- Provide easily copy-pastable attribute templates (DOCS-1079)
- Update config builder screenshots to get rid of attribute guide (DOCS-1184) and to show realistic examples (where easily possible)
- Fix JSON indentation from 4 to 2
- Add `"namespace": "rdk"` and make name-model-type-namespace consistent with order in app (DOCS-1188)
- Move required attributes to the tops of tables and templates (above optional intrinsics etc.)
- Remove errant space after `"model"` where it applicable


npentrel (approved):
LGTM but before merging please reduce file sizes on the images you are adding - you can use https://tinypng.com/ for that and just add the images and download the result and replace them with that.

PR #1970, opened 2023-10-05
remove pwm_frequency_hz from pca9685 board config

penguinland:
# Description

This config value was literally unused in the code, and has no semantic meaning (I believe each pin on the board has its own frequency independent of the other pins). So, https://github.com/viamrobotics/rdk/pull/3032 removes the value from the config, and this PR updates the docs to match.

I wasn't sure who should review, so I picked the first person Github suggested. Feel free to reassign if that's wrong.


npentrel (approved):
Feel free to merge! Thank you!

PR #1971, opened 2023-10-05
DOCS-1189: Update obstacles depth attributes

npentrel:
Is `ground_plane_normal_vec` also present or is that only for the `obstacles_pointcloud` segmenter?


bhaney:
> Is `ground_plane_normal_vec` also present or is that only for the `obstacles_pointcloud` segmenter?

`ground_plane_normal_vec` is only for obstacles_pointcloud. The _depth model has a fixed vector.

bhaney (approved):

PR #1990, opened 2023-10-09
DOCS-1243: Make ultrasonic components consistent with others

JessamyT:
- Make both ultrasonic pages consistent with the recent config instructions updates
- Provide instructions on same within the component model template


sguequierre (approved):
lgtm great work!!

PR #1997, opened 2023-10-10
Remove cloud sync interruption warning

agreenb:
Remove a caution for data capture not resuming if cloud sync is disabled mid-sync. Data capture will continue regardless of cloud sync, and cloud sync will continue once reenabled.


AaronCasas (approved):

npentrel (approved):
Thank you!

PR #1998, opened 2023-10-10
DOCS-1076: Simplify MR landing pages

andf-viam:
First stab at refactoring MR landing pages.

- Initial dumping of remotes and subtypes into an `advanced` left-nav entry. This likely not ideal, but good to refocus on rest of MR stuff as we start considering options.
- This mostly just for team brainstorming!
- Caught many aliases, but probably have a few more to catch.


docs/modular-resources/key-concepts.md
```diff
@@ -16,8 +16,15 @@ description: "The key concepts behind how Viam's resource APIs and models are un
 no_list: true
 ---
 
-You can extend the features of an existing {{< glossary_tooltip term_id="resource" text="resource" >}} API by [creating a custom module](/extend/modular-resources/create/) to define a new [model](#models) that implements that API.
-A custom module can provide one or more modular resource models.
+Viam's [Robot Development Kit (RDK)](/internals/rdk/) provides built-in support for a variety of {{< glossary_tooltip term_id="resource" text="resources" >}}:
```
npentrel, line 19: I wonder if these few lines are more approachable than what’s on the top level page now but we can address that separately

npentrel (changes requested):
I think we are missing some aliases. Every page currently under extend needs an alias to where it was. Otherwise good!

docs/modular-resources/_index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Extend Viam with Modular Resources"
 linkTitle: "Modular Resources"
-weight: 10
+weight: 50
```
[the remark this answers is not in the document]
andf-viam: I've fixed this in [most recent commit](https://github.com/viamrobotics/docs/pull/1998/commits/bdc3101ebe20bf90af856ff742edfeaab1910b0b).

docs/modular-resources/key-concepts.md
andf-viam: Let's try it!

andf-viam:
@npentrel Thanks, you're absolutely right. Think I got all the missing aliases now!

docs/modular-resources/create/_index.md
```diff
@@ -14,21 +14,23 @@ tags:
   ]
 description: "Use the Viam module system to implement modular resources that can be included in any Viam-powered smart machine."
 no_list: true
+aliases:
+  - "/extend/modular-resources/create/create-subtype/"
```
npentrel, line 18: the create-subtype page is covered in that file
```suggestion
  - "/extend/modular-resources/create/"
```

npentrel (approved):
One remaining alias to fix

PR #2008, opened 2023-10-11
DOCS-1127: Add brief clarification for public vs private module state updating

sguequierre:
* Added clarification that you can't update a module from public to private re. feedback
* Split out create commands for private v. public CLI options


andf-viam (approved):
This LGTM, one optional language suggestion.

I think there was also some initial confusion about the default value for `visibility` early on (it was empty on purpose!), but with a configured default of `private` this largely removes the two things Ian ran into in his comment thread: an empty default, and trying to use public with an ord-id.

sguequierre:
> This LGTM, one optional language suggestion.
> 
> I think there was also some initial confusion about the default value for `visibility` early on (it was empty on purpose!), but with a configured default of `private` this largely removes the two things Ian ran into in his comment thread: an empty default, and trying to use `public` with an `org-id`.

@andf-viam sorry, am confused-- is there any other changes I can make to clear up confusion? Where is this empty default?

andf-viam:
@sguequierre Oh sorry! Not at all! I was just sharing some of where I think the initial confusion came about, before eng changes were made (new default: private, now populating this value instead of `""`).

No changes needed here! Sorry for confusion. This 👍 .

npentrel:
This needs rebasing. I think the bit that caused confusion is the `viam module create` command. There should be two options there:

```
1. Next, run the `viam module create` command to select a new custom module name and generate module metadata.

   {{< tabs >}}
{{% tab name="Private" %}}

Get the org id for your {{< glossary_tooltip term_id="organization" text="organization" >}} from your organization's **Settings** page in [the Viam App](https://app.viam.com/)  and run the following command  from the same directory as your custom module to generate metadata for your module:

viam module create --name <your-module-name> --org-id <your-org-id>

{{% /tab %}}
{{% tab name="Public" %}}

1. If you haven't already, [create a new namespace](/manage/fleet/organizations/#create-a-namespace-for-your-organization) for your organization.
   If you have already created a namespace, you can find it on your organization's **Settings** page in [the Viam App](https://app.viam.com/).

1. To generate metadata for your module using your public namespace, run the following command from the same directory as your custom module:

viam module create --name <your-module-name> --public-namespace <your-unique-namespace>

This command creates a new `meta.json` metadata file in your current working directory, which serves as a template.   Editing and then uploading the `meta.json` file sets important configuration information about your module, such as whether it will be publicly available to all Viam users, or only available within your organization.

{{% /tab %}}
{{< /tabs >}}
```


The docs for the CLI explain the options well I think: https://docs.viam.com/manage/cli/#using-the---org-id-and---public-namespace-arguments

docs/modular-resources/upload/_index.md
```diff
@@ -30,21 +30,37 @@ To upload your custom module to the [Viam registry](https://app.viam.com/registr
...
+2. Next, run the `viam module create` command to select a new custom module name and generate module metadata.
+   Run this command according to your desired visibility for your module:
 
-   1. If you haven't already, [create a new namespace](/manage/fleet/organizations/#create-a-namespace-for-your-organization) for your organization.
-      If you have already created a namespace, you can find it on your organization's **Settings** page in [the Viam App](https://app.viam.com/).
+{{< tabs >}}
+{{% tab name="Private" %}}
 
-   1. To generate metadata for your module using your public namespace, run the following command from the same directory as your custom module:
+Get the org id for your {{< glossary_tooltip term_id="organization" text="organization" >}} from your organization's **Settings** page in [the Viam App](https://app.viam.com/) and run the following command from the same directory as your custom module to generate metadata for your module:
```
andf-viam, line 39: Nit: Either `org-id` or "organization ID"

docs/modular-resources/upload/_index.md
```diff
@@ -30,21 +30,37 @@ To upload your custom module to the [Viam registry](https://app.viam.com/registr
 
 1. First, [install the Viam CLI](/manage/cli/#install) and [authenticate](/manage/cli/#authenticate) to Viam, from the same machine that you intend to upload your module from.
 
-1. Next, run the `viam module create` command to select a new custom module name and generate module metadata.
+2. Next, run the `viam module create` command to select a new custom module name and generate module metadata.
+   Run this command according to your desired visibility for your module:
```
andf-viam, line 34: I think this would be a good place to add that visibility is private by default (i.e. just before the choice of which to select)

andf-viam (approved):
Just wanted to mention which state is the default state somewhere before the tabbed breakdown of the two states.

docs/modular-resources/upload/_index.md
sguequierre, line 39:
```suggestion
Get the `org-id` for your {{< glossary_tooltip term_id="organization" text="organization" >}} from your organization's **Settings** page in [the Viam App](https://app.viam.com/) and run the following command from the same directory as your custom module to generate metadata for your module:
```

docs/modular-resources/upload/_index.md
sguequierre, line 34:
```suggestion
   Run this command according to your desired visibility for your module, which is by default private at creation:
```

sguequierre: @andf-viam This is the best I can think of and sounds super awkward. Any ideas to make it fit in better with the flow here?

npentrel (approved):

docs/modular-resources/upload/_index.md
sguequierre: It just seems a little confusing to mention it here-- I'm like "run this command for what you want" and then like "actually that doesn't matter because it's created as private then you edit the meta.json file to set visibility"

sguequierre: @andf-viam Edited, sounds a bit awkward with the flow but I can't think of a better point to introduce

PR #2010, opened 2023-10-11
Remove named dep from step 4

andf-viam:
VLC is a specific dependency needed if I coded some video thing in my robot code, not a generic always-add. Removing in favor of re-used example text in code: `other required dependencies`.


npentrel (approved):
😂

PR #2025, opened 2023-10-12
Change position_polling_frequency to position_polling_frequency_hz

tessavitabile:
Found a small error while looking over the docs to send to a customer.


sguequierre (approved):
LGTM, thanks for catching this Tess!!

PR #2036, opened 2023-10-13
DOCS-1025: Add control tab view for gripper

sguequierre:
* Adds control tab view for gripper component


andf-viam (approved):
LGTM, thanks!

PR #2046, opened 2023-10-16
DOCS-1026: Add consistent test control view for input controllers

sguequierre:
* Add includes control view
* Remove a lot of extra info from `webgamepad` and `gamepad` models former usage section that I think was more confusing than helpful, if want to keep can though 
* Uncomment/add attributes on `fake` model because I think that's actually useful + I tested them when using the fake input controller for this


assets/components/input-controller/input-controller-control-tab.png
npentrel, line 1: is this the fake one? I think in this one instance it would be very good to show what it looks like with a real one...

sguequierre: Ok I can do that! I'll borrow a gamepad

sguequierre: @npentrel done!

npentrel: That's well good!

PR #2048, opened 2023-10-16
DOCS-1154: Fix remote data capture config

JessamyT:


docs/services/data/configure-data-capture.md
```diff
@@ -203,10 +208,12 @@ The following example captures data from two analog readers that provide a volta
 {
   "services": [
     {
+      "name": "data_manager",
+      "type": "data_manager",
       "attributes": {
         "capture_dir": "",
         "sync_disabled": true,
-        "sync_interval_mins": 5,
+        "sync_interval_mins": 5
```
JessamyT, line 216:
```suggestion
        "sync_interval_mins": 5,
```

npentrel (approved):

docs/services/data/configure-data-capture.md
```diff
@@ -238,7 +245,7 @@ The following example captures data from two analog readers that provide a volta
               {
                 "method": "Gpio",
                 "capture_frequency_hz": 1,
-                "name": "rdk:component:board/board",
+                "name": "rdk:component:board/my-esp32",
                 "additional_params": {
                   "27": ""
```
agreenb, line 250: This should be grabbing the GPIO pins by name (again, this will be more clear with the refactor we're making now & I'll update the docs after), but then this should be "15"

Under the hood, this is calling [GPIOPinByName](https://github.com/viamrobotics/rdk/blob/d292ecc835cab4dc88c2e51e6529d40766111319/components/board/board.go#L62)

docs/services/data/configure-data-capture.md
```diff
@@ -227,7 +234,7 @@ The following example captures data from two analog readers that provide a volta
               {
                 "method": "Analogs",
                 "capture_frequency_hz": 1,
-                "name": "rdk:component:board/board",
+                "name": "rdk:component:board/my-esp32",
                 "additional_params": {
                   "A2": "",
```
agreenb, line 239: This should be grabbing the analog reader by name, so this should be "sensor"

Under the hood, this is calling [AnalogReaderByName](https://github.com/viamrobotics/rdk/blob/d292ecc835cab4dc88c2e51e6529d40766111319/components/board/board.go#L56)

agreenb (changes requested):
Thanks for your hard work to clean this up! Just some additional_params feedback, then LGTM

docs/services/data/configure-data-capture.md
JessamyT: Ooh i will change "sensor" to A1 above and add an A2

docs/services/data/configure-data-capture.md
```diff
@@ -171,25 +171,30 @@ To add them to your JSON configuration you must explicitly add the remote compon
...
+    {
+      "name": "my-esp32",
+      "model": "esp32",
+      "type": "board",
+      "attributes": {
+        "pins": [15],
+        "analogs": [
+          {
+            "pin": "34",
+            "name": "sensor"
```
JessamyT, line 191:
```suggestion
            "name": "A1"
```

docs/services/data/configure-data-capture.md
```diff
@@ -171,25 +171,30 @@ To add them to your JSON configuration you must explicitly add the remote compon
...
+          }
+        ]
```
JessamyT, line 193:
```suggestion
          },
          {
            "pin": "35",
            "name": "A2"
          }
        ]
```

docs/services/data/configure-data-capture.md
JessamyT: Also should I change the pin numbers to be actual ADC capable pins--pin 34 wouldn't work would it? Or does it not matter because we're using an external ADC?

docs/services/data/configure-data-capture.md
JessamyT: I was thinking 15 is the number of pins as opposed to a pin number, but I'm not sure where to find the info on what the esp32 config stuff even does/means

docs/services/data/configure-data-capture.md
agreenb: Don't have board-specific configuration answers — just know that if they're configured in a way that works at RDK reconfiguration time, they can have data captured off of them.

docs/services/data/configure-data-capture.md
agreenb: I would ask Nicolas about that. I think it's also completely reasonable to have a non-ESP32 board example here.

docs/services/data/configure-data-capture.md
```diff
@@ -171,25 +171,34 @@ To add them to your JSON configuration you must explicitly add the remote compon
...
-    "A2": "",
-    "A1": ""
-  }
+  "components": [
+    {
+      "name": "my-esp32",
+      "model": "esp32",
+      "type": "board",
+      "attributes": {
+        "pins": [15],
```
JessamyT, line 187:
```suggestion
        "pins": [27],
```

docs/services/data/configure-data-capture.md
JessamyT: changed it to 27 above

PR #2064, opened 2023-10-18
Make ESP32 board link to micro-RDK page in internals rather than setup

sguequierre:
Request from Rand/team to surface the internals page more, then the internals page links to setup

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -45,7 +45,7 @@ sitemap:
       </a>
     </li>
     <li id="c1_slide5">
-      <a href="/internals/micro-rdk/">
+      <a href="internals/micro-rdk/">
         {{<imgproc src="installation/thumbnails/esp32-espressif.png" resize="148x120" declaredimensions=true alt="E S P 32 - espressif">}}
         <p>Espressif ESP32</p>
       </a>
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -45,7 +45,7 @@ sitemap:
       </a>
     </li>
     <li id="c1_slide5">
-      <a href="installation/prepare/microcontrollers/">
+      <a href="/internals/micro-rdk/">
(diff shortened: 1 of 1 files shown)


randhid (approved):
LGTM!

docs/_index.md
```diff
@@ -45,7 +45,7 @@ sitemap:
       </a>
     </li>
     <li id="c1_slide5">
-      <a href="installation/prepare/microcontrollers/">
+      <a href="/internals/micro-rdk/">
```
npentrel, line 48:
```suggestion
      <a href="internals/micro-rdk/">
```

npentrel:
Why? That page has less information than the other one?

npentrel:
the top of the page is the same, is it not?

randhid:
> Why? That page has less information than the other one?

We need to point users to the central place where we're going to add the hardware boards we support, the APIs and component links are going to be edited to include the APIs and components that we actually support. This page is a good candidate to add this information, and you have a link to getting started.

Since the current page is hidden under viam-internals it's harder for users to find, and it would be hard to find the links to the component pages as well as the hardware supported - important to know before getting started at all, existent APIs etc.

npentrel:
ok - could wait till we have more boards but I don't mind if that's what you want

npentrel (approved):

docs/_index.md
npentrel: both will work on prod but this will make it work on staging

randhid:
> ok - could wait till we have more boards, but I don't mind if that's what you want

The thing is - they won't be different models - it's just going to expand hardware as we test it - there's no straight analogy to how boards operate in rdk - it won't be a different component model. I think this will look very different for GA in six months or so - it'll likely have to be placed in a new area of docs completely - but for now, this is a good compromise I think.

Thank you!

npentrel:
Sierra up to you if you want to merge now or later. Whatever is easiest

PR #2087, opened 2023-10-24
DOCS-1322: Suggest parameters, returns, descriptions for missing SDK methods

sguequierre:
Pulling more info automated


npentrel (commented):
you currently look for all objects that have the type `"sig sig-object py"`. I think you could instead look for objects that have the class `"py method"`. Then you know that the first child which also has the class `"sig sig-object py"` is the object where you can get the id from. But, more importantly, the second child of that object has the description text. And then you can also get the parameters with their descriptions from that I think. Can you give that a go?

npentrel:
Nice, now do you think you could change the formatting of the output to resemble what the docs need so that it's a simple copy and paste task?

sguequierre:
> Nice, now do you think you could change the formatting of the output to resemble what the docs need so that it's a simple copy and paste task?

@npentrel Tried to format it the best I can think of, LMK what you think!

npentrel:
So this is an example of what you're getting as the output
```
Method: viam.services.slam.client.SLAMClient.get_latest_map_info 
 Method Parameters, Returned, Description: 
Get the timestamp of the last update to the point cloud SLAM map. 
  -The timestamp of the last update. 
  -datetime
```

I think pulling the pieces apart and adding the markup we need for the docs should work to create something like the following output. The close we can get to something we can just copy and paste and have it work, the better

```
Get the timestamp of the last update to the point cloud SLAM map.

**Parameters:**
- `timeout` (Optional[float]):

**Returns:**
- `datetime` (datetime.datetime): The timestamp of the last update.

For more information, see the [Python SDK Docs](URL - which I think you can also get).
```

sguequierre:
Thank you, that makes sense! @npentrel I tried to update it to match your sample but this is the closest I can't get it. They use relative links on the page and I couldn't figure out how to pull the method url precisely, tried a couple selectors, just looks like what we already have for ID so for URL I am using page url. That ok for now?

.github/workflows/check_python_methods.py
```diff
@@ -50,19 +53,61 @@ def parse(type, names):
...
+
+            method_text.append("\n **Returns:** \n")
+
+            # Returns
+            return_element = tag_sigobject.find_all("span", class_="sig-return")
+            if return_element:
+                for element in return_element:
+                    method_text.append("- " + element.text + "\n")
+
+            method_text.append(f"\n For more information, see the [Python SDK Docs]({url})")
```
npentrel, line 104: You can make the full url with the id. You could also obtain the rest of the link from the headerlink class:
```suggestion
            method_text.append(f"\n For more information, see the [Python SDK Docs]({url}#{id})")
```

.github/workflows/check_python_methods.py
```diff
@@ -50,19 +53,61 @@ def parse(type, names):
...
+            #     index = 0
+            #     for element in description_elements:
+            #         if index >= 1:
+            #             method_text.append("-" + element.text + "\n")
+            #         else:
+            #             method_text.append(element.text + "\n")
+            #         index += 1
+            description_element = tag.find("p")
+            if description_element:
+                method_text.append(description_element.text + "\n\n")
```
npentrel, line 86: You're currently only getting the first line of the description. This will get it all
```suggestion
            description_elements = tag.find_all("p")
            if description_elements:
                for elem in description_elements:
                    method_text.append(elem.text + "\n")
            method_text.append("\n")
```

.github/workflows/check_python_methods.py
```diff
@@ -50,19 +53,61 @@ def parse(type, names):
...
+            # UNCOMMENT IF YOU WANT TO CATCH ALL EXTRA RETURNS INFORMATION/RETURNS DESCRIPTION
+            # description_elements = tag.find_all("p")
+            # if description_elements:
+            #     index = 0
+            #     for element in description_elements:
+            #         if index >= 1:
+            #             method_text.append("-" + element.text + "\n")
+            #         else:
+            #             method_text.append(element.text + "\n")
+            #         index += 1
```
npentrel, line 83:
```suggestion
```

npentrel, line 83:
```suggestion
```

PR #2090, opened 2023-10-25
DOCS-1297: Improve ML landing page

npentrel:


andf-viam (approved):
LGTM! Straightforward steps side by side with supporting images works really well, especially like the alternating left/right positioning -- might use this approach for Viam in 3 Minutes conceptual intro.

PR #2094, opened 2023-10-25
RSDK-5397 Use forks from viamrobotics not npmenard

acmorrow:
We will be moving the `esp-idf` fork from @npmenard's org to `viamrobotics`, so this PR updates the one place in the docs where we reference the old name. Note that we intend to transfer the repo, so GH should forward requests and the old name will continue to work, so this does not need an immediate deployment. This will only be merged after the transfer is actually completed.

### DIFF
--- a/docs/installation/prepare/microcontrollers/development-setup.md
+++ b/docs/installation/prepare/microcontrollers/development-setup.md
@@ -154,7 +154,7 @@ Clone Viam's fork of the ESP-IDF, the development framework for Espressif SoCs (
 ```sh { class="command-line" data-prompt="$"}
 mkdir -p ~/esp
 cd ~/esp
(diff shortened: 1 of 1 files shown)


sguequierre (approved):

acmorrow:
@sguequierre - Thanks for the LGTM. There seems to be a failing job, but I don't think it has anything to do with my change. Unfortunately that blocks my ability to squash and merge. Are you able to bypass the failing task and merge this for me?

sguequierre:
> @sguequierre - Thanks for the LGTM. There seems to be a failing job, but I don't think it has anything to do with my change. Unfortunately that blocks my ability to squash and merge. Are you able to bypass the failing task and merge this for me?

No problem, and the failing test isn't an issue! I will merge now!

PR #2105, opened 2023-10-26
DOCS-1302: Remove imu-vectornav

JessamyT:
Marking as draft instead of outright removing in case we get it working again in future


randhid (approved):

npentrel (approved):

randhid:
Actually, one thing - would you mind moving the Eva arm to draft? No one has been actively developing on it, and I'm worried it's in the same or worse state as it was when I last saw it move.

JessamyT:
Sure thing—made a ticket to do that in a separate PR.

PR #2106, opened 2023-10-27
DOCS-1197: Review modular resource docs

JessamyT:


docs/modular-resources/key-concepts.md
```diff
@@ -36,24 +36,21 @@ A module provides definitions for one or more pairs of [APIs](#valid-apis-to-imp
...
-A modular resource runs in the module process. This differs from built-in resources, which run as part of `viam-server`.
+A _modular resource_ is a resource that is provided by a [module](#modules), and not built into the RDK.
+A modular resource runs in the module process.
+This differs from built-in resources, which run as part of `viam-server`.
 
 ## Models
 
 A _model_ describes a specific implementation of a [resource](#resources) that implements (speaks) its [API](/program/apis/).
-Models allow you to control different instances of a resource with a consistent interface, even if the underlying implementation differs.
+Models allow you to control hardware or software of a similar category (such as motors) with a consistent interface (set of methods), even if the underlying implementation differs.
```
npentrel, line 53:
```suggestion
Models allow you to control hardware or software of a similar category, such as motors, with a consistent set of methods as an interface, even if the underlying implementation differs.
```

npentrel (approved):
Lgtm with one small comment

docs/modular-resources/configure.md
```diff
@@ -12,17 +12,17 @@ tags:
...
 
-You can extend Viam by adding a module on your robot that provides one or more {{< glossary_tooltip term_id="resource" text="modular resources" >}} ([components](/components/) or [services](/services/)).
+To use a [modular resource](/modular-resources/) on your robot, you need to:
 
-You can [add a module from the Viam registry](#add-a-module-from-the-viam-registry), or you can [add a local module](#local-modules).
-
-See [Key Concepts of Modular Resource APIs](/modular-resources/key-concepts/) for more information.
+1. Configure a {{< glossary_tooltip term_id="module" text="module" >}}, either one [from the registry](#add-a-module-from-the-viam-registry) or a [local module](#local-modules).
+   This makes the modular resource available to the robot.
+1. Then configure the modular resource itself.
```
andf-viam, line 25: You can add a modular resource directly without needing to do anything with the module that provides it. Most users will add the modular resource and not bother with any module-specific conf, in fact.

Conceptually - I agree - it makes sense to add the module, then add the modular resource it provides. But in practice, eng has designed the process so that the MRs can be added directly without needing awareness of its parent module.

To see in action:

App: Config > Create Component > `camera / realsense` > Add module > name it > Create > Poof module also added with default settings. No module conf required to start using realsense MR.

JessamyT: Hmm. It is very streamlined but as you say you do click "Add module" before configuring the resource itself. That button configures the module. 

The current version of this intro section talks all about modules and says nothing about configuring the resource itself, so it sounds like we need to change that in some way, just need to figure out in which way, does that sound right?

andf-viam: It does. Perhaps the heavy usage of 'Module' on this page should be 'modular resource' in more places, I could see that!

andf-viam: Although, I guess I'd like to see the Add module button read "Add modular X" (component or service), since that's actually what is taking place (we came here from X tab, not Modules tab. We are adding to X tab, not modules tab -- although the providing module is also added to modules behind the scenes. Ex: We see a new camera component appear when we click, not the intel-realsense module appear).

The text itself "Add module" seems like a stretch to consider as a configuration step. Configuration seems to me to be more update management, and the upcoming [module environment variable support](https://github.com/viamrobotics/docs/pull/2102). Those steps configure the _module specifically_, so I think there is some cause to differentiate between configuring a module and its modular resource.

The real fix is better button text though.

JessamyT: I thought "Add module" is correct because that button adds the module on the modules tab, and then you continue configuring the modular resource in place?

JessamyT:
In an attempt to address Andrew's feedback I ended up making bigger changes to /configure. Not sure if you still approve @npentrel  Moved order so that everything about adding _resources_ from the registry is in one place, distinct from changing the config of the _module_. Changed some headings to try to not overload the word "configure" since we use it interchangeably with "add" in much of the docs. And tried to disambiguate module vs resource some.

npentrel:
@JessamyT I'll unassign myself from the review and let Andrew handle this, sound like you two have it covered :)

docs/modular-resources/configure.md
```diff
@@ -49,17 +51,17 @@ To add a modular [component](/components/) from the Viam registry to your robot:
 
    {{<imgproc src="modular-resources/configure/add-component-by-name.png" resize="400x" declaredimensions=true alt="The add a component modal showing results for the intel realsense module when searching by the name 'realsense'">}}
 
-1. After selecting the modular component, click the **Add module** button, enter a name for your modular component, and click **Create** to add it to your robot.
+1. After selecting the modular component, click the **Add module** button, enter a name for your modular component, and click **Create** to add it to your robot's component config.
```
andf-viam, line 54: I would only use the shorthand when it matches the UI.

```suggestion
1. After selecting the modular component, click the **Add module** button, enter a name for your modular component, and click **Create** to add it to your robot's component configuration.
```

docs/modular-resources/configure.md
```diff
@@ -49,17 +51,17 @@ To add a modular [component](/components/) from the Viam registry to your robot:
...
 
    {{<imgproc src="modular-resources/configure/add-component-screen.png" resize="400x" declaredimensions=true alt="The add a component modal showing the intel realsense module pane, with the 'Add module' button shown">}}
 
    Be sure the modular component you select supports the [platform](/manage/cli/#using-the---platform-argument) you intend to use it with, such as `linux arm64`.
    You can see which platforms the module supports at bottom of the module information screen before you add it.
 
 When you add a module from the Viam registry, the custom modular component it provides appears under the **Components** subtab like any other component.
-You can also find [the module itself](#configure-a-module-from-the-viam-registry) listed as **Deployed** under the **Modules** subtab.
+You can also find [the underlying module](#edit-the-configuration-of-a-module-from-the-viam-registry) listed as **Deployed** under the **Modules** subtab.
```
andf-viam, line 62: Underlying!! Yes, this is the right word, perfect. Thanks for finding it!

docs/modular-resources/configure.md
```diff
@@ -78,66 +80,29 @@ To add a modular [service](/services/) from the Viam registry to your robot:
 
    {{<imgproc src="modular-resources/configure/add-service-by-name.png" resize="400x" declaredimensions=true alt="The add a component modal showing results for the mlmodelservice triton module when searching by the name 'triton'">}}
 
-1. After selecting the modular service, click the **Add module** button, enter a name for your modular service, and click **Create** to add it to your robot.
+1. After selecting the modular service, click the **Add module** button, enter a name for your modular service, and click **Create** to add it to your robot's service config.
```
andf-viam, line 83:
```suggestion
1. After selecting the modular service, click the **Add module** button, enter a name for your modular service, and click **Create** to add it to your robot's service configuration.
```

docs/modular-resources/key-concepts.md
```diff
@@ -130,7 +129,7 @@ In addition, you should chose a name for the `family` of your model based on the
   }
   ```
 
-A model with the `viam` namespace is always Viam-provided.
+The `viam` namespace is reserved for models provided by Viam.
```
andf-viam, line 132: Much better, thank you!

docs/modular-resources/configure.md
```diff
@@ -78,66 +80,29 @@ To add a modular [service](/services/) from the Viam registry to your robot:
...
 
-When you add a module to your robot, you can also configure how that module updates itself when a newer version becomes available from the Viam registry.
-By default, a newly-added module is set to pin to the specific patch release (**Patch (X.Y.Z)**) of the version you added, meaning that the module will _never automatically update itself_.
-
-If you wish to allow automatic module updates when a new version of the module becomes available in the Viam registry, you can set the **Version type** for your module in the **Modules** subtab.
-Updating to a newer version of a module brings new functionality and bug fixes, but requires restarting the module to apply the update.
-The following update options are available:
+Follow the same steps as when you added the first modular resource, clicking **Create component** or **Create service** as applicable.
+You will be prompted to click **Add module** again while configuring the resource, though no duplicate module will be added to the `modules` section of the config.
```
andf-viam, line 103:
```suggestion
You will be prompted to click **Add module** again while configuring the resource, though no duplicate module will be added to the `modules` section of the configuration.
```

docs/modular-resources/configure.md
```diff
@@ -12,19 +12,21 @@ tags:
...
-You can extend Viam by adding a module on your robot that provides one or more {{< glossary_tooltip term_id="resource" text="modular resources" >}} ([components](/components/) or [services](/services/)).
+To use a [modular resource](/modular-resources/) on your robot, you need to:
 
-You can [add a module from the Viam registry](#add-a-module-from-the-viam-registry), or you can [add a local module](#local-modules).
+1. Add a {{< glossary_tooltip term_id="module" text="module" >}}, either one [from the registry](#add-a-modular-resource-from-the-viam-registry) or a [local module](#local-modules).
+   This makes the modular resource available to the robot.
+1. Then add the modular resource itself.
 
-See [Key Concepts of Modular Resource APIs](/modular-resources/key-concepts/) for more information.
+These two steps automatically happen at the same time when you add a modular resource from the registry.
```
andf-viam, line 27: What do you think of this rephrase:

```suggestion
When you add a modular resource from the registry, the underlying module that provides it is automatically added at the same time.
To add a modular resource from a local module, you must add the module first.
```

- "is automatically added" rather than "time" so we explicitly note the (absence of) user agency: no action required for the underlying module when from registry.
- Bit more explicit on what to do for each path to adding a modular resource, especially since we just introduced the two just above. Now we state how to initially interact with modules for both as well.

docs/modular-resources/key-concepts.md
```diff
@@ -36,24 +36,21 @@ A module provides definitions for one or more pairs of [APIs](#valid-apis-to-imp
...
-A modular resource runs in the module process. This differs from built-in resources, which run as part of `viam-server`.
+A _modular resource_ is a resource that is provided by a [module](#modules), and not built into the RDK.
+A modular resource runs in the module process.
+This differs from built-in resources, which run as part of `viam-server`.
 
 ## Models
 
 A _model_ describes a specific implementation of a [resource](#resources) that implements (speaks) its [API](/program/apis/).
-Models allow you to control different instances of a resource with a consistent interface, even if the underlying implementation differs.
+Models allow you to control hardware or software of a similar category, such as motors, with a consistent set of methods as an interface, even if the underlying implementation differs.
```
andf-viam, line 53: "with a consistent set of methods as an interface" sounds like it's missing a direct object. Is there a way to split this sentence to help clarify what we mean?

docs/modular-resources/configure.md
```diff
@@ -275,7 +280,7 @@ You can also add the module directly, without first adding its modular component
 
    This example shows the configuration for adding a [CSI camera](/modular-resources/examples/csi/) as a local module.
 
-## Configure a local module
+### Edit local module configuration
 
 Once you have added a modular resource to your robot, you can view and configure the module itself from the **Modules** subtab:
```
andf-viam, line 285:
```suggestion
Once you have added a modular resource to your robot, you can view and edit the underlying module from the **Modules** subtab:
```

I like "underlying"! 👍

docs/modular-resources/configure.md
```diff
@@ -275,7 +280,7 @@ You can also add the module directly, without first adding its modular component
 
    This example shows the configuration for adding a [CSI camera](/modular-resources/examples/csi/) as a local module.
 
-## Configure a local module
+### Edit local module configuration
```
andf-viam, line 283:
```suggestion
### Edit the configuration of a local module
```

andf-viam (approved):
LGTM, thanks for the help!

I've left some tiny language suggestions, but overall I think this really helps -- thank you!

docs/modular-resources/configure.md
andf-viam: I think you've really helped to clarify things with the latest edit. "Underlying module" is the language I've been looking for all this time, ha! Your callout about the module being added at the same time as the modular resource I think is also a great step forward, but have added a slight language suggestion as well. Thanks for thinking through this with me!

docs/modular-resources/configure.md
JessamyT: good points. I like it.

docs/modular-resources/key-concepts.md
JessamyT:
```suggestion
Models allow you to use the same set of methods to control hardware or software of a similar category, even if the underlying implementation differs.
```
Is this better?

andf-viam: Perfect!

PR #2123, opened 2023-10-30
Replace wiring diagram

mestcihazal:
The ultrasonic wiring numbers were 44, 44 which is wrong. Replaced the diagram with the correct 38, 40.


HipsterBrown (approved):
LGTM 🚀

npentrel (approved):

PR #2124, opened 2023-10-31
Change alpha to beta in micro rdk docs

randhid:
# Description

Changes alpha to beta in micro-rdk docs.

I think this is all that's needed. From my understanding this file is included in a few other places and generates the warning. Will double check when the link is generated.


sguequierre (approved):
Thanks rand!!

PR #2135, opened 2023-11-01
Update spacing in python sdk coverage script

npentrel:


.github/workflows/check_python_methods.py
```diff
@@ -111,11 +111,11 @@ def parse(type, names):
                     description_extras[field] = extras_values
 
                 # Parameters
-                method_text.append("**Parameters:** \n")
+                method_text.append("**Parameters:**\n\n")
                 no_parameters = True
                 if "Parameters:" in description_extras:
                     no_parameters = False
-                    extras_values = description_extras["Parameters:"]
+                    extras_values = description_extras["**Parameters:**\n\n"]
```
sguequierre, line 118: @npentrel Sorry I couldn't suggest so I figured it'd be ok to commit directly in this case, added spacing here too because it was formatting some just blank

sguequierre: Actually that might not be correct let me check

sguequierre: Sorry, this did not fix the problem -- reverting

sguequierre, line 118:
```suggestion
                    extras_values = description_extras["Parameters:"]
```

.github/workflows/check_python_methods.py
```diff
@@ -142,12 +142,12 @@ def parse(type, names):
...
 
                 # Returns
-                method_text.append("**Returns:** \n")
+                method_text.append("**Returns:**\n\n")
                 if "Returns:" in description_extras:
                     # Return type
                     if "Return type:" in description_extras:
                         return_type = html_to_markdown(url, description_extras["Return type:"]).strip()
-                        return_info = html_to_markdown(url, description_extras["Returns:"])
+                        return_info = html_to_markdown(url, description_extras["**Returns:**\n\n"])
```
sguequierre, line 150:
```suggestion
                        return_info = html_to_markdown(url, description_extras["Returns:"])
```

sguequierre (approved):
This isn't working for spacing when there's `None` parameters, tried to debug but I can't figure out why. Ok with merging as is though as NBD

npentrel:
Fixed I think

PR #2141, opened 2023-11-01
DOCS-1360: Fix image spacing on Raspberry Pi install page

sguequierre:
* Bug Hazal noticed-- not getting same issue with this markdown image format and the size isn't a big difference, but I'm not entirely sure why it was happening with `imgproc`.

### DIFF
--- a/docs/installation/prepare/rpi-setup.md
+++ b/docs/installation/prepare/rpi-setup.md
@@ -50,9 +50,13 @@ To install Raspberry Pi OS (formerly called Raspbian) on a microSD card from whi
 
 2.  Select `CHOOSE OS`.
     Click on `Raspberry Pi OS (other)`.
-    {{< imgproc alt="Raspberry Pi Imager window showing Raspberry Pi OS (Other) is selected." src="/installation/rpi-setup/select-other-custom-os.png" resize="800x" declaredimensions=true >}}
+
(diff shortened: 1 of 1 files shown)


docs/installation/prepare/rpi-setup.md
andf-viam, line 1: Good catch! But I think we just need to introduce spacing between the images. Retaining the `imgproc` gives us good resizing. Think all we need is this:


```
2. Select `CHOOSE OS`.
   Click on `Raspberry Pi OS (other)`.

   {{< imgproc alt="Raspberry Pi Imager window showing Raspberry Pi OS (Other) is selected." src="/installation/rpi-setup/select-other-custom-os.png" resize="800x" declaredimensions=true >}}

   Select `Raspberry Pi OS Lite (64-bit)` from the  menu.

   {{< imgproc alt="Raspberry Pi Imager window showing Raspberry Pi OS Lite (64-bit) is selected." src="/installation/rpi-setup/select-other-rpi.png" resize="800x" declaredimensions=true >}}
   
   You should be brought back to the initial launch screen.
```

andf-viam (changes requested):
We can fix this without losing the image resizing, see my comment. Thanks for catching!

andf-viam (approved):

PR #2142, opened 2023-11-01
DOCS-1112: Clarify RPi raspi-config

andf-viam:
Add explicit recommendations on which protocols to enable for your Rpi, based on which hardware you are using, without mentioning the Rover.
- Rover set up doc already links [directly to](https://docs.viam.com/try-viam/rover-resources/rover-tutorial/#install-raspberry-pi-os) this edited section, so also benefits from this clarification.

GUESS: "will likely need to enable the serial port" is just 'cause most of our ( movement | power ) sensors require a `serial_path`. Did I guess right?

### DIFF
--- a/docs/installation/prepare/rpi-setup.md
+++ b/docs/installation/prepare/rpi-setup.md
@@ -186,10 +186,10 @@ These protocols are required to support certain hardware, such as analog-to-digi
 
 1. Enable the relevant protocols to support your specific hardware:
 
-   - If you are using an analog-to-digital converter (ADC), enable SPI.
-   - If you are using an accelerometer, enable I2C.
-   - If you are using a CSI v1.3 or v2.0 camera, enable legacy camera support.
-   - If you are using a movement sensor or power sensor, you will likely need to enable the serial port.
+   - If you are using an analog-to-digital converter (ADC), enable **SPI**.
+   - If you are using an accelerometer, enable **I2C**.
+   - If you are using a CSI v1.3 or v2.0 camera, enable **Legacy Camera** support.
+   - If you are using a movement sensor or power sensor, you will likely need to enable the **Serial Port**.
      Check the documentation for your specific [movement sensor](/components/movement-sensor/) or [power sensor](/components/power-sensor/) to confirm.
 
 1. Then, to apply the changes, restart your Raspberry Pi if it hasn't already prompted you to do so.
--- a/docs/installation/prepare/rpi-setup.md
+++ b/docs/installation/prepare/rpi-setup.md
@@ -172,26 +172,31 @@ sudo apt upgrade
 ## Enable Communication Protocols
 
 If you are using hardware that requires I2C, SPI, serial, or one-wire protocols to communicate with your Pi, you will need to enable them using `raspi-config`.
+These protocols are required to support certain hardware, such as analog-to-digital converters (ADCs), accelerometers, and sensors.
 
-Launch the config tool by running the following command:
+1. Launch the config tool by running the following command:
 
-```sh {class="command-line" data-prompt="$"}
-sudo raspi-config
-```
+   ```sh {class="command-line" data-prompt="$"}
+   sudo raspi-config
+   ```
 
-Use your keyboard to select "Interface Options" and enable the relevant protocols.
+1. Use your keyboard to select "Interface Options", and press return.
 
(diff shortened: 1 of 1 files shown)


docs/installation/prepare/rpi-setup.md
```diff
@@ -172,26 +172,31 @@ sudo apt upgrade
 ## Enable Communication Protocols
 
 If you are using hardware that requires I2C, SPI, serial, or one-wire protocols to communicate with your Pi, you will need to enable them using `raspi-config`.
+These protocols are required to support certain hardware, such as analog-to-digital converters (ADCs), accelerometers, and sensors.
```
JessamyT, line 175: [opt] To me, the current way sounds almost redundant/odd because it's essentially saying twice that it's required. I think I like the second suggestion  best because it keeps the lists in separate sentences whilst introducing the why before the how, though it makes "them" ambiguous.
```suggestion
Certain pieces of hardware, such as analog-to-digital converters (ADCs), accelerometers, and sensors, communicate with your Pi using I2C, SPI, serial, or one-wire protocols.
If you are using hardware that requires any of these protocols, you need to enable them using `raspi-config`:
```
```suggestion
Certain hardware requires I2C, SPI, serial, or one-wire protocols to communicate with your Pi.
If you are using hardware such as an analog-to-digital converter (ADC), accelerometer, or sensor that requires any of these protocols, you need to enable them using `raspi-config`:
```

JessamyT (approved):
LGTM % consideration of rewording of the one bit. Though I'm no SME with respect to legacy cameras etc.

docs/installation/prepare/rpi-setup.md
mcvella: I slightly like the first way better

docs/installation/prepare/rpi-setup.md
```diff
@@ -172,26 +172,31 @@ sudo apt upgrade
...
+   {{< imgproc alt="Screenshot of the Raspi Config screen with a red box and red arrow pointing to the '3 Interface Options' option where you can find the I2C and other drivers" src="/installation/rpi-setup/Installation-Raspberry-Pi-I2C-Raspi-Config-Interfacing-Options.png" resize="800x" declaredimensions=true >}}
 
-{{< alert title="Important" color="note" >}}
-When using a CSI v1.3 or v2.0 camera, you need to enable legacy camera support.
-{{< /alert >}}
+1. Enable the relevant protocols to support your specific hardware:
 
-For these changes to take effect, you need to restart your Raspberry Pi if it hasn't already prompted you to do so.
+   - If you are using an analog-to-digital converter (ADC), enable **SPI**.
+   - If you are using an accelerometer, enable **I2C**.
```
mcvella, line 190: I feel this is too specific, LOTS of hardware use i2c.  I think we need to either give specific examples (and say "for example") or just tell them to check their hardware specs

docs/installation/prepare/rpi-setup.md
```diff
@@ -172,26 +172,31 @@ sudo apt upgrade
...
+   - If you are using a CSI v1.3 or v2.0 camera, enable **Legacy Camera** support.
+   - If you are using a movement sensor or power sensor, you will likely need to enable the **Serial Port**.
```
mcvella, line 192: Actually some use i2c.  And that's kind of the point, it really varies by hardware.  Some even have both i2c and spi interfaces.

mestcihazal (approved):
looks good to me as long as we apply jessamy's feedback and in the section "If you are using an etc. etc." we say 'for example if you are using an etc. etc. you may want to etc. etc.' so its clear that its not just those hardware but those are just given as examples

andf-viam:
Updated with feedback, thanks!

docs/installation/prepare/rpi-setup.md
```diff
@@ -171,27 +171,33 @@ sudo apt upgrade
 
 ## Enable Communication Protocols
 
-If you are using hardware that requires I2C, SPI, serial, or one-wire protocols to communicate with your Pi, you will need to enable them using `raspi-config`.
+Certain hardware, such as analog-to-digital converters (ADCs), accelerometers, and sensors, communicate with your Pi using specialized communications protocols, including I2C, SPI, serial, or one-wire protocols.
```
JessamyT, line 174: [nit] "Certain hardware" is singular but then the list is of plural items so "communicate" works for the plural but not the singular.

PR #2151, opened 2023-11-02
DOCS-1354: Update setup tab instructions in prepare microcontrollers page

sguequierre:
Add instructions and remove download information now that that's handled in setup tab


npentrel (approved):
nice

gvaradarajan (approved):
Perfect! Thank you!

PR #2163, opened 2023-11-03
Do not flag unimplemented methods

npentrel:
I was checking for new cloud methods and was halfway through adding boilerplate when I realized they were all unimplemented. So I changed the script to check for that.  Leaving the boilerplate in to make it easier to add them when we do


sguequierre (approved):
LGTM!

PR #2173, opened 2023-11-06
DOCS-1247: Document Navigation `Paths()`

sguequierre:
* Document `Paths` method of navigation service

### DIFF
--- a/docs/services/navigation/_index.md
+++ b/docs/services/navigation/_index.md
@@ -583,7 +583,7 @@ obstacles = await my_nav.get_obstacles()
 
 ### Paths
 
-Get each path, a series of geo points the robot will travel through, in the robot's motion planning.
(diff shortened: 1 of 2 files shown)


raybjork (approved):
I have questions about if the Python docs are up to date or not because I havent been keeping close track of how that SDK is set up.  Is there no extra param that function takes?

sguequierre:
> I have questions about if the Python docs are up to date or not because I havent been keeping close track of how that SDK is set up. Is there no extra param that function takes? Other than that it looks good

From the autogen site it doesn't look like it does, normally would be in https://python.viam.dev/autoapi/viam/services/navigation/client/index.html#viam.services.navigation.client.NavigationClient.get_paths

npentrel (approved):

PR #2182, opened 2023-11-08
DOCS-1382: Switch order of Python and Go samples in ML model and navigation services to match rest of docs

sguequierre:
Noticed out of sync, thought would align


andf-viam (approved):
LGTM! Thanks for catching this!!

PR #2204, opened 2023-11-13
Update segmentation.md

bhaney:
# With_Geometries is not used

This description of how "with_geometries" works for `obstalces_depth` was removed in the final iteration of the scope doc. We don't have a "with_geometries" attribute any more.


sguequierre (approved):
LGTM, and good catch

PR #2214, opened 2023-11-14
RSDK-5612: Correct the use of the logging package name.

dgottlieb:
* There was a typo in the package name. `logger.NewDevelopmentalLogger` and `logger.Logger` should have been `logging.*`.
* Steer away from the Developmental logger as an API. Developmental for viam employees is not the same as external users.
* I noticed the "go in a square" example still did not compile. The example code was missing trailing commas.

### DIFF
--- a/docs/components/input-controller/_index.md
+++ b/docs/components/input-controller/_index.md
@@ -214,7 +214,7 @@ func handleController(controller input.Controller) {
 }
 
 func main() {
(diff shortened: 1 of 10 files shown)


npentrel (approved):
Thank you! I thought we had recently fixed all of the loggers but I guess to the wrong thing...

PR #2227, opened 2023-11-16
DOCS-1045: Add filter module tutorial pt2

skyleilani:
Adding an image to the data filtering module tutorial.


andf-viam (approved):
LGTM! Thanks for adding `languages` too!

PR #2233, opened 2023-11-17
Change spelling of dropdown

npentrel:


andf-viam (approved):
LGTM! Agree with Vale changes too!

PR #2236, opened 2023-11-20
Reduce image file size

npentrel:


andf-viam (approved):
LGTM, thanks for fixing!

PR #2255, opened 2023-11-27
Update dataset docs

npentrel:


docs/data/dataset.md
```diff
@@ -115,9 +115,14 @@ You can click on any image in the dataset and use its appearing menu in order to
 
 To train a model on a dataset, see [Train a model](/ml/train-model/).
 
+### Rename a dataset
+
+To rename a dataset, go to the [**DATASETS** subtab](https://app.viam.com/data/datasets).
```
agreenb, line 120: For both rename and delete, you can also do so from the dataset page itself by clicking on the three-dot icon next to the dataset name.

agreenb (approved):
Just one thing then LGTM!

PR #2256, opened 2023-11-27
DOCS-1450: Add upload to customlinux

andf-viam:
Add mention of `viam board upload` on `customlinux` and Pumpkinboard (which uses `customlinux`) board reference pages.
- Changing term "pin mappings file" to "board definitions file" to match attribute name / CLI command syntax & help.

QUESTIONS:
- Does this need to be on Pumpkin board as well, or is that too much given that Pumpkin already prominently links to `customlinux`?


npentrel (approved):
looks good but maybe use an include? Worried that somethign will change and we'll not update it on the board page. I think fine to not include on the pumpkin page

andf-viam:
Removed most content from Pumpkin, but left "board definitions file" rename. I think we're OK without an include if the steps are then only on the board page itself.

PR #2267, opened 2023-11-28
DOCS-1463: Update micro-RDK GPIO motor attributes

sguequierre:
* Fix `en_low`
* Add new pin option


docs/build/micro-rdk/motor/gpio.md
```diff
@@ -51,14 +51,15 @@ Edit and fill in the attributes as applicable.
       "namespace": "rdk",
       "attributes": {
         "pins": {
-          <...>
+          "dir": "<int>",
+          "pwm": "<int>",
+          "en_low": "<int>"
```
gvaradarajan, line 56: actually we should just remove `en_low`, we're not going to support this in micro-RDK (also applies to L. 136 and L. 143 below)

gvaradarajan (approved):
LGTM. There's an issue because I forgot to update the doc explanation based on a conversation had [here](https://github.com/viamrobotics/micro-rdk/pull/115#issuecomment-1822985140)

docs/build/micro-rdk/motor/gpio.md
sguequierre: Ok sounds good!

PR #2271, opened 2023-11-29
DOCS-1361: Add labels to changelog

JessamyT:
Still need to figure out how to make it so that code-styled parts (for example, `extra`) appear in monospace within the heading.

Looks like this: 
image


npentrel:
Ok this looks pretty and is very close but functionally has some issues. You are getting javascript errors on the page (the anchors are the problem - the ids you are setting on the headers are not what they should be). You can solve that issue in two ways (by making it create ids OR with a solution that will also fix the code styling). Have a look at the manualcard.html file for inspiration. 

Lastly I think the span should be outside the header element. That should also then allow you to align them a little better (the label boxes are a bit low compared to the headers themself) :)

If you cannot figure it out let me know. I did some poking already.

JessamyT:
Yeah exactly :( I thought that was the case RE headers but was hoping there'd be some magic to fix it hehe.

Okie doke.

assets/scss/_styles_project.scss
```diff
@@ -1539,3 +1539,39 @@ input[type=search]::-webkit-search-cancel-button {
...
+
+// Changelog labels
+
+.changelog-outer {
+  display: flex;
+}
+
+.changelog div {
+  padding-right: 4px;
+}
```
npentrel, line 1551: I think you should be able to merge these
```suggestion
.changelog div {
  display: flex;
  padding-right: 4px;
}
```

layouts/shortcodes/changelog.html
```diff
@@ -0,0 +1,10 @@
+{{ $color := .Get "color" }}
+{{ $title := .Get "title" }}
+<div class="changelog changelog-outer">
+  <div >
+    <span class="changelog-{{ $color }}">{{ $color }}</span>
+  </div>
+  <div >
```
npentrel, line 7: nit
```suggestion
  <div>
    <span class="changelog-{{ $color }}">{{ $color }}</span>
  </div>
  <div>
```

docs/appendix/changelog.md
```diff
@@ -12,7 +12,7 @@ aliases:
 
 ## November 2023
 
-### Changed: Model training from datasets
+{{% changelog color="changed" title="Model training from datasets" %}}
 
 To make it easier to iterate while training machine learning models from image data, you now train models from [datasets](/data/dataset/).
```
npentrel, line 18: I think you're missing one on line 19

JessamyT: Ooh a recent addition--good catch

PR #2276, opened 2023-11-29
Update the mongosh command to connect the Viam org's MongoDB Atlas instance

agreenb:
The previous command did not work; updated with one that does.


andf-viam (approved):
Thanks!

dmhilly (approved):
LGTM! Thanks Alexa!

PR #2287, opened 2023-12-04
DOCS-1466: Add consistent vis names

andf-viam:
Standardize vision service names on `<vision-service-name>` for templates and `my_vision_service` for examples.
- Leaving screenshot (step 4) alone for now, we will return to App UI screenshots after upcoming robot details page redesign, I think.


andf-viam:
@stevebriskin Hi Steve super simple vision service name standardization for these examples, per your ticket. Let me know if you'd like to see any changes -- thanks!

docs/ml/vision/detection.md
```diff
@@ -181,7 +181,7 @@ In your vision service's panel, fill in the **Attributes** field.
 
 ```json {class="line-numbers linkable-line-numbers"}
 {
-  "mlmodel_name": "<detector_name>"
+  "mlmodel_name": "<vision-service-name>"
```
stevebriskin, line 184: i think this is incorrect...should the ml service name i believe

docs/ml/vision/detection.md
```diff
@@ -197,7 +197,7 @@ Add the vision service object to the services array in your raw JSON configurati
     "type": "vision",
     "model": "mlmodel",
     "attributes": {
-      "mlmodel_name": "<detector_name>"
+      "mlmodel_name": "<vision-service-name>"
```
stevebriskin, line 200: same

docs/ml/vision/detection.md
```diff
@@ -213,7 +213,7 @@ Add the vision service object to the services array in your raw JSON configurati
     "type": "vision",
     "model": "mlmodel",
     "attributes": {
-      "mlmodel_name": "person_detector"
+      "mlmodel_name": "my_vision_service"
```
stevebriskin, line 216: same. but why no <>?

docs/ml/vision/detection.md
```diff
@@ -247,7 +247,7 @@ If you intend to use the detector with a camera that is part of your robot, you
          "type": "detections",
          "attributes": {
            "confidence_threshold": 0.5,
-           "detector_name": "my_detector"
+           "detector_name": "my_vision_service"
```
stevebriskin, line 250: i believe this is right (it's the vision service), but should have <> for consistency.

stevebriskin (changes requested):
defering to @bhaney

docs/ml/vision/detection.md
andf-viam: The brackets in these examples represent placeholder values for the JSON Template tab, while example strings are used under the JSON Example tab. Compare the two JSON tabs [here](https://docs.viam.com/components/camera/webcam/) for example.

stevebriskin: ah, got it.

npentrel (approved):
Approved once Steve approves. Thanks for changing these!

docs/ml/vision/detection.md
bhaney: agreed, <camera-name> has brackets below, but this one has quotes

bhaney (changes requested):

docs/ml/vision/detection.md
andf-viam: Ah I see now, updated with brackets for `my_vision_service` then, you're both right here.

bhaney (approved):

PR #2297, opened 2023-12-06
DOCS-1366: Update I2C attributes for `bme280` sensor

sguequierre:
* Make index 
* Remove `board` attribute


JessamyT (approved):
LGTM from a docs perspective except, maybe we need to provide a sentence somewhere in here explaining why you don’t need to specify the board? Not sure. I am personally confused as to how that works and users might be too since we do require board specification for many other components.

penguinland (approved):
LGTM! It might help Jessamy's confusion to remove the "-on-board" from the templates, so you never even imply that this is related to a board component. but either way works for me!

JessamyT:
Per convo with Alan offline I’d say disregard my comment—good to go as-is!

PR #2300, opened 2023-12-06
DOCS-1389: Update TMC5072 motor attributes

JessamyT:


npentrel (approved):

docs/components/motor/tmc5072.md
```diff
@@ -86,12 +86,11 @@ Edit and fill in the attributes as applicable.
       "type": "motor",
       "namespace": "rdk",
       "attributes": {
-        "board": "example-board",
-        "chip_select": "36",
-        "index": 0,
+        "chip_select": "0",
+        "index": 1,
```
penguinland, line 90: Dang, very good catch here! I'd totally forgotten than 0 is not a valid index.

docs/components/motor/tmc5072.md
```diff
@@ -107,9 +106,8 @@ The following attributes are available for `TMC5072` motors:
 <!-- prettier-ignore -->
 | Name | Type | Inclusion | Description |
 | ---- | ---- | --------- | ----------- |
-| `board` | string | **Required** | `name` of the [board](/components/board/) to which the motor controller is wired. |
 | `spi_bus` | string | **Required** | The board [SPI bus](/components/board/#spis) over which the TMC chip communicates with the board. |
-|`chip_select` | string | **Required** | The {{< glossary_tooltip term_id="pin-number" text="pin number" >}} of the GPIO pin on the board wired to the TMC controller. The board sets this high or low to let the TMC chip know whether to listen for commands over SPI. |
+|`chip_select` | string | **Required** | The chip select number (CSN) that the TMC5072 is wired to. For Raspberry Pis, use `"0"` if the CSN is wired to {{< glossary_tooltip term_id="pin-number" text="pin number" >}} 24 (GPIO 8) on the Pi, or use `"1"` if you wire the CSN to pin 26. The board sets this high or low to let the TMC chip know whether to listen for commands over SPI. |
```
penguinland, line 110: Although this information is correct, it's unclear to me whether we need it here. These are the standard pins with the standard chip select numbers for almost all SBCs, and if Viam as a company wasn't so focused on the rpi4 and libpigpio, this would already be second-nature for us.

...and yet, because we have been so rpi4-centric, maybe we do need to add this in, at least for a year or two. I dunno; maybe this PR is perfect as-is.

JessamyT: Good points!

PR #2311, opened 2023-12-08
DOCS-1424: Update module cwd

andf-viam:
Update `VIAM_MODULE_ROOT` description to indicate that it now serves as the `cwd` of the module process.
- Updated example path to no longer be based on `VIAM_HOME`

QUESTION:
- Is it still true that local module users must set `VIAM_MODULE_ROOT` manually if needed? Wasn't sure from scope / ticket / RDK PR. Thank you!

### DIFF
--- a/docs/registry/configure.md
+++ b/docs/registry/configure.md
@@ -320,7 +320,7 @@ When a module is instantiated, it has access to the following default environmen
 | Name | Description |
 | ---- | ----------- |
 | `VIAM_HOME` | The root of the `viam-server` configuration.<br>Default: `$HOME/.viam` |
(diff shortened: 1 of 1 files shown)


andf-viam:
@zaporter-work Hi Zack, here's a quick update to capture your work in https://github.com/viamrobotics/rdk/pull/3227. Small one! I left a question in the PR description. Let me know if you'd like to see any changes -- thank you!

zaporter-work (approved):
LGTM! 

For now, modules are still in `$VIAM_HOME/packages/.data/...`

However, I like the idea of not using that path so that people don't accidentally start developing their modules in $VIAM_HOME.. which would be bad.

andf-viam:
@zaporter-work ahh gotcha, thanks for explanation, that makes sense!

npentrel (approved):

PR #2320, opened 2023-12-12
DOCS-1510: Replace RPi5 config screenshot

JessamyT:
SCUTTLE tutorial doesn't use this image anymore anyway so moving it to the appropriate place, deleting the outdate RPi screenshot, and adding an RPi5 screenshot now that that new model is available in the app.


andf-viam (approved):
LGTM, thank you!

PR #2321, opened 2023-12-13
DOCS-1512: Remove I2Cs and SPIs from tutorials and other places

sguequierre:
* Update tutorials -- tipsy doesn't appear to have needed/been using I2C initially so I removed entirely there
* Scrub from remaining places in docs


docs/appendix/glossary/board.md
```diff
@@ -2,11 +2,11 @@
 title: Board
 id: board
 full_link: /components/board/
-short_description: A board is the signal wire hub of a machine that provides access to GPIO pins, SPI or I2C.
+short_description: A board is the signal wire hub of a machine that provides access to GPIO pins.
 aka:
 ---
 
-A board is the signal wire hub of a machine that provides access to GPIO pins, SPI or I2C.
+A board is the signal wire hub of a machine that provides access to GPIO pins.
```
penguinland, line 9: Off-topic question: what's a single wire hub? Boards have multiple wires, which makes me think I've misunderstood. Should that part be rephrased? Same (still off-topic) question on line 5.

docs/tutorials/configure/scuttlebot.md
```diff
@@ -119,7 +88,7 @@ Add the following JSON objects to the `components` array:
       "board": "local",
```
penguinland, line 88: You can remove this line; the encoder no longer depends on the board. Same comment on line 102, below.

docs/tutorials/configure/scuttlebot.md
```diff
@@ -36,42 +36,11 @@ cost: 540
 
 ## Configure the board
 
-Add your first component, the [board](/components/board/):
+Add your first component, the [board](/components/board/).
```
penguinland, line 39: Discussion question: technically, I think you can now set up a scuttlebot without having a board component! Should we include this step at all? Part of me says "no, the minimum working example does not need a board component." Part of me says "yes, we should have components for every piece of hardware, and we should show people how to use the board if they want to use it for something else (e.g., they want to put an ultrasonic sensor on their scuttlebot)." and then the first part of me says "so don't introduce the board component until it's needed in a different tutorial. It's unused here." What do you think?

penguinland (approved):
LGTM! All my feedback is minor or not really actionable.

docs/appendix/glossary/board.md
penguinland: :facepalm: I misread "signal" as "single" like 3 times. This sentence is fine as-is!

docs/tutorials/configure/scuttlebot.md
sguequierre: I think it makes sense to remove. removing!

penguinland (approved):

docs/components/motor/tmc5072.md
```diff
@@ -14,9 +14,6 @@ This model supports stepper motors controlled by the [TMC5072 chip](https://www.
 
 Whereas a basic low-level stepper driver supported by the [`gpiostepper` model](/components/motor/gpiostepper/) sends power to a stepper motor based on PWM signals from GPIO pins, the TMC5072 chip uses SPI bus to communicate with the board, does some processing on the chip itself, and provides convenient features including StallGuard2<sup>TM</sup>.
 
-To configure a `TMC5072` motor as a component of your robot, first configure the [board](/components/board/) to which the motor driver is wired.
```
npentrel, line 17: Is that no longer a requirement?

npentrel (approved):
I assume these changes mean boards are no longer required to be configured for encoded motors to work

docs/components/motor/tmc5072.md
penguinland: Right. The TMC motor component now talks directly to the SPI bus without needing to go through the board component.

penguinland:
> I assume these changes mean boards are no longer required to be configured for encoded motors to work

Yes, I believe that's the case!

PR #2322, opened 2023-12-13
DOCS-1517: Update rpi os setup

skyleilani:
Update images containing the Raspberry Pi Lite os to the Raspberry Pi Legacy Full os.


andf-viam (changes requested):
Thanks for tackling this! Two things:
- Is 32-bit correct? I thought we'd need the `Raspberry Pi OS Full (64-bit)` image to workaround the current wifi issue?
- We also need to update the text to match, such as step 2 [here](https://docs.viam.com/get-started/installation/prepare/rpi-setup/#install-raspberry-pi-os).

andf-viam (approved):
This LGTM % one additional spot. Thanks for this!

I found one more instance here: `/tutorials/custom/custom-base-dog.md` - I'd recommend just making that mention generic: `to install Raspberry Pi OS`.

skyleilani:
> This LGTM % one additional spot. Thanks for this!
> 
> I found one more instance here: `/tutorials/custom/custom-base-dog.md` - I'd recommend just making that mention generic: `to install Raspberry Pi OS`.

Thank you so much for the review!

PR #2324, opened 2023-12-15
Add note to explain modules not allowed on rovers

npentrel:
Placing this here because it feels odd with the try viam content that should be more introductory. Open to other ideas


andf-viam (approved):
This LGTM, but if you wanted to also mention it on a Try page for thoroughness, I'd suggest a `Limitations` or similar header just below [this one](https://docs.viam.com/get-started/try-viam/reserve-a-rover/#cancel-my-reservationsession).

PR #2334, opened 2023-12-18
DOCS-1511: Change JSON Template title in plant watering

JessamyT:
Changed two instances of "JSON Template" to "Raw JSON" since they're not so much templates as just exactly what to do. Also fixed some newlines.


mestcihazal (approved):
lgtm!

PR #2345, opened 2023-12-20
Add more info about vision service and camera to `mlmodel` page

sguequierre:
Matt Vella sent over feedback from a community user where he filmed a YouTube video and apparently spent a lot of time on this docs page not realizing he needed a vision service/that that was his "next steps". We do make it pretty clear already on that page but this is what I could think of to make it even more clear

* Add sentence under "Used with" talking about combination of mlmodel service, mlmodel vision service, and transform camera

### DIFF
--- a/docs/ml/deploy.md
+++ b/docs/ml/deploy.md
@@ -34,7 +34,8 @@ For some models, like the [Triton MLModel](https://github.com/viamrobotics/viam-
 {{< relatedcard link="/components/camera/">}}
 {{< /cards >}}
 
-After deploying your model, configure an [`mlmodel` vision service](/ml/vision/) and [`transform` camera](/components/camera/transform/) to visualize the predictions your model makes.
+After deploying your model, you need to configure an additional service to use the deployed model.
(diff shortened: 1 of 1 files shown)


npentrel (approved):

docs/ml/deploy.md
```diff
@@ -34,6 +34,8 @@ For some models, like the [Triton MLModel](https://github.com/viamrobotics/viam-
 {{< relatedcard link="/components/camera/">}}
 {{< /cards >}}
 
+After deploying your model, configure an [`mlmodel` vision service](/ml/vision/) and [`transform` camera](/components/camera/transform/) to visualize the predictions your model makes.
```
npentrel, line 37: So the one issue I have with this is that I believe the models are not just for vision services. That's the only thing we currently support natively but it doesn't necessarily have to be a vision service consuming this. 

And the transform camera is entirely optional.

```suggestion
After deploying your model, you need to configure an additional service to use the deployed model.
For example, you can configure an [`mlmodel` vision service](/ml/vision/) and a [`transform` camera](/components/camera/transform/) to visualize the predictions your model makes.
```

sguequierre: @npentrel Oh, I did not know that. Do you have an example of what else they're used in?

PR #2350, opened 2023-12-21
DOCS-1358: Add changelog filtering

JessamyT:


assets/js/changelog.js
```diff
@@ -0,0 +1,79 @@
...
+      lastclassname = nextel.className;
+      if (nextel.className.includes("show")) {
+        AddClass(headings[i], "show");
+        RemoveClass(headings[i], "hidethisone");
+        break;
+      } else {
+        RemoveClass(headings[i], "show");
+        AddClass(headings[i], "hidethisone");
+      }
+      if (nextel.nextElementSibling !== "undefined" && nextel.nextElementSibling !== "null") {
```
npentrel, line 59: You're currently checking that that element is unequal to the string `"undefined"` and the string `"null"` rather than the concept. 

You could instead do this, which tests that there is something that isn't false (undefined and null are both false):
```suggestion
      if (nextel.nextElementSibling) {
```

npentrel (changes requested):
Very nice! 
- the console error is because at some point you try to access `tagName` and it is `null`. You tried to ensure it's not null but aren't actually doing that :). See my comment for more details.
- The removed item is not working. There's a removed item but it's not showing up when I filter for removed things.

JessamyT:
"Removed" works when the April expander content is commented out; I un-commented it because Andrew's expander shortcode PR should fix that problem and render this PR good to go (theoretically)

npentrel (approved):
Awesome work!

PR #2363, opened 2024-01-04
Fix expander shortcode

andf-viam:
Fixes a few issues with the expander shortcode:
- Fixes indents and logic scope (`-`) to be renderable within HTML, and retain CSS styling.
	- This should allow this shortcode to be useable outside of pure MD, such as in HTML tables. I think!
- Fixes typo in default expander title to now display a default title on the expander if one is not otherwise supplied.
	- Previously, unset titles would result in blank expander titles.


JessamyT (approved):
LGTM though I'm no expert

npentrel (approved):
so it's a whitespace issue what's breaking for Jessamy's glossary changes? Or is this a separate issue?

andf-viam:
- Whitespace (indent) fixed non-functional on-click expander expanding behavior
- Usage of `{- * -}` notation fixed stylesheets not applying (rendering in monospace (raw), and only partially)

- Glossary fix was separate: include `scripts.html` also on `changelog.html` page, which seems to handle both tab and tooltip JS functionality.

PR #2364, opened 2024-01-04
DOCS-1531: Add tip reserved SQL keywords

andf-viam:
Add note to SQL Query page advising that reserved SQL keywords-as-fieldnames must be escaped.

QUESTION:

- Is it appropriate to link to Wikipedia's standard SQL reserved keywords list, as I have done? (i.e. is it the same list for standard SQL per wiki as for MDB-$SQL?) I was not able to find a MDB resource that talked about reserved or protected keywords. Thanks!


andf-viam:
@dmhilly Hi! I've added a small tip to our data query doc advising of the need to -- and how to -- escape reserved SQL keywords if needed. Let me know if you'd like to see any changes. Thanks!

docs/data/query.md
```diff
@@ -185,6 +185,16 @@ For example, to connect to your Viam organization's MongoDB Atlas instance and q
      [ { numStanding: 215 } ]
      ```
 
+If you use a data field that is named the same as a [reserved SQL keyword](https://en.wikipedia.org/wiki/List_of_SQL_reserved_words), such as `value` or `position`, you must escape that field name in your query using backticks ( <file>\`</file> ).
```
dmhilly, line 189: > Is it appropriate to link to Wikipedia's standard SQL reserved keywords list, as I have done? (i.e. is it the same list for standard SQL per wiki as for MDB-$SQL?) I was not able to find a MDB resource that talked about reserved or protected keywords. Thanks!

Very good question - so Atlas SQL is based on/mostly follows the SQL-92 standard, per the docs [here](https://www.mongodb.com/docs/atlas/data-federation/query/sql/language-reference/#compatability-and-limitations). We might want to document that somewhere and/or link to the Atlas SQL Dialect Language Reference. It doesn't have a list of reserved keywords in the docs, but [this is the section](https://www.mongodb.com/docs/atlas/data-federation/query/sql/language-reference/#identifiers) of their documentation that I found that talks about reserved keywords. I guess one could also use double quotes, which I hadn't realized and haven't tested.

It would probably be better to link to a list of SQL-92 reserved keywords, but I can't find a good link anywhere. I think linking to the Wikipedia page is okay since as I understand it, SQL:2023 keywords are just a superset of SQL-92, so the list will include any reserved keywords in Atlas SQL (+ some extras). Up to you though, as you are the documentation expert!

docs/data/query.md
```diff
@@ -185,6 +185,16 @@ For example, to connect to your Viam organization's MongoDB Atlas instance and q
...
+For example, to query against a data field named `value` in the `readings` collection, you would use:
+
+```mongodb {class="command-line" data-prompt=">"}
+select sensorData.`value` from readings
```
dmhilly, line 193: Nit, but I think we should have this say 
```sql
data.`value`
```
since there's no such field `sensorData` in the `readings` collection (`sensorData` is actually the name of the database containing the `readings` collection/table). But there is a field called `data`, that often has a subfield called `value` (depending on which components you've captured data from).

dmhilly (changes requested):
Thanks so much for your work on this Andrew - left a reply to your question about whether to link the SQL reserved keywords wiki. Curious to hear your thoughts!

docs/data/query.md
andf-viam: Will try to balance both options: retaining link to keywords (most helpful resource) with additional link to MDB Atlas data federation docs to further clarify: (SQL-92 + some other possibly-salient limitations). LMK if you agree!

dmhilly: I ran into the same issue when I did a quick search - so I agree! Worst case scenario is the customer uses backticks to delimit the identifier when the backticks are not necessary, but the query would still work as expected.

dmhilly (approved):
LGTM! 🎉

PR #2365, opened 2024-01-04
DOCS-1507: Add API keys details in RBAC

sguequierre:
* Update outdated screenshot
* Add instructions around using the API keys table which is in the same place as the org members/user access functionality the rest of this page covers


npentrel (approved):
LGTM - thank you!

PR #2366, opened 2024-01-04
Add link to micro-rdk from Components

JessamyT:


andf-viam (approved):
This LGTM! I like the explicit 64-bit mention, since 32-bit support is the hallmark of the micro-RDK, and should further help the user be sure they're on the right page! Nice work.

JessamyT:
Thanks! I didn't want 64-bit users accidentally running away from this page if they should indeed be here!

PR #2372, opened 2024-01-08
DOCS-1572: Add note to `GetImages()` method explaining imagers usage

sguequierre:
Clyde confirmed method is supposed to work with other cameras but was created for realsense and similar cameras. 

* Add note to explain that


bazile-clyde (approved):
LGTM

npentrel (approved):

PR #2412, opened 2024-01-22
DOCS-1667/1713: Update docs for security tab removal and update tabs on RBAC page R2D2

sguequierre:
**Remember to do a final sweep after merging in new changes to check for missed spots in between**


andf-viam (approved):
This LGTM, thanks!

PR #2416, opened 2024-01-23
Update create-subtype.md

andf-viam:
Fix three links to moved files, now hosted here: https://github.com/viam-labs/speech-service-api


andf-viam:
@HipsterBrown just a sanity check request that I got updated link targets right. Thanks so much!

HipsterBrown (approved):
Thanks for the quick update 🙌

PR #2417, opened 2024-01-23
Update verification system with higher quality video

sguequierre:
High-quality source is a .mov. 
I'm not sure if I made the conversion with the recommended commands or with a website last time, so trying with the recommended commands and with the .webm one adjusted for quality. 

Played through and it seems fine to read the words now?


static/tutorials/verification-system/demo.mp4
npentrel, line 1: it's odd that this became smaller isn't it

npentrel (approved):

PR #2418, opened 2024-01-23
Update preview to use video

npentrel:
What do you think about using the video for preview?


sguequierre (approved):
It doesn't look like change reflects in preview so idk for sure what it looks like but I'm good with this! Thanks Naomi!

PR #2422, opened 2024-01-23
Add dashboards for searchability

npentrel:


andf-viam (approved):
Ah gotcha, works for me!

PR #2427, opened 2024-01-24
Remove weight and fix date II

andf-viam:
Unless you want weight on this one!


docs/tutorials/services/data-mlmodel-tutorial.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: "Capture Data and Train a Model"
 linkTitle: "Capture Data and Train a Model"
-weight: 4
```
npentrel, line 4: with this one we do want to keep it because it's one of the 4 we do want at the top. We just should probably discuss if we add or remove one of them

npentrel (approved):

PR #2436, opened 2024-01-26
DOCS-1528: Add encoded motor videos

npentrel:


sguequierre:
I think this looks good. The quality could be better but it improves when the rover is moving?

sguequierre (approved):

npentrel:
Updated the quality - managed to use the mp4 file you had originally created and just shortened the control-view to be the same length which works

PR #2438, opened 2024-01-26
RSDK-6277 Update customlinux.md to remove board packages

randhid:
# Description

We are removing board packages from the package registry. This and #2439  are where I could find mentions of this command.


npentrel:
How does this work without the command. Can you still use the model?

npentrel (approved):

PR #2439, opened 2024-01-26
RSDK-6277 Update cli docs to remove board package

randhid:
# Description

We are removing board packages from the registry. This and #2438 are were I could find mentions of this command


npentrel:
It's also used here - how does the customlinux model work now that that's removed? https://docs.viam.com/components/board/customlinux/#upload-your-board-definitions-file

npentrel (approved):

PR #2448, opened 2024-01-30
Update Titles.yml

sguequierre:
@npentrel Is it ok to keep this `warning` level and not `error` like on local? There's lots of cases where it flags nouns being capitalized and creates errors-- thinking these would cause checks to fail--


npentrel:
sounds good - I think then we can close this PR

PR #2457, opened 2024-02-02
DOCS-1773: Add info about preview images to contributing page

JessamyT:


docs/appendix/contributing.md
```diff
@@ -545,6 +545,13 @@ The `2gif` commands only turn the first 5 seconds of a video into a low res gif.
 
 {{% /expand%}}
 
+{{% alert title="Note" color="note" %}}
+
+We _do_ include <file>.gif</file> file types in front matter of some docs (such as tutorials) when we want to display a GIF as the link preview image on external sites like social media.
+In the front matter we include an MP4, WebM and GIF of the same video because the GIF is compatible with external sites while the lower-bandwidth formats are displayed on our docs site.
```
npentrel, line 551:
```suggestion
If you use `webmSrc` and `mp4Src` to add a card video for a page (such as to a tutorial), you should also add a <file>.gif</file> version of the video to the front matter.
Hugo uses the GIF as a link preview image that gets displayed on external sites when someone shares the link, for example on Slack or Twitter.

Why?

- GIFs take more bandwidth than MP4 and WEBM videos, so on our page, we prefer to use MP4 and WEBM.
- Link previews using `og:image` only support GIFs
- Automatic conversion between the formats is possible but we also need to make sure that the resulting GIFs are presentable and under 1MB.
```

So I did just look into this a bit further to write this AND there is an `og:video` type which we may be able to use. We will need to test if slack etc support it.

docs/tutorials/configure/pet-photographer.md
```diff
@@ -4,15 +4,11 @@ linkTitle: "Pet Photographer"
 type: "docs"
 description: "Use the filter modular component in the Viam app to photograph your pet in their collar."
 tags: ["vision", "filter", "camera", "detector", "services"]
-image: "/tutorials/pet-photographer/data-capture.png"
```
npentrel, line 7: This does unfortunately break previews. You can confirm here: https://docs-test.viam.dev/2457/components/base/ at the bottom. 

So when I said I wasn't sure whether you could remove one of them and the other would automatically use the same field I thought if you removed `images` it might automatically take the images available in `image`. It doesn't. Read [this here](https://gohugo.io/templates/internal/#configure-open-graph), it says "The first 6 URLs from the images array are used for image metadata. If [page bundles](https://gohugo.io/content-management/page-bundles/) are used and the images array is empty or undefined, images with file names matching *feature* or *cover*,*thumbnail* are used for image metadata.". 

What we can do is change our card partial and shortcode to use the 1st image in the `images` array if it is not a gif. 

However the other interesting thing that this bit of their docs points to is that there is a videos array that's natively part of the front matter.  We could probably use that instead of webmSrc and mp4Src and use both based on the file type.

Those are somewhat complex changes because our `list.typesense.json` file (which populates our tutorials database) currently assumes there is an `image` or a `webmSrc` and `mp4Src` variable in the front matter.

I'd suggest doing that in 3 parts:
Changing everything for images and then changing everything for videos and if it works, removing gifs as a third step. Would you like to do that work? Absolutely ok if that goes beyond what you want to do here but I want to give you the option :).

layouts/partials/card.html
```diff
@@ -19,34 +19,32 @@
...
-				</div>
-				{{ end }}
-        <div {{ if .Params.image }}class="small-hover-card-div"{{ end }}{{ if .Params.webmSrc }}class="small-hover-card-div"{{ end }}>
-          {{ if $customTitle }}
-            <div>{{ $customTitle }}</div>
+      {{ else if .Params.images }}
+        <div class="hover-card-img">
+          {{ $img := resources.GetMatch (index .Params.images 0) }}
+          {{ if $img }}
+            {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "400x" "alt" (.Params.imageAlt) "style" "") }}
```
npentrel, line 26: I believe this is all that needs to change. The problem is with pages where there is an images field with an image in it but no alt text. Alternatively you could add an alt text for each such page.

```suggestion
            {{ if .Params.imageAlt }}
              {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "400x" "alt" (.Params.imageAlt) "style" "") }}
            {{ else }}
              {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "400x" "alt" (.Params.linkTitle) "style" "") }}
              {{ warnf "using linktitle for %s" (index .Params.images 0)}}
            {{ end }}
            {{ else }}
```

npentrel (commented):
There is something else in `list.typesense.json` that needs to change. When you build the page this generates a json called `typesense.json` and now the image fields are blank. They'll also need to take the first specified image in images

Update:
I've made the changes to the typesense file. It wasn't just that it needed to use the first item in the array - it also needed to check if it's an image. A fairly small change. I've also taken the opportunity to update that file so that it doesn't generate a lot of white space around it.

layouts/partials/card.html
```diff
@@ -46,14 +47,26 @@
           {{ if $customTitle }}
             <div>{{ $customTitle }}</div>
           {{ else }}
-            <div>{{ .LinkTitle }}</div>
-          {{ end }}
-          {{ if ne $class "small" }}
-          <p> {{ if $customDescription }} {{ $customDescription }} {{ else }} {{ .Description }} {{ end }} </p>
+            {{ errorf "Couldn't get images %s" .Params.images }}
```
npentrel, line 50: this doesn't make sense here. Something went wrong with the merge and because it's a merge commit, it has changed the history in ways I'm not super keen to untangle.

layouts/partials/card.html
```diff
@@ -46,14 +47,26 @@
           {{ if $customTitle }}
             <div>{{ $customTitle }}</div>
           {{ else }}
-            <div>{{ .LinkTitle }}</div>
-          {{ end }}
-          {{ if ne $class "small" }}
```
npentrel, line 51: also no idea why this is beign deleted

layouts/partials/card.html
npentrel: I recommend you do a hard reset to before the merge conflict and then pull rebase and have a closer look at what changed and how the changes should now fit together.

JessamyT: Yeah, yesterday I was trying to merge and resolve conflicts but there were mega conflicts with the recent change to this file so was gonna work more on fixing today

npentrel: I did manage to come up with something that merges the logic when trying to rebase, If you do want to copy that from [here](https://github.com/npentrel/viam-docs/blob/0b5e49a0f088d0d7ffaea2be2484f52a96574837/layouts/partials/card.html) you can. I believe somehow some stuff gets duplicated even during the rebase. A bit odd.

docs/appendix/contributing.md
JessamyT: Does this comment still apply if we aren't using og:video?

JessamyT:
@npentrel actually disregard request for re-review; now that we're not catching the singular "image," I need to update a bunch of frontmatter, more strongly say in the template (and contributing) that we don't want "image" at all, and figure out why the alt text still isn't being pulled in. Will request after I work on this more tomorrow!

layouts/partials/tutorialcard-no-js.html
```diff
@@ -18,17 +18,17 @@
             </video>
           </div>
           {{ end }}
```
npentrel, line 20:
```suggestion
```

layouts/partials/tutorialcard-no-js.html
```diff
@@ -18,17 +18,17 @@
...
-          {{ if .Params.image }}
+          {{ if .Params.images }}
           <div class="hover-card-img">
-            {{ partial "imgproc.html" (dict "src" .Params.image "resize" "300x" "alt" .Params.imageAlt "style" "" "class" "" "declaredimensions" false ) }}
+            {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "300x" "alt" .Params.imageAlt "style" "" "class" "" "declaredimensions" false ) }}
           </div>
           {{ end }}
           {{ if .Params.icon }}
```
npentrel, line 26:
```suggestion
          {{ else if .Params.icon }}
```

layouts/partials/tutorialcard-no-js.html
```diff
@@ -18,17 +18,17 @@
             </video>
           </div>
           {{ end }}
-          {{ if .Params.image }}
+          {{ if .Params.images }}
```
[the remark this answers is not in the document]
npentrel: this will only check for images if there's not already a video there. That's essentially all you were missing I think

layouts/partials/card.html
```diff
@@ -9,52 +9,54 @@
...
-    {{if $customCanonicalLink }}
+{{ if site.GetPage ($link | string) }}
+  {{ with site.GetPage ($link | string) }}
+    {{ if $customCanonicalLink }}
     <a href="{{$customCanonicalLink}}">
     {{ else }}
     <a href="{{.RelPermalink}}">
     {{ end }}
+    {{ if $imageOverwrite }}
+        <div class="hover-card-icon">
```
npentrel, line 20:
```suggestion
        <div class="hover-card-img">
```

layouts/shortcodes/manualcard.html
```diff
@@ -7,7 +7,7 @@
           {{if $img }}
             {{ partial "imgproc.html" (dict "src" (.Get "img") "resize" "400x" "alt" (.Params.imageAlt) "style" "") }}
           {{ else }}
-            {{ errorf "Couldn't get image %s" .Params.image }}
+            {{ errorf "Couldn't get image %s" .Params.images }}
```
npentrel, line 10: I think this should be .Get img

layouts/shortcodes/tutorialcard.html
```diff
@@ -14,17 +14,17 @@
...
           </div>
           {{ end }}
-          {{ if .Params.image }}
+          {{ if .Params.images }}
           <div class="hover-card-img">
-            {{ partial "imgproc.html" (dict "src" .Params.image "resize" "300x" "alt" .Params.imageAlt "style" "" "class" "" "declaredimensions" false ) }}
+            {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "300x" "alt" .Params.imageAlt "style" "" "class" "" "declaredimensions" false ) }}
           </div>
           {{ end }}
           {{ if .Params.icon }}
```
npentrel, line 22: You need to do the same as for the no-js one here

npentrel (commented):
A few tweaks

layouts/shortcodes/manualcard.html
JessamyT:
```suggestion
            {{ errorf "Couldn't get image %s" .Get img }}
```
Like this? Why?

layouts/shortcodes/tutorialcard.html
JessamyT, line 22:
```suggestion
          {{ else if .Params.icon }}
```

layouts/shortcodes/tutorialcard.html
npentrel, line 1: In fact if you wanted to you could also just delete this file, we no longer use this shortcode

npentrel (approved):

PR #2464, opened 2024-02-02
DOCS-1668: Add iterative dev steps

andf-viam:
Add iterative development steps for locally testing and refining a module:
- Local option, for same platform/same arch
- Remote option, based around Mutagen, for different platform/different arch.

Included troubleshooting section, mostly for troubleshooting common `ssh` setup issues
Added some minor, out-of-scope additional links to examples on Create page. Can ignore in review.

Notes:
- Originally considered this for appearance immediately preceding [this header](https://docs.viam.com/registry/create/#additional-example-modules) on Create, but deemed too long to include at location (too long to wait for Deploy steps if not using Iterative, for example), so moved to own dedicated page under Advanced. Open to other ideas!
- There should be no need to `kill -9` the module proc, as `viam-server` should handle child proc management when it itself is restarted. James should also rather prefer we did not recommend `-9` to end users as a viable approach. Was the intent here to restart module proc without restarting the `viam-server` proc itself? I could see merit there (slight time saving over bring down, then up, `viam-server` and all other modules it manages), but not at the cost of using `-9` in our docs as any form of recommendation. For most users (who only develop one module at a time, and have only one module added to their development machine), a full `viam-server` restart should be quick enough.
- `ssh` is tricky for many reasons. Opted for verbosity here, especially as context of upcoming utility for this piece is in the classroom.
- Did not mention Docker, but Mutagen approach should also work for Docker containers with minimal changes. Useful for same arch, different OS. Still not an option for different arch. Chose not to mention (not an explicit ask, already lengthy read), but could add back in if desired.


docs/registry/advanced/iterative-development.md
```diff
@@ -0,0 +1,246 @@
+---
+title: "Iterative Module Development"
+linkTitle: "Iterative Development"
+weight: 40
+type: "docs"
+tags: ["modules", "modular resources", "registry"]
+description: "Iterate on and test your modules locally before uploading to the Viam registry."
+icon: "/services/icons/ml.svg"
```
andf-viam, line 8: Just for now, BD has new icons on the way soon!

docs/registry/advanced/iterative-development.md
```diff
@@ -0,0 +1,246 @@
...
+   1. If you haven't already, ensure that your target test system is up and running, and accessible to `ssh`.
+      Mutagen uses `ssh` to sync files between your systems in an unattended manner (that is, without user input), and so requires that you set up a valid `ssh` key for it to use to authenticate during syncs.
+      If you already have an `ssh` keypair configured to your remote testing system, you can skip this step and proceed to step 3.
+      Follow the steps below to create a new `ssh` key to use, and copy it to your remote testing system:
+
+      1. Create a new `ssh` key on your development workstation, meeting your security requirements.
+         For example, on a macOS machine, you can generate an `ssh` key using the `ed25519` algorithm with a key size of `4096` bits using the following command:
+
+         ```sh { class="command-line"}
+         ssh-keygen -t ed25519 -b 4096 -C "developer@example.com"
```
andf-viam, line 110: Current, generic, non-specific, catch-all recommendation for general use.

andf-viam:
@mcvella Hi Matt -- here is the docs walkthrough of your iterative module development steps. Let me know if you'd like to see any changes! Left some comments in the PR description addressing some of your comments in your walkthrough as well, if you have feedback. Thanks much!

npentrel (approved):
Looks great to me from the docs side - excellent work

andf-viam:
Thanks @mcvella 

I didn't run into any troubleshooting items that might expand the current selection outside of context:ssh in my testing. Happy to add yours if you have some good ones! I've updated the section header to focus just on `ssh`. We generally include troubleshooting content specific to a page on that page, where a user is more likely to see it. At the bottom, however, to not interrupt against those who don't need it. So I've retained it here.

We've talked offline about the `kill -9` recommendation. I think it'd be appropriate for you to bring up in your classes / workshops as a tip/trick if you so choose. But I don't believe it can be in our public-facing docs as a recommendation, as we've discussed. 

This LGTY to ship? I'd bet you get some feedback on it when you use it in-class/in-workshop: feel free to reach back out with any classroom-sourced refinements; happy to enact. Thanks much!

PR #2472, opened 2024-02-05
DOCS-1726: Update references to config tab subtabs

sguequierre:
This isn't quite everything, ticketed everything that came up that I couldn't do yet that wasn't already ticketed

* Change save config instructions
* Update references to config tab subtabs for new workflows

### DIFF
--- a/docs/build/micro-rdk/board/esp32.md
+++ b/docs/build/micro-rdk/board/esp32.md
@@ -86,7 +86,7 @@ Copy the following JSON template and paste it into your configuration inside the
 {{< /tabs >}}
 
 Edit and fill in the attributes as applicable.
-Click **Save config**.
+Press **Command+S** to save your config.
 
 The following attributes are available for `esp32` boards:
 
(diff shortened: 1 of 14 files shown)


docs/build/micro-rdk/board/esp32.md
```diff
@@ -86,7 +86,7 @@ Copy the following JSON template and paste it into your configuration inside the
 {{< /tabs >}}
 
 Edit and fill in the attributes as applicable.
-Click **Save config**.
+Press **Command+S** to save your config.
```
npentrel, line 89: Is there no save button anymore? Not everyone is on a mac so if we need to do this we need to also offer the windows/linux equivalent...

npentrel (approved):

docs/build/micro-rdk/board/esp32.md
sguequierre: Oh that is a very good point. There is a save button. I'll change the command to that

PR #2474, opened 2024-02-05
Change changelog for rss feed

npentrel:


layouts/docs/changelog.html
```diff
@@ -3,32 +3,70 @@
...
-                  <button class="filter-button" onclick="filterSelection('changelog-improved')" id="improved-button">Improvements</button>
-                  <button class="filter-button" onclick="filterSelection('changelog-added')" id="added-button">Additions</button>
-                  <button class="filter-button" onclick="filterSelection('changelog-changed')" id="changed-button">Changes</button>
-                  <button class="filter-button" onclick="filterSelection('changelog-removed')" id="removed-button">Removals</button>
-                  <button class="filter-button active" onclick="filterSelection('all')" id="all-button">Show All</button>
+                  <button
+                    class="filter-button"
+                    onclick="filterSelection('changelog-improved')"
+                    id="improved-button"
+                  >
```
JessamyT, line 39: Why are these all stretched out? Is this how it should be formatted?

layouts/docs/changelog.html
```diff
@@ -3,32 +3,70 @@
...
+                    Additions
+                  </button>
+                  <button
+                    class="filter-button"
+                    onclick="filterSelection('changelog-changed')"
+                    id="changed-button"
+                  >
+                    Changes
+                  </button>
```
JessamyT, line 55:
```suggestion
                  >Changes</button>
```
[opt/just wondering] Wouldn't at least this much condensation actually make it more readable and consistent with other tags?

layouts/docs/tutorials.html
JessamyT, line 1: Is this related to changelog?

layouts/shortcodes/changelog.xml
```diff
@@ -0,0 +1,7 @@
+<item>
+  <title>{{ .Get "title" }}</title>
+  <link>https://docs.viam.com/appendix/#{{ .Get "title" | anchorize }}</link>
+  <pubDate>{{ .Get "date" | time.Format "Mon, 02 Jan 2006 15:04:05 -0700" }}</pubDate>
```
JessamyT, line 4: Not sure how the date formatting works but, this is different from the format passed into the shortcodes in changelog.md...how does this work? And do we really need to have a date instead of just a month, when the date is not actually the first of the month?

docs/appendix/changelog.md
```diff
@@ -21,52 +22,48 @@ Check out [this tutorial](/tutorials/projects/filtered-camera/) for more informa
 
 {{% /changelog %}}
 
-{{% changelog color="added" title="Raspberry Pi 5 Support" %}}
+{{% changelog date="2023-12-01" color="added" title="Raspberry Pi 5 Support" %}}
```
JessamyT, line 25: If possible I think we shouldn't put the first of the month since that's not accurate

layouts/docs/changelog.html
npentrel: I think my editor possibly automatically formatted this

layouts/shortcodes/changelog.xml
npentrel: This is the format RSS expects. The commands change the formatting to match

npentrel: I think we do need to add a full date but I'll test tomorrow

JessamyT: Okay. I don't love the idea that someone could get the RSS feed and think that they should be looking for a viam-server version that came out the first of the month

JessamyT: If it's not possible to eliminate the full date day, do you think it'd be worth it to (in a separate ticket) fix these with more accurate dates? We used to have most of them in here anyway I think, so could just re-apply them

docs/appendix/changelog.md
npentrel: changed to the end of the month. I don't think any dates except the vision service are accurate. But we did used to announce everything at the end of the month so may as well stick with that

layouts/docs/changelog.html
npentrel: changed back

docs/appendix/changelog.md
```diff
@@ -692,24 +679,22 @@ You will need to first register the machine learning model file with the [ML mod
 
 {{% /changelog %}}
 
-## March 2023
-
-{{% changelog color="added" title="Machine learning for image classification models" %}}
+{{% changelog date="2023-03-31" color="added" title="Machine learning for image classification models" %}}
 
 You can now [train](/ml/train-model/) and [deploy](/ml/deploy/#create-an-ml-model-service) image classification models with the [data management service](/data/) and use your machine's image data directly within Viaml
```
JessamyT, line 684:
```suggestion
You can now [train](/ml/train-model/) and [deploy](/ml/deploy/#create-an-ml-model-service) image classification models with the [data management service](/data/) and use your machine's image data directly within Viam.
```

JessamyT (approved):
LGTM! Noticed an out-of-diff nit if you care to fix

PR #2479, opened 2024-02-06
DOCS-1772: Update upload-model.md label files not required

sguequierre:
Update wording for changes making label files not required but suggested in App upload workflow.

Q for reviewers: does anybody prefer "(Suggestion)" or "(Recommended)" to "(Optional)"?

### DIFF
--- a/docs/ml/upload-model.md
+++ b/docs/ml/upload-model.md
@@ -14,6 +14,7 @@ description: "Upload a Machine Learning model to Viam to use it with the ML Mode
 The ML model service works with models trained inside and outside the Viam app.
 To use a model that you have trained yourself outside the Viam app, [upload it as a new model privately or share it in the Viam registry](#upload-a-new-model-or-new-version).
 If you need to update a previously uploaded model, you can also [upload a new version](#upload-a-new-model-or-new-version).
+To work with the `tflite_cpu` ML model service, an ML model is comprised of a <file>.tflite</file> model file which defines the model, and optionally a <file>.txt</file> labels file which provides the text labels for your model.
 
 If you have [trained](/ml/train-model/) or uploaded an ML model privately and now want to make it available for reuse, you can at any point [make the existing model public in the registry](#make-an-existing-model-public-in-the-registry).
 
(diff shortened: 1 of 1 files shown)


andf-viam (commented):
I think `(Optional)` works well, but I feel that we need a mention of why you might want to use labels, so the user can decide how optional this optional step is to their use case.

What do you think of including an additional sentence somewhere between lines 15 and 18? I think we need to not only to mention that it is optional here, but that it is part of this flow -- that it exists -- at all. Whatever you think fits in here,  but maybe something similar to:

> An ML model is comprised of a `.tflite` model file which defines the model, and optionally a `.txt` labels file which provides the text labels for your model.

docs/ml/upload-model.md
```diff
@@ -43,7 +43,7 @@ Select **Next steps** to continue.
 1. Specify a name for the model.
 2. Specify the **Model type**.
 3. Add a `.tflite` model file.
-4. Add a `.txt` label file.
+4. (Optional) Add a `.txt` label file.
```
npentrel, line 46: just talking about this, if we can do some research on what label files should look like or link out to something that woudl be good

andf-viam: Sure, here's a [common audio ML model](https://www.kaggle.com/models/google/yamnet/frameworks/tfLite/variations/classification-tflite/versions/1?tfhub-redirect=true), the download zip includes a good example labels.txt file. Looks like this:

```
bananajr:Downloads andf$ head -5 labels.txt
Person
Bicycle
Car
Motorcycle
Airplane
```

sguequierre: This will usually be generated for them I think.

docs/ml/upload-model.md
```diff
@@ -14,6 +14,7 @@ description: "Upload a Machine Learning model to Viam to use it with the ML Mode
 The ML model service works with models trained inside and outside the Viam app.
 To use a model that you have trained yourself outside the Viam app, [upload it as a new model privately or share it in the Viam registry](#upload-a-new-model-or-new-version).
 If you need to update a previously uploaded model, you can also [upload a new version](#upload-a-new-model-or-new-version).
+To work with the `tflite_cpu` ML model service, an ML model is comprised of a <file>.tflite</file> model file which defines the model, and optionally a <file>.txt</file> labels file which provides the text labels for your model.
```
sguequierre, line 17: Added first part of sentence because hypothetically other models don't have to use tensorflow right?

andf-viam: I believe you are correct! The `tflite_cpu` model indeed requires TF & `.tflite` files specifically, but other models, perhaps delivered as a module, could support their own.

This sentence LGTM!

PR #2491, opened 2024-02-08
DOCS-1491: Add info on creating modules for ESP32

JessamyT:


docs/build/micro-rdk/generic/_index.md
```diff
@@ -13,6 +13,14 @@ no_list: true
 The _generic_ component {{< glossary_tooltip term_id="subtype" text="subtype" >}} enables you to add support for unique types of hardware that do not already have an [appropriate API](/build/program/apis/#component-apis) defined for them.
 For more information and models supported by the {{< glossary_tooltip term_id="rdk" text="RDK" >}}, see [Generic Component](/components/generic/).
 
+If your micro-RDK machine includes a resource that isn't a [base](/build/micro-rdk/base/), [board](/build/micro-rdk/board/),[encoder](/build/micro-rdk/encoder/), [movement sensor](/build/micro-rdk/movement-sensor/), [motor](/build/micro-rdk/motor/), or [servo](/build/micro-rdk/servo/), you can create a {{< glossary_tooltip term_id="modular-resource" text="modular resource" >}} to add support for it as a custom model of the generic subtype.
+
+{{< alert title="Important" color="note" >}}
+The micro-RDK works very differently from the RDK, so creating modular resources for it is different.
```
npentrel, line 19: modifier doesn't add anything
```suggestion
The micro-RDK works differently from the RDK, so creating modular resources for it is different.
```

docs/build/micro-rdk/generic/_index.md
```diff
@@ -13,6 +13,14 @@ no_list: true
...
+Refer to the [Micro-RDK Module Template on GitHub](https://github.com/viamrobotics/micro-rdk/tree/main/templates/module) for information on how to create custom resources for your micro-RDK machine.
```
npentrel, line 20: That template and walkthrough is what we should pull into the docs - it's not as findable on github and one link to it isn't really enough. For the modules their docs living in the github readme works because we link to their readmes everywhere but that is not true in this case.

We also need to include at least a reference on the create page of the registry, possibly under the advanced section if it doesn't fit with the create page

npentrel (changes requested):

docs/build/micro-rdk/generic/_index.md
JessamyT: Added note to create

npentrel (approved):

PR #2494, opened 2024-02-08
Update the docs front page

npentrel:


sguequierre (approved):
I do feel like this hides boards carousel and Viam in 3 minutes, but guessing that's intentional?

npentrel:
Yeah - it is intentional. I think we can consider swapping try viam and viam in 3 minutes possibly once we've revised viam in 3 mintues. 
For the boards carousel - what would you think about that being at the bottom of the page? Or on a more dedicated get-started page? I think I'll create a ticket for us to discuss it further. For now let's move forward with this and we can readd as needed

PR #2502, opened 2024-02-09
Fix HTML, description for generic service

andf-viam:
Fix generic service PR post-merge:
- Add entry to landing page services list
- Fix HTML for services list on build index page
- Lengthen `fake` generic service description to exceed minimum word count.


JessamyT (approved):
LGTM but TBH not 100% sure I see all the diffs in the big long html lines so I trust that you did the stuff that works :D

PR #2505, opened 2024-02-10
DOCS-1601: Make processes page and clean up run.md

JessamyT:
- Made processes its own page
- As I dug into references to processes, it turned out the run code page needed changing so did that alongside. For example, removed section that doesn't actually apply to processes.
- Regarding where the processes page belongs: I considered putting this under Program, but processes are not necessarily tied to SDK code (you can definitely run non-SDK code processes), so I kept it in Configure


docs/appendix/glossary/process.md
```diff
@@ -2,14 +2,14 @@
 title: Process
 id: process
 full_link:
-short_description: Processes are binaries or scripts that run on a part.
+short_description: Processes are commands such as binaries or scripts that run on a part.
```
npentrel, line 5:
```suggestion
short_description: You can configure instances of programs or scripts to run on your machine part.
```

docs/appendix/glossary/process.md
```diff
@@ -2,14 +2,14 @@
...
 aka:
 ---
 
-Processes are binaries or scripts that run on a {{< glossary_tooltip term_id="part" text="part" >}}.
+Processes are binaries, scripts, or other commands that run on a {{< glossary_tooltip term_id="part" text="part" >}}.
```
npentrel, line 9: As is this actually isn't correct
```suggestion
Processes are instances of binaries, scripts, or other commands that run on a {{< glossary_tooltip term_id="part" text="part" >}}.
```

docs/build/configure/_index.md
```diff
@@ -225,44 +225,10 @@ Find more information in our [remotes documentation](/build/configure/parts-and-
 
 ## Processes
 
-To automatically run a specified script when the machine boots, configure a _process_.
+To automatically run a specified command when the machine boots, configure a _{{< glossary_tooltip term_id="process" text="process" >}}_.
+A process can be any command, for example one that executes a binary or a script.
```
npentrel, line 229:
```suggestion
You can configure any command, for example one that executes a binary or a script, to run as a process.
```

docs/build/configure/_index.md
```diff
@@ -225,44 +225,10 @@ Find more information in our [remotes documentation](/build/configure/parts-and-
...
-      "args": [
-        "my_cool_script.py"
-      ],
-      "cwd": "/home/myName/project/"
-    }
-  ]
-```
-
-{{% /expand %}}
+Find more information in our [processes documentation](/build/configure/processes/).
```
npentrel, line 231: avoid first person pronouns

assets/build/program/on-robot.png
npentrel, line 1: why delete this?

docs/build/program/run.md
```diff
@@ -77,62 +77,53 @@ If you subsequently lose internet connectivity, but stay connected to LAN or WAN
 
 ## Run code on-machine
 
-If you run [PID control loops](https://en.wikipedia.org/wiki/PID_controller) or your machines have intermittent or no network connectivity, you can ensure this does not interfere with the code's execution by running the control code on the same board that is running `viam-server`.
+You can run SDK code directly on your machine.
+If you run [PID control loops](https://en.wikipedia.org/wiki/PID_controller) or your machines have intermittent or no network connectivity, you can ensure lags in communication do not interfere with the machine's performance by running the control code on the same board that is running `viam-server`.
+Running everything on one machine is also convenient if you have a machine (for example, an air quality sensor) that runs all the time, and you don't want to have to connect to it from a separate computer constantly.
 
-{{<imgproc src="/build/program/on-robot.png" resize="800x" declaredimensions=true alt="A client running on a machine">}}
```
npentrel, line 82: why remove this?

docs/build/program/run.md
```diff
@@ -77,62 +77,53 @@ If you subsequently lose internet connectivity, but stay connected to LAN or WAN
...
-
-If you want to run your code on-machine automatically when your machine boots, you can configure Viam to run your code as a [process](/build/configure/#processes).
-
-To be able to run your code from your board, you need to install the relevant SDK as well as other required dependencies:
+Install the appropriate programming language and [Viam SDK](/sdks/) on your machine and run a command to execute the program in the terminal of that machine instead of from a separate computer:
 
 {{< tabs >}}
 {{% tab name="Python" %}}
 
-1. [`ssh` into your board](/get-started/installation/prepare/rpi-setup/#connect-with-ssh) and install `pip`:
```
npentrel, line 95: We do need to move this to the process page. The whole installing into a folder thing is not intuitive. I'd consider putting that into an include to include it in both pages maybe even I think it's helpful

npentrel (changes requested):
Good start I like the new page. I think we need to keep the python instructions somewhere, and possibly best to put them in both places - but open to ideas

assets/build/program/on-robot.png
JessamyT: Because it's exactly the same as the other image; I think it was named "on-robot.png" erroneously since it doesn't actually show code running on-robot.

docs/build/program/run.md
JessamyT: See above

docs/build/program/run.md
JessamyT: It doesn't need to be installed into a given folder though. Virtual environments may affect this but I've never installed it into the same folder and it has worked!

docs/appendix/glossary/process.md
JessamyT: Hmm but an instance of a command (which isn't a binary or a script) isn't really how you'd word it either is it? Not sure how to word

docs/build/configure/_index.md
JessamyT:
```suggestion
Find more information in the [processes documentation](/build/configure/processes/).
```

docs/build/program/run.md
npentrel: The arrow and the box surrounding the client is different. It's subtle but I think it would be better to make that less subtle than to delete

docs/appendix/glossary/process.md
npentrel: It's generally and instance of a program. 
```suggestion
Processes are instances of programs, invoked by commands running binaries or scripts on a {{< glossary_tooltip term_id="part" text="part" >}}.
```

docs/build/program/run.md
npentrel: I found where we discussed this: https://viaminc.slack.com/archives/C02SYTDQ56X/p1683924608761639

docs/appendix/glossary/process.md
JessamyT: That is how we've tended to think about them but, since that's not technically the whole truth it seems the definition here should say what all they can be. But I guess since the edits to the Processes section and page say they can be anything it's okay?

docs/build/program/run.md
JessamyT: Woowwww that is so subtle haha. I totally played "spot the difference" before I first took this out and didn't notice--totally interpreted the big grey box as just a background. Okay will revert this and made a BD ticket and docs ticket to make less subtle.

docs/build/program/run.md
JessamyT: Ahhh I'm sorry it turns out the process I was running doesn't actually use Viam Python SDK methods--it's a non-Viam Python script. Will amend, and I will also add a note to make it clear that these steps are only necessary if you're running a **Viam SDK** script **as a process**, not just if you're running any Python script as a process.

docs/build/program/run.md
JessamyT: ^^Done

docs/build/program/run.md
JessamyT: Done

JessamyT: I added it to the processes page but I don't think it needs to be here too since this page no longer covers any instructions for setting up a process; folks need to follow the link anyway.

PR #2514, opened 2024-02-12
Update module build setup docs to have `./setup.sh`

zaporter-work:
# Description

This comes from this PR: https://github.com/viamrobotics/rdk/pull/3572

This came from a misunderstanding of the scope doc and what the setup/build fields should accept.

We are restricting it to only accept valid bash commands (and the name of a shell script without a ./ is not a valid bash shell command)


npentrel (approved):

zaporter-work:
@npentrel The htmltest keeps failing. Is that due to my changes or is it just flaky?

PR #2519, opened 2024-02-13
Ignore chrome.google.com

npentrel:
# Description


.htmltest.yml
```diff
@@ -28,6 +28,7 @@ IgnoreURLs:
   - "electronics-tutorials.ws"
   - "marketplace.visualstudio.com"
   - "chromewebstore.google.com"
```
npentrel, line 30:
```suggestion
```

PR #2524, opened 2024-02-14
Update gps-nmea.md add GetAccuracy as supported method

sguequierre:
Figured this makes sense as it's supported by all GPS models?


andf-viam (approved):
LGTM, thanks!

PR #2525, opened 2024-02-14
DOCS-1745: Remove `disable_nmea` option from movement sensor

sguequierre:
* Remove disable nmea option from gps-nmea movement sensors-- this is the only model where it was


andf-viam (approved):
LGTM!

PR #2527, opened 2024-02-14
DOCS-1818: Add new SVG icons

andf-viam:
Add new SVG icons for Data Management service and Registry content.
- Most updates to Data Management, Registry, and Use Cases sections.

QUESTION:
- Do we want to enable icons [here](https://docs.viam.com/data/#on-device-data-management) and [here](https://docs.viam.com/data/#cloud-data-management) now that we have them?


npentrel:
I think where possible we should add icons as long as we don't use the same icon in too many places (that can then get confusing

npentrel (approved):

PR #2532, opened 2024-02-15
Update image alignment on mobility page

npentrel:


JessamyT (approved):
LGTM! Why is it that the p tags are necessary on the frame system image but not any other gifs or images?

npentrel:
If an image is before text somehow doesn't end up with the same width constraints as when it's below text. It's not a problem unless you also align right. There's probably a better way to fix it but this does fix it for now

PR #2544, opened 2024-02-16
DOCS-1767: Add ml models table

sguequierre:
WIP

### DIFF
--- a/assets/js/mlmodels.js
+++ b/assets/js/mlmodels.js
@@ -1,0 +1,98 @@
+const { TypesenseInstantSearchAdapter, instantsearch } = window;
+
+const model = document.getElementsByClassName("mr-model")[0].id;
+console.log(model)
+const typesenseInstantsearchAdapter = new TypesenseInstantSearchAdapter({
+  server: {
+    apiKey: "Qhooem9HCRuFMVZPNQOhABAdEWJaSnlY", // Be sure to use an API key that only allows search operations
+    nodes: [
+      {
+        host: "cgnvrk0xwyj9576lp-1.a1.typesense.net",
+        port: "443",
+        protocol: "https",
+      },
+    ],
+    cacheSearchResultsForSeconds: 2 * 60, // Cache search results from server. Defaults to 2 minutes. Set to 0 to disable caching.
+  },
+  // The following parameters are directly passed to Typesense's search API endpoint.
+  //  So you can pass any parameters supported by the search endpoint below.
+  //  query_by is required.
+  additionalSearchParameters: {
+    query_by: "model-id,description",
+    sort_by: "total_organization_usage:desc,total_robot_usage:desc",
+    infix: "always"
+  },
+});
+const searchClient = typesenseInstantsearchAdapter.searchClient;
+
+const search = instantsearch({
+  indexName: "mlmodels",
+  searchClient,
+});
+
+let filters;
+let itemtemplate;
+
+if (model == "") {
+  filters = {
+    hitsPerPage: 5,
+  };
+  itemtemplate = `
+  <div class="type"><p><code>{{#helpers.highlight}}{ "attribute": "model" }{{/helpers.highlight}}</code></p></div>
+  <div class="name"><p><a href="{{url}}"><code>{{#helpers.highlight}}{ "attribute": "model" }{{/helpers.highlight}}</code></a></p></div>
+  <div class="description">{{#helpers.highlight}}{ "attribute": "description" }{{/helpers.highlight}}</div>
+  `;
+} else {
+  filters = {
+    facetFilters: ["model: " + model],
+    hitsPerPage: 5,
+  };
+  itemtemplate = `
+  <div class="name"><p><a href="{{url}}"><code>{{#helpers.highlight}}{ "attribute": "model" }{{/helpers.highlight}}</code></a></p></div>
+  <div class="description">{{#helpers.highlight}}{ "attribute": "description" }{{/helpers.highlight}}</div>
+  `;
+}
+
+
+search.addWidgets([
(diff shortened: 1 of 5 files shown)


assets/js/mlmodels.js
```diff
@@ -0,0 +1,98 @@
+const { TypesenseInstantSearchAdapter, instantsearch } = window;
```
npentrel, line 1: The way these files get included by the browser, every variable in this file will cause a name collision. Try to change the variable names and remove this first line. They'll alsready be included by the models.js file. We'll want to make sure that these can't load in the wrong order, so we'll probably have to put the contents of this file in the models.js file after all, but you can keep it here for now to keep things separate.

assets/js/mlmodels.js
```diff
@@ -0,0 +1,98 @@
...
+        protocol: "https",
+      },
+    ],
+    cacheSearchResultsForSeconds: 2 * 60, // Cache search results from server. Defaults to 2 minutes. Set to 0 to disable caching.
+  },
+  // The following parameters are directly passed to Typesense's search API endpoint.
+  //  So you can pass any parameters supported by the search endpoint below.
+  //  query_by is required.
+  additionalSearchParameters: {
+    query_by: "model-id,description",
```
npentrel, line 21: You need to use the same field name as in the collection
```suggestion
    query_by: "model_id,description",
```

docs/registry/_index.md
```diff
@@ -40,6 +41,22 @@ See [Machine Learning](/ml/) for more information.
...
+<div class="mr-model" id="">
+  <div class="modellistheader">
+    <div class="type">API</div>
+    <div class="name">Model</div>
+    <div>Description</div>
+  </div>
+<div id="hits" class="modellist">
+</div>
+<div id="pagination"></div>
+</div>
```
npentrel, line 58: id's should be unique across a page. In this case this is part of why this currently isn't working. If you read through the bottom half of the mlmodels.js script it does reference #hits and #searchbox and #searchstats and #pagination. That is why it's not working. It's all name collisions here as well

sguequierre: ahh!! Thanks for explaining!

docs/registry/_index.md
```diff
@@ -40,6 +41,21 @@ See [Machine Learning](/ml/) for more information.
 The Viam registry hosts trained ML models that users have made public, which you can use to deploy classifiers or detectors for your use case onto your robot instead of training your own.
 You can also [upload your own model to the registry](/ml/upload-model/).
 
+You can search the available ML models from the Viam registry here:
+
```
sguequierre, line 45: @npentrel I applied all your suggestions and it is almost ready, but the styling on this table is looking wild and I haven't been able to troubleshoot. Any ideas? I tried looking in "modellist" css rules and I can't tell if anything is acting on the table columns

assets/scss/_styles_project.scss
```diff
@@ -1494,9 +1494,9 @@ a.ais-Pagination-link:hover {
   display: none;
 }
 
-// styling for model list from modular resources
+// styling for model list from registry
 
-.mr-component {
+.mr-component, mr-model {
   margin-bottom: 1rem;
 }
```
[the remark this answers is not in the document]
npentrel: And find where the pagination id gets its styling and add pagination2 there as well

npentrel (commented):
Very good work, try the comments to fix the css

assets/scss/_styles_project.scss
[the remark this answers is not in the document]
sguequierre: @npentrel Thank you!

docs/ml/deploy.md
```diff
@@ -85,7 +86,7 @@ To see more details about a model, open its page in [the registry](https://app.v
 {{<imgproc src="/services/deploy-model.png" resize="700x" alt="Create a machine learning models service with a model to be deployed">}}
 
 {{% /tab %}}
-{{% tab name="Existing Model" %}}
+{{% tab name="Path to Existing Model On Robot" %}}
```
sguequierre, line 89: Don't need to make this change but I have found it a bit confusing w/ just Deploy Model and Existing Model

assets/js/mlmodels.js
npentrel, line 1: So a few things:
- instead of appending `2`, can you append something like `ML`? That way the variable names are a bit more logically named.
- Then can we move then contents of this script into the `models.js` file at the end and wrap the code into an `if` statement so that the ml part only executes if needed?

assets/scss/_styles_project.scss
```diff
@@ -1404,6 +1404,10 @@ ul.sectionlist > li:hover {
   margin: 2rem auto;
 }
 
+#pagination2 {
+  margin: 2rem auto;
+}
```
npentrel, line 1409: I think there's already a #pagination item somewhere in this file, can you merge the two like we did for mr-component and mr-model?

assets/js/mlmodels.js
```diff
@@ -0,0 +1,84 @@
+const mlmodel = document.getElementsByClassName("mr-model")[0].id;
+console.log(mlmodel)
```
npentrel, line 2:
```suggestion
```

assets/js/models.js
```diff
@@ -96,3 +96,94 @@ search.addWidgets([
 ]);
 
 search.start();
+
+const mlmodels = document.getElementsByClassName("mr-model")
+
+if (mlmodels.length !== 0) {
+
+  const mlmodel = document.getElementsByClassName("mr-model")[0].id;
+  console.log(mlmodel)
```
sguequierre, line 105:
```suggestion
```

npentrel (approved):

PR #2466, opened 2024-02-03
DOCS-1600: Add configure a fleet page

JessamyT:

PR #2545, opened 2024-02-17
DOCS-1671: Add fragments reference copy

andf-viam:
Update Build page Fragments entry with additional reference material.
- Include dedicated links to new Configure a Fleet page that covers working with fragments.
- Also link back from "fragments usage" page to now point to "fragments reference" page.


docs/fleet/configure-a-fleet.md
```diff
@@ -16,6 +16,8 @@ If there are differences between your machines, you can use a fragment to quickl
 You can then configure the differing resources separately, outside of the fragment.
 For example, if you have multiple identical rovers but one has an arm attached, you can add the rover configuration fragment (including the motors and base components), and then configure the arm on just that one rover.
 
+See [Fragments](/build/configure/#fragments) for more information.
```
npentrel, line 19: should the fragment explanation live with this page instead maybe?

docs/build/configure/_index.md
```diff
@@ -381,11 +381,42 @@ Switch to **Raw JSON** mode to configure webhooks as follows:
 
 ## Fragments
 
-Fragments are a way of sharing and managing identical {{< glossary_tooltip term_id="resource" text="resource" >}} configuration files across multiple machines.
-For example, if you have multiple machines with the same hardware, wired the same way, you can create and share a fragment and add it to any number of machines.
-When changes are made to the fragment, those changes are automatically carried to all machines that include the fragment in their config.
+You can use fragments to share identical {{< glossary_tooltip term_id="resource" text="resource" >}} configuration files across multiple machines in your fleet.
```
npentrel, line 384: We went back and forth on this phrasing when @JessamyT wrote the configure a fleet page. I don't like the word "identical" here. Because it can be a partial configuration that is shared. And then this can be too easily misconstrued as "it's not possible to use fragments unless all my machines are exactly identical"

docs/build/configure/_index.md
```diff
@@ -381,11 +381,42 @@ Switch to **Raw JSON** mode to configure webhooks as follows:
...
+A fragment can define one, several, or all resources on a machine.
+You can add multiple fragments to a single machine, if needed, or can add additional resources to a machine that has already been configured with a fragment.
+
+When you edit a fragment that you've already deployed to one or more machines, the Viam app updates the configuration on each deployed machine that uses that fragment.
+
+{{< alert title="Alert" color="alert" >}}
+Be cautious when making changes to fragments that have been deployed to production machines.
+We recommend that you create a duplicate fragment, make your desired change to that second fragment, and then deploy that fragment to a test machine that is configured identically to your production machines.
+
+Once you are confidant that your configuration change works as expected, you can safely make the same change to the fragment in use on your production fleet, and the Viam app will deploy that change to all machines using that fragment.
```
npentrel, line 396:
```suggestion
Once you are confident that your configuration change works as expected, you can safely make the same change to the fragment in use on your production fleet, and the Viam app will deploy that change to all machines using that fragment.
```

npentrel (commented):
What is your reasoning for having this info on the configure a machine page vs the configure a fleet page?

docs/build/configure/_index.md
JessamyT: I see that the existing sentence still contains the word "identical"--perhaps I should've changed that in https://github.com/viamrobotics/docs/pull/2466 in addition to changing the wording on the other page--oops! Agree it should be updated here. I think the next sentence with the example of one specific component does help illustrate the point that it could be for just one or more components etc. among many.

docs/build/configure/_index.md
```diff
@@ -381,11 +381,42 @@ Switch to **Raw JSON** mode to configure webhooks as follows:
 
 ## Fragments
 
-Fragments are a way of sharing and managing identical {{< glossary_tooltip term_id="resource" text="resource" >}} configuration files across multiple machines.
-For example, if you have multiple machines with the same hardware, wired the same way, you can create and share a fragment and add it to any number of machines.
-When changes are made to the fragment, those changes are automatically carried to all machines that include the fragment in their config.
+You can use fragments to share identical {{< glossary_tooltip term_id="resource" text="resource" >}} configuration files across multiple machines in your fleet.
+For example, if you have multiple machines with the same camera hardware, wired the same way, you can create a fragment to configure that camera and share it easily across all of your machines, without needing to individually configure the camera component for each machine.
```
JessamyT, line 385: I like this more specific example but, how about changing it to "motor" instead of "camera" because cameras are more likely to just be USB or something versus what we think of as "wired the same way"--a motor usually has several wires and a more complex description of that wiring in its config, such that it's useful to put it in a fragment.

andf-viam: Great point, updating to motor, thank you!

docs/build/configure/_index.md
andf-viam: Updating to just `similar` in both places, good points all!

npentrel (approved):
Brilliant, thanks both!

docs/fleet/configure-a-fleet.md
```diff
@@ -70,3 +87,9 @@ If you need to modify the config of just one machine that uses a fragment you ca
 5. Paste the raw fragment contents into the **Raw JSON** config field.
 6. Click **Save config**.
 7. Now, you can edit the config either in **Raw JSON** mode or in **Builder** mode.
+
+## Next steps
+
+Viam provides several [pre-made fragments](https://app.viam.com/fragments) you can use as a template for writing your own fragments.
```
andf-viam, line 93:
```suggestion
Viam provides several [pre-made fragments](https://app.viam.com/fragments) which you can use as templates for writing your own fragments.
```

PR #2550, opened 2024-02-20
Update local module instructions

npentrel:
I went through these last week and this didn't make sense. While you can add the module and the model in whatever order, the sentence "You can also add the module directly, without first adding its modular component or service:" is confusing, because it doesn't then say you need to add the component or service after? I think pointing out that you can add it in any order isn't super valuable so let's stick with one way or the other

### DIFF
--- a/docs/registry/configure.md
+++ b/docs/registry/configure.md
@@ -380,7 +380,19 @@ See the instructions to [compile your module into an executable](/registry/creat
 
 ### Add a local module
 
-To add a local module on your machine:
+To add a local module on your machine, first add its module, then the component or service it implements:
+
+1. Navigate to the **Config** tab of your machine's page in [the Viam app](https://app.viam.com).
+1. Click on the **Modules** subtab.
+1. Scroll to the **Add local module** section.
+1. Enter a **Name** for this instance of your modular resource.
+1. Enter the [module's executable path](/registry/create/#compile-or-package-your-module).
+   This path must be the absolute path to the executable on your machine's filesystem.
+1. Then, click the **Add module** button, and click **Save config**.
+
+   {{<imgproc src="registry/configure/add-local-module-csi-cam.png" resize="600x" declaredimensions=true alt="The add a local module pane with name 'my-csi-ca' and executable path '/usr/local/bin/viam-csi'">}}
+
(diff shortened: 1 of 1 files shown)


docs/registry/configure.md
```diff
@@ -403,20 +415,6 @@ To add a local module on your machine:
 
 1. Click **Create** to create the modular resource provided by the local module.
 
-You can also add the module directly, without first adding its modular component or service:
```
andf-viam, line 406: Agreed with this deletion on principle, but I think it helps to retain awareness with a single sentence: some users will want to add a local module, and then 4 instances of the motor component it provides for their four-wheel machine. What about:

> Once you've added your local module using steps 1-6, you can repeat steps 7-10 to add as many additional instances of your modular resource as you need.

Or similar.

andf-viam (approved):
LGTM % tiny suggestion! Initially last Sept, eng wanted to provide both approaches, because dev engineers were using both and we weren't sure which would be the one users use / we recommend. But I agree it's time to pick: and this is the way. Right on!

docs/registry/configure.md
```diff
@@ -403,20 +415,6 @@ To add a local module on your machine:
 
 1. Click **Create** to create the modular resource provided by the local module.
```
npentrel, line 417:
```suggestion

Once you've added your local module using steps 1-6, you can repeat steps 7-10 to add as many additional instances of your modular resource as you need.
```

docs/registry/configure.md
```diff
@@ -403,20 +415,7 @@ To add a local module on your machine:
...
-1. Enter the [module's executable path](/registry/create/#compile-or-package-your-module).
-   This path must be the absolute path to the executable on your machine's filesystem.
-1. Then, click the **Add module** button, and click **Save config**.
-
-   {{<imgproc src="registry/configure/add-local-module-csi-cam.png" resize="600x" declaredimensions=true alt="The add a local module pane with name 'my-csi-ca' and executable path '/usr/local/bin/viam-csi'">}}
-
-   This example shows the configuration for adding a [CSI camera](https://github.com/viamrobotics/csi-camera/) as a local module.
-
+Once you've added your local module using steps 1-6, you can repeat steps 7-10 to add as many additional instances of your modular resource as you need.
 ### Edit the configuration of a local module
```
npentrel, line 419:
```suggestion

### Edit the configuration of a local module
```

PR #2563, opened 2024-02-21
Change icon frontmatter field

npentrel:


docs/components/component/_index.md
```diff
@@ -6,9 +6,8 @@ weight: 10
 type: "docs"
 description: "A NAME is a ... description of what the component is."
 tags: ["camera", "components"]
-icon: "/icons/components.png"
-# icon: "path/to/svg/file"
-images: ["/icons/components.png"]
+# icon: true
+# images: ["/icons/components.svg"]
```
JessamyT, line 10: Why don't we want an icon for this page?

docs/data/_index.md
```diff
@@ -13,7 +13,8 @@ aliases:
   - "/data-management/"
   - "/data-management/"
   - "/services/data/"
-icon: "/services/icons/data-management.svg"
+icon: true
+images: ["/services/icons/data-management.svg"]
 imageAlt: "The data management service."
```
JessamyT, line 18: Lots of these have sorta meaningless imageAlt...do you want to delete it too in this PR?

layouts/partials/card.html
```diff
@@ -40,7 +40,7 @@
         </div>
       {{ else if .Params.icon }}
         <div class="hover-card-icon">
-          {{ partial "imgproc.html" (dict "src" (.Params.icon) "resize" "400x" "alt" (.Params.linkTitle) "style" "") }}
+          {{ partial "imgproc.html" (dict "src" (index .Params.images 0) "resize" "400x" "alt" (.Params.linkTitle) "style" "") }}
```
JessamyT, line 43: If this is changing to be identical to line 33, how about we get rid of icon altogether? The no imgAlt case is already covered by the recent changes so maybe we don't need it anymore?

JessamyT: I see that we use class `hover-card` but do we need `hover-card-icon` anymore? It appears it's used with no-js situations? Is the specific css for it necessary or can we merge the two?

docs/components/component/_index.md
npentrel: I mean it's a template and the svg doesn't exist, I can use the png though.

npentrel, line 10:
```suggestion
icon: true
images: ["/icons/components.png"]
```

layouts/partials/card.html
npentrel: We do still need the `hover-card-icon` class, the icons get treated a tiny bit differently in css. `object-fit` changes to contain instead of `cover` which makes the components page have cards that contain the entire icon instead of being zoomed in. On smaller screens it also makes icons take up 30% of the card width instead of 50%. It's not an image so it doesn't need to be that big. It looks odd to me with icons that big, that's why they're treated differently.

npentrel: This is the difference
Screenshot 2024-02-22 at 11 42 20

Screenshot 2024-02-22 at 11 42 35

JessamyT: Okay! When I built the site before some updates here, I saw that the generated html didn't seem to include hover-card-icon anywhere but now I see it's using it again after the recent commits!

JessamyT (approved):

PR #2582, opened 2024-02-26
DOCS-1639: Recommend BSON date

andf-viam:
Recommend that MQL query users use the built-in BSON date type instead of converting from string.
- Usage example include both usages users care about:
   - Using date type for explicit date specified as string to `Date()` constructor
   - Using `Date()` constructor without argument to use current timestamp.
   
Question:
- I can link to [BSON spec itself](https://bsonspec.org/spec.html) if too many MDB docs links already, up to you.


andf-viam:
Hi @dmhilly - quick clarification to recommend using the BSON date data type in MQL queries, with an example based on Jack's code as linked in the DOCS ticket. LMK if you'd like to see any changes -- thanks!

docs/data/query.md
```diff
@@ -202,6 +202,31 @@ See the [MongoDB Atlas Documentation](https://www.mongodb.com/docs/atlas/data-fe
 
 For information on connecting to your Atlas instance from other MQL clients, see the MongoDB Atlas [Connect to your Cluster Tutorial](https://www.mongodb.com/docs/atlas/tutorial/connect-to-your-cluster/).
 
+#### Query by date
+
+When using MQL to query your data by date or time range, you can optimize query performance by avoiding the MongoDB `$toDate` expression, using the [BSON `date` type](https://www.mongodb.com/docs/manual/reference/bson-types/#date) instead.
+
+For example, you could use the following to query a dataset over a date range, specifying a target start timestamp in BSON date format, and using the `Date()` constructor to use the current time as the end timestamp:
```
dmhilly, line 209:
```suggestion
For example, you could use the following to query a dataset over a date range, specifying a target start timestamp in BSON date format, and using the JavaScript `Date()` constructor to use the current time as the end timestamp:
```
Thoughts? Just to make it clear that this is specific to JavaScript/the shell, since to construct a date type in another language they will need to use the language-specific constructs

docs/data/query.md
```diff
@@ -202,6 +202,31 @@ See the [MongoDB Atlas Documentation](https://www.mongodb.com/docs/atlas/data-fe
...
+
+```mongodb {class="command-line" data-prompt=">"}
+// Switch to sensorData database:
+use sensorData
+
+// Set desired start and end times:
+const startTime = 'Mon Feb 10 2024 14:45:07 GMT-0500 (Eastern Standard Time)'
```
dmhilly, line 216: Ah I just noticed that this is a string type, not a date. That might confuse users. I don't really know JS but maybe we want something like:

```js
const startTime = new Date('Mon Feb 10 2024 14:45:07 GMT-0500 (Eastern Standard Time)')
```

docs/data/query.md
andf-viam: Right on, this is indeed continuing the `mongoh` example, so JS it is. Clarified this a bit, thank you!

docs/data/query.md
andf-viam: Ah, you're right I think, and we can simplify it a bit too if we're using `Date()` then. Thanks!

dmhilly (approved):
The changes look great - LGTM! 🎉

npentrel (approved):

PR #2584, opened 2024-02-26
DOCS-1887: Add WSL note

andf-viam:
Add note that Windows users can install WSL to use `viam-server` "on windows".
- Including direct link to WSL install instructions

NOTE:
- Alternate: remove alert styling for notice, to draw less attention to it, if desired.
- Not adding Windows to "Viam-server supports:" list, since it's via WSL which is Linux, which is already listed. "Windows" as listed in this list would encourage users thinking of the .NET C# thing underneath the Linux, which we most definitely do not support.

### DIFF
--- a/docs/get-started/installation/_index.md
+++ b/docs/get-started/installation/_index.md
@@ -28,7 +28,7 @@ Installing on a laptop or desktop computer is useful if you don't have an SBC av
 
 `viam-server` supports:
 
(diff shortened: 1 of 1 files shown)


npentrel (approved):
This looks good to me. It may be a good idea for you to look at the associated app design on the other ticket to see if that all looks reasonable

PR #2596, opened 2024-02-28
Update mlmodel.md TF Lite caps

JessamyT:


andf-viam (approved):
Thanks - you're right!

PR #2598, opened 2024-02-29
Update model version and description

WWWWWWWWWWWMWWWWWWWWWWWWWWWWWWWWWWWWWWW:
# Description

Model version format is outdated in 3 documentation pages (the only 3 occurrences). Updated formats, description and package config that was missing "type" key in 2 out of 3 places. Wasn't sure people would understand `T` in `YYYY-MM-DDThh-mm-ss` is just a letter 'T' so also gave an example of a version.


npentrel (approved):
Thank you!

PR #2604, opened 2024-03-01
Update _index.md

nataliajacobowitz:
changing arm64-->aarch64 to be consistent with the language we use in our app and elsewhere in docs

# Description


andf-viam (approved):
This LGTM, thank you!

PR #2606, opened 2024-03-01
DOCS-1909: Remove duplicate status method

skyleilani:
There seems to be a duplicate of the robot client `Status` method. I removed the duplicate definition without a TS version.

Screenshot 2024-03-01 at 12 04 21 PM
 
Screenshot 2024-03-01 at 12 04 28 PM


npentrel (approved):
Great find!

PR #2611, opened 2024-03-05
DOCS-1899: Remove update from module create

andf-viam:
Remove `viam module update` from the module creation workflow steps
- Rationale: less steps better, as no use case exists for manual run of `update` (i.e. without `upload` immediately following).
	-  If not, please correct!
- Adjust CLI docs to reflect that `upload` automatically runs `update`, but retain mention there as the command still exists.


npentrel (approved):
Nice change

andf-viam:
Hi @michaellee1019, quick update to remove `update` from the upload steps. LMK if you'd like to see any changes; thank you!

PR #2638, opened 2024-03-11
DOCS-1897: Add attributes to ML model vision service

sguequierre:
* Make attributes table
* Add new attributes
* Clean up rest of page a bit

### DIFF
--- a/docs/ml/vision/mlmodel.md
+++ b/docs/ml/vision/mlmodel.md
@@ -125,9 +125,9 @@ The following attributes are available for an `mlmodel` detector or classifier:
 | `mlmodel_name` | string | **Required** | The name of the [ML model service](/ml/deploy/) you want to use the model from. |
 | `remap_output_names` | object | Optional | The names of your output tensors, mapped to the service requirements. See [Tensor names](#tensor-names) for more information. |
 | `remap_input_names` | object | Optional | The name of your input tensor, mapped to the service requirements. See [Tensor names](#tensor-names) for more information. |
-| `input_image_bgr` | bool | Optional | Set this to `true` if the ML model service expect the input image to have BGR pixels, rather than RGB pixels. |
-| `input_image_mean_value` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. |
-| `input_image_std_dev` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. |
+| `input_image_bgr` | bool | Optional | Set this to `true` if the ML model service expects the input image to have BGR pixels, rather than RGB pixels. <br> Default: `false` |
+| `input_image_mean_value` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. <br> Default: `[0.5, 0.5, 0.5]` |
+| `input_image_std_dev` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. <br> Default: `[0.5, 0.5, 0.5]` |
 
 ## Tensor names
 
--- a/docs/ml/vision/mlmodel.md
+++ b/docs/ml/vision/mlmodel.md
@@ -143,8 +143,6 @@ Both the `mlmodel` detector and classifier require that the input and output ten
 If you [trained your ML model using the Viam app](/ml/train-model/), your `mlmodel` tensors are already named in this fashion, and you can proceed to [test your detector or classifier](#test-your-detector-or-classifier).
 However, if you [uploaded your own ML model](/ml/upload-model/), or are using one from the [Viam registry](https://app.viam.com/registry), you may need to remap your tensor names to meet this requirement, and should follow the instructions to [remap tensor names](#remap-tensor-names).
 
-{{%expand "Click for details on how to remap tensor names." %}}
-
 ### Remap tensor names
 
 If you need to remap the tensor names defined by your ML model to meet the tensor name requirements of the `mlmodel` detector or classifier, you can use the `remap_input_names` and `remap_output_names` attributes:
@@ -199,8 +197,6 @@ To remap your model's tensor names to work with an `mlmodel` classifier, add the
 {{< /tabs >}}
 
 When done, click **Save config**, then proceed to [test your detector or classifier](#test-your-detector-or-classifier).
-
-{{% /expand%}}
 
 ## Test your detector or classifier
 
--- a/docs/ml/vision/mlmodel.md
+++ b/docs/ml/vision/mlmodel.md
@@ -31,7 +31,6 @@ You can [add an existing model](/ml/upload-model/) or [train your own models](/m
 <h4>Deploy your model</h4>
 
 To make use of ML models with your machine, use the built-in [ML model service](/ml/) to deploy and run the model.
-s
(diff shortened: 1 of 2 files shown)


docs/ml/vision/mlmodel.md
```diff
@@ -186,7 +211,7 @@ You can also test classifiers with [existing images on a computer](#existing-ima
 
 {{< alert title="Note" color="note" >}}
 
-The feature is only available for classifiers that were uploaded after September 19, 2023.
+The feature is only available for classifiers that are uploaded after September 19, 2023.
```
sguequierre, line 214:
```suggestion
The feature is only available for classifiers that were uploaded after September 19, 2023.
```

docs/ml/vision/mlmodel.md
```diff
@@ -31,7 +31,7 @@ You can [add an existing model](/ml/upload-model/) or [train your own models](/m
 <h4>Deploy your model</h4>
 
 To make use of ML models with your machine, use the built-in [ML model service](/ml/) to deploy and run the model.
-
+s
```
sguequierre, line 34:
```suggestion

```

docs/ml/vision/mlmodel.md
```diff
@@ -31,7 +31,6 @@ You can [add an existing model](/ml/upload-model/) or [train your own models](/m
 <h4>Deploy your model</h4>
 
 To make use of ML models with your machine, use the built-in [ML model service](/ml/) to deploy and run the model.
-
 {{% /manualcard %}}
```
sguequierre, line 34:
```suggestion

{{% /manualcard %}}
```

docs/ml/vision/mlmodel.md
```diff
@@ -117,6 +117,18 @@ Add the vision service object to the services array in your raw JSON configurati
...
+
+<!-- prettier-ignore -->
+| Parameter | Type | Inclusion | Description |
+| --------- | ---- | --------- | ----------- |
+| `mlmodel_name` | string | **Required** | The name of the [ML model service](/ml/deploy) you want to use the model from. |
+| `remap_output_names` | object | Optional | The names of your output tensors, mapped to the service requirements. See [Tensor names](#tensor-names) for more information. |
+| `remap_input_names` | object | Optional | The name of your input tensor, mapped to the service requirements. See [Tensor names](#tensor-names) for more information. |
+| `input_image_bgr` | bool | Optional | Set this to `true` if the ML model service expect the input image to have BGR pixels, rather than RGB pixels. |
+| `input_image_mean_value` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. |
+| `input_image_std_dev` | array | Optional | The standard deviation of the RGB (or BGR) values. Only required if the ML model service expects the input image to be normalized. |
```
sguequierre, line 130: @bhaney questions here: in jira you phrased as "ML model expects" so just want to confirm it's ML model service and not the vision service?
- How would a user know if the ML model expects the input image to be normalized (or have BGR pixels)?
- What are the defaults for input_image_mean_value and input_image_std_dev? I couldn’t find in code. Also, input_image_bgr default I'm assuming is true?

docs/ml/vision/mlmodel.md
```diff
@@ -117,6 +117,18 @@ Add the vision service object to the services array in your raw JSON configurati
 
 Click **Save config**.
 
+The following attributes are available for an `mlmodel` detector or classifier:
+
+<!-- prettier-ignore -->
+| Parameter | Type | Inclusion | Description |
+| --------- | ---- | --------- | ----------- |
+| `mlmodel_name` | string | **Required** | The name of the [ML model service](/ml/deploy) you want to use the model from. |
```
sguequierre, line 125:
```suggestion
| `mlmodel_name` | string | **Required** | The name of the [ML model service](/ml/deploy/) you want to use the model from. |
```

docs/ml/vision/mlmodel.md
```diff
@@ -128,55 +140,68 @@ Both the `mlmodel` detector and classifier require that the input and output ten
   - The _input tensor_ must be named `image`
   - The _output tensor_ must be named `probability`
 
-If you [trained your ML model using the Viam app](/ml/train-model/), your `mlmodel` tensors is already named in this fashion, and you can proceed to [test your detector or classifier](#test-your-detector-or-classifier).
+If you [trained your ML model using the Viam app](/ml/train-model/), your `mlmodel` tensors are already named in this fashion, and you can proceed to [test your detector or classifier](#test-your-detector-or-classifier).
 However, if you [uploaded your own ML model](/ml/upload-model/), or are using one from the [Viam registry](https://app.viam.com/registry), you may need to remap your tensor names to meet this requirement, and should follow the instructions to [remap tensor names](#remap-tensor-names).
 
+{{%expand "Click for details on how to remap tensor names." %}}
```
andf-viam, line 146: Nit: Why hide this content within an expander? I see that it streamlines the page a little, but then loses the flow we get from things like the `[remap tensor names](#remap-tensor-names)` link that can't link within the expander. Personally, I'd present this section as a regular old subheader `### Remap tensor names` without using the expander, but up to you if you want to retain.

andf-viam (approved):
LGTM from docs perspective, mod optional nit. Thanks!

docs/ml/vision/mlmodel.md
npentrel: seems small enough to take out of the expander

npentrel (approved):

docs/ml/vision/mlmodel.md
bhaney: - Correct, it's the ML Model service that is expecting these things
- Unfortunately, you just have to know about the model you've chosen. Models are not really standardized - people create the inputs and outputs based on their whims. Places like google and meta have standard, which we have baked into the default behavior of the vision service, but if you are using an ML Model service that loads a model you got from a random person online, or that you built yourself, only the user will know if they need these options or not.
- correct, input_image_bgr is false by default, meaning is assumes images come in an RGB format. the default forn mean_value is [0.5, 0.5, 0.5] and the default for std_dev is also [0.5, 0.5, 0.5].

bhaney (approved):
Looks good - though I will say, the vision service also takes into account an edge case for a "detector" ML Model service that might be complicated to describe. 

Just for your info, an ML Model Service that implements a detector sometimes might only have two output tensors: `location` and `score`. Some ml model services will stuff the category information into a extra dimension in the output tensor of `score`. The vision service can recognize that case if there are only two output tensors called `location` and `score` and the dimensional of the`score` output tensor is compensating for the extra info. I could go into more details if you'd like, but honestly this looks great as it is.

docs/ml/vision/mlmodel.md
sguequierre: @bhaney perfect, thanks!

sguequierre:
> Looks good - though I will say, the vision service also takes into account an edge case for a "detector" ML Model service that might be complicated to describe.
> 
> Just for your info, an ML Model Service that implements a detector sometimes might only have two output tensors: `location` and `score`. Some ml model services will stuff the category information into a extra dimension in the output tensor of `score`. The vision service can recognize that case if there are only two output tensors called `location` and `score` and the dimensional of the`score` output tensor is compensating for the extra info. I could go into more details if you'd like, but honestly this looks great as it is.

We definitely could document that if you think it worthwhile / it ends up coming up.

PR #2644, opened 2024-03-12
DOCS-1914: Add resource name requirements to configure a machine

sguequierre:
* Add requirements from regex

Choose not to add regex itself but can if think appropriate?


npentrel (approved):
This is good - don't include the regex itself

PR #2650, opened 2024-03-13
DOCS-1720: Include digital interrupts config and control

skyleilani:
Add a control section for digital interrupts with Config Builder instructions.

Since [`analogs`](https://docs.viam.com/components/board/#analogs) isn't in the scope of this ticket I can open a separate ticket to keep uniformity on the board page


npentrel (approved):
awesome! Thanks! Just one missing word, then good to merge

PR #2653, opened 2024-03-13
Remove unnecessary --save option

WWWWWWWWWWWMWWWWWWWWWWWWWWWWWWWWWWWWWWW:
# Description

`--save` option is no longer necessary when installing JS/TS packages. It's the default since [npm 5.0.0](https://blog.npmjs.org/post/161081169345/v500) which was published in 2017.

[NPM documentation](https://docs.npmjs.com/cli/v6/commands/npm-install) doesn't even list `--save` as an available option.

This is the only occurrence of `--save`.

### DIFF
--- a/docs/build/program/_index.md
+++ b/docs/build/program/_index.md
@@ -59,7 +59,7 @@ go get go.viam.com/rdk/robot/client
 {{% tab name="TypeScript" %}}
 
 ```sh {class="command-line" data-prompt="$"}
-npm install --save @viamrobotics/sdk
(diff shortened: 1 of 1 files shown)


npentrel (approved):
Thank you very much - I appreciate all your help!

WWWWWWWWWWWMWWWWWWWWWWWWWWWWWWWWWWWWWWW:
Thank you for the kind words, Naomi. :slightly_smiling_face:

PR #2674, opened 2024-03-19
DOCS-1684: Clean up module configure page R2D2

sguequierre:
This ticket was to update the section `#configure-version-update-management-for-a-registry-module` but I found I actually already did that while working on another ticket-- then looking at the section saw some potential for comprehension improvement 

- Move `Edit modular resources` out from "Add additional modular resources from a registry module" and to "Edit the configuration of a module from the Viam registry" -- I would say module includes resources and the info was getting a little lost above
- Add better image and deleted unused image
- Make images on page less blurry with Jessamy's resize trick


docs/registry/configure.md
```diff
@@ -209,18 +215,18 @@ The custom model is configured as a component with the name "my-realsense".
 {{% /tab %}}
 {{% /tabs %}}
 
-## Edit the configuration of a module from the Viam registry
+### Edit the module
 
-Once you have added a modular resource to your machine, you can view and edit the configuration of the underlying module from the **CONFIGURE** tab:
+You can also view and edit the configuration of their underlying module from the **CONFIGURE** tab:
```
sguequierre, line 220:
```suggestion
You can also view and edit the configuration of a modular resource's underlying module from the **CONFIGURE** tab:
```

PR #2678, opened 2024-03-20
Remove mention of Oak-D and Intel Real Sense from Dual and Single Server Camera model Docs

hexbabe:
The way to consume these cameras now is by using the modules.

# Description


JessamyT (approved):
LGTM assuming there's no use case for this anymore at all, like for other camera models?

npentrel (approved):

sguequierre (approved):

PR #2704, opened 2024-03-28
Update parts-and-remotes to add remote instructions for select

npentrel:
# Description


andf-viam (approved):
This LGTM! Tiny note: seems `SELECT` mode is now the default, user doesn't need to select it. Fine either way, though.

PR #2712, opened 2024-04-03
DOCS-1574: Fix use case page nits

JessamyT:
The double asterisks appear for "Test your components", and adding a line break fixed that problem but makes the linter mad about emphasis used instead of a heading, so this disables that MD rule for that section.

For Deploy, this fixes a copy pasta error.


npentrel (approved):
good catch - I will merge this now since it's a small change and I don't like seeing the `**` in the docs...

PR #2733, opened 2024-04-05
DOCS-2102: Update servo builder mode 

sguequierre:
* Update screenshots and instructions for R2D2 interface


assets/components/servo/gpio-servo-ui-config.png
JessamyT, line 1: [opt] Might be nice to collapse the board config for this one like you did on the pi servo one, so that the screenshot is more compact on the page, and so that the emphasis is on the servo config rather than the board config.

docs/components/servo/pi.md
```diff
@@ -28,38 +28,9 @@ Click the **+** icon next to your machine part in the left-hand menu and select
 Select the `servo` type, then select the `pi` model.
 Enter a name for your servo and click **Create**.
 
-{{< imgproc src="/components/servo/pi-servo-ui-config.png" alt="An example configuration for a pi servo in the Viam app Config Builder." resize="600x" >}}
+![An example configuration for a pi servo in the Viam app Config Builder.](/components/servo/pi-servo-ui-config.png)
```
JessamyT, line 31: ooh should these all be standardized? Is there a reason you changed this such that I should be using the MD syntax instead of imgproc for my component screenshot update PRs or is it arbitrary and doesn't matter?

JessamyT (approved):
LGTM % opt collapse the one board config card

docs/components/servo/pi.md
sguequierre: I changed it because I noticed an issue with resizing (image not shrinking with page) on the mmovement sensor pr images-- looked over it with Naomi and `resize="1200x" style="width:650px"` instead of `max-width` fixes the issue and allows the images to be slightly bigger-- would suggest that. I think either way is fine but imgproc resized looks a bit better than markdown so I'm going to change that back here

PR #2757, opened 2024-04-11
Update query.md alt text

JessamyT:
Trying to make this more accessible; screenshot includes specific strings that I want to represent and maybe this will work per https://forum.freecodecamp.org/t/escaping-characters-accessibility/439122/4. If this doesn't work, maybe we can put all these examples into code blocks instead of screenshots.


JessamyT:
Syntax not working; closing and making a ticket to revisit. https://viam.atlassian.net/browse/DOCS-2133

PR #2759, opened 2024-04-11
Update configure.md for emphasis

JessamyT:


andf-viam (approved):
I think this is a good change, LGTM!

PR #2763, opened 2024-04-12
Add python WSL notice

skyleilani:


docs/build/program/_index.md
```diff
@@ -38,12 +38,16 @@ Next, to install your preferred Viam SDK on your Linux or macOS development mach
 
 If you are using the Python SDK, [set up a virtual environment](/build/program/python-venv/) to package the SDK inside before running your code, avoiding conflicts with other projects or your system.
 
+For macOS (both Intel `x86_64` and Apple Silicon) or Linux (`x86`, `aarch64`, `armv6l`), run the commands outlined below.
```
npentrel, line 41: Vale should be underlining "below". Both below and above aren't great because it doesn't work for blind people. It's more accurate to say the following 
```suggestion
For macOS (both Intel `x86_64` and Apple Silicon) or Linux (`x86`, `aarch64`, `armv6l`), run the following commands:
```

docs/build/program/_index.md
```diff
@@ -38,12 +38,16 @@ Next, to install your preferred Viam SDK on your Linux or macOS development mach
...
+
 ```sh {class="command-line" data-prompt="$"}
 python3 -m venv .venv
 source .venv/bin/activate
 pip install viam-sdk
 ```
 
+If you are using Windows, install `viam-sdk` in [WSL](https://learn.microsoft.com/en-us/windows/wsl/install).
+
```
npentrel, line 50: This is slightly awkward because WSL is technically linux. I think if we 

```suggestion
Windows is not supported. If you are using Windows, use the [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install) and install the Python SDK using the preceding instructions.
For other unsupported systems, see [Installing from source](https://python.viam.dev/#installing-from-source).
```

npentrel (commented):
Good place to put these I think. I'm making slight changes for clarity and style

npentrel (approved):

PR #2767, opened 2024-04-12
DOCS-2121: Update base remote control for R2D2

sguequierre:
This one was pretty simple as they decided to leave card json mode but updated instructions a bit with screenshot?


npentrel (approved):
LGTM % nit

PR #2806, opened 2024-04-18
DOCS-2161: Replace AnalogReader with Analog Pt2

skyleilani:
I think these are leftover instances from initial cleanup


npentrel (approved):
Good catch!

PR #2810, opened 2024-04-18
DOCS-2154: Fix css bug on videos

npentrel:
# Description


JessamyT (approved):
Ah so this is how you get it to be pretty on small screens! LGTM

PR #2814, opened 2024-04-19
DOCS-2162: Remove support for some builtin camera models

npentrel:


docs/components/camera/_index.md
```diff
@@ -59,11 +59,6 @@ For configuration information, click on the model name:
 | [`velodyne`](velodyne/) | Uses velodyne lidar. |
 | [`webcam`](webcam/) | A standard camera that streams camera data. |
 | [`fake`](fake/) | A camera model for testing. |
-| [`single_stream`](single-stream/) | A HTTP client camera that streams image data from an HTTP endpoint. |
-| [`dual_stream`](dual-stream/) | A HTTP client camera that combines the streams of two camera servers to create colorful point clouds. |
-| [`join_color_depth`](join-color-depth/) | Joins the outputs of a color and depth camera already registered in your config to create a third "camera" that outputs the combined and aligned image. |
```
hexbabe, line 64: I think we should only be deleting `single_stream`, `dual_stream`, `align_color_depth_extrinsics`, `align_color_depth_homography`, not the join cameras. The join cameras should have their docs modified later after we make the module and run a migration script to have all robot configs to subscribe to the module.

hexbabe (changes requested):

docs/components/camera/_index.md
npentrel: well the `join_color_depth` page seems to indicate that you need to pass it the two cameras we're removing so I don't see how that makes sense

npentrel: ok will fix

npentrel: done

hexbabe (approved):
Thank you for getting this out quickly! LGTM

PR #2823, opened 2024-04-22
[RSDK-7278] Update minimum hardware esp32

npmenard:
Since the latest release of micro-rdk we have dropped support of older models (ESP32-WROOM)  I have updated the minimum required hardware to reflect that.


npentrel (approved):
Thank you @npmenard !

PR #2834, opened 2024-04-24
Update provisioning to include typescript

npentrel:
@ale7714 there is nothing better to link to for typescript than the implementation code currently right?


ale7714 (approved):
Yes! For now, the implementation is the only that we have to publish.

PR #2835, opened 2024-04-24
DOCS-2022: Document `StreamTicks()`

sguequierre:
* Add in Python and Go

### DIFF
--- a/docs/components/board/_index.md
+++ b/docs/components/board/_index.md
@@ -546,6 +546,71 @@ err := myBoard.WriteAnalog(context.Background(), "11", 48, nil)
 {{% /tab %}}
 {{< /tabs >}}
 
+### StreamTicks
+
+Start a stream of [`DigitalInterrupt`](/components/board/#digital_interrupts) ticks.
+
+{{< tabs >}}
+{{% tab name="Python" %}}
+
+**Parameters:**
+
+- `interrupts` (List[[DigitalInterrupt](https://python.viam.dev/autoapi/viam/components/board/board/index.html#viam.components.board.board.Board.DigitalInterrupt)]): List of digital interrupts to receive ticks from.
+
+**Returns:**
+
+- [(`TickStream`)](https://python.viam.dev/autoapi/viam/components/board/board/index.html#viam.components.board.board.TickStream): [Stream](https://python.viam.dev/autoapi/viam/streams/index.html#viam.streams.Stream) of [Ticks](https://python.viam.dev/autoapi/viam/proto/component/board/index.html#viam.proto.component.board.StreamTicksResponse), objects containing `pin_name`, `time`, and `high` fields.
+
+```python {class="line-numbers linkable-line-numbers"}
+my_board = Board.from_robot(robot=robot, name="my_board")
+
+di8 = await my_board.digital_interrupt_by_name(name="8")
+di11 = await my_board.digital_interrupt_by_name(name="11")
+
+# Stream ticks from the listed digital interrupts on pins 8 and 11.
+ticks = await my_board.stream_ticks([di8, di11])
+```
+
+For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/viam/components/board/client/index.html#viam.components.board.client.BoardClient.stream_ticks).
+
+{{% /tab %}}
+{{% tab name="Go" %}}
+
+**Parameters:**
+
+- `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
+- `interrupts` [([]DigitalInterrupt)](https://pkg.go.dev/go.viam.com/rdk/components/board#DigitalInterrupt): Slice of digital interrupts to receive ticks from.
+- `ch` ([chan](https://go.dev/tour/concurrency/2) [Tick](https://pkg.go.dev/go.viam.com/rdk/components/board#Tick)): The channel to stream Ticks, structs containing `Name`, `High`, and `TimestampNanosec` fields.
+- `extra` [(map\[string\]interface{})](https://go.dev/blog/maps): Extra options to pass to the underlying RPC call.
+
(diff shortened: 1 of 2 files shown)


docs/components/board/_index.md
```diff
@@ -546,6 +546,71 @@ err := myBoard.WriteAnalog(context.Background(), "11", 48, nil)
...
+
+if di8, err := myBoard.DigitalInterruptByName("8"); err == nil {
+    interrupts = append(interrupts, di8)
+}
+if di11, err := myBoard.DigitalInterruptByName("11"); err == nil {
+    interrupts = append(interrupts, di11)
+}
+
+// Stream ticks on ticksChan from the listed digital interrupts on pins 8 and 11.
+err := myBoard.StreamTicks(context.Background(), interrupts, ticksChan, nil)
```
sguequierre, line 608: @oliviamiller Is `err := ` incorrect here? I cribbed this partially from [adxl345 code](https://github.com/viamrobotics/rdk/blob/b1ad5688640d562a20282b297b3337fa6e2a2cbf/components/movementsensor/adxl345/adxl345.go#L341) and noticed it was `=` instead but I've never seen that with variable initialization so thought I'd double check

docs/components/board/_index.md
```diff
@@ -546,6 +546,71 @@ err := myBoard.WriteAnalog(context.Background(), "11", 48, nil)
...
+
+For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/viam/components/board/client/index.html#viam.components.board.client.BoardClient.stream_ticks).
+
+{{% /tab %}}
+{{% tab name="Go" %}}
+
+**Parameters:**
+
+- `ctx` [(Context)](https://pkg.go.dev/context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
+- `interrupts` [([]DigitalInterrupt)](https://pkg.go.dev/go.viam.com/rdk/components/board#DigitalInterrupt): Slice of digital interrupts to receive ticks from.
```
sguequierre, line 582: @oliviamiller just want to confirm this is correct and not []string -- looks like was updated but not yet released to Go SDK docs?

docs/components/board/_index.md
```diff
@@ -546,6 +546,71 @@ err := myBoard.WriteAnalog(context.Background(), "11", 48, nil)
...
+- [(`TickStream`)](https://python.viam.dev/autoapi/viam/components/board/board/index.html#viam.components.board.board.TickStream): [Stream](https://python.viam.dev/autoapi/viam/streams/index.html#viam.streams.Stream) of [Ticks](https://python.viam.dev/autoapi/viam/proto/component/board/index.html#viam.proto.component.board.StreamTicksResponse), objects containing `pin_name`, `time`, and `high` fields.
+
+```python {class="line-numbers linkable-line-numbers"}
+my_board = Board.from_robot(robot=robot, name="my_board")
+
+di8 = await my_board.digital_interrupt_by_name(name="8")
+di11 = await my_board.digital_interrupt_by_name(name="11")
+
+# Stream ticks from the listed digital interrupts on pins 8 and 11.
+ticks = await my_board.stream_ticks([di8, di11])
```
sguequierre, line 571: @oliviamiller noticed code sample doesn't use await but this is async-- is that because of stream or would it normally use await? can remove 

```suggestion
ticks = my_board.stream_ticks([di8, di11])
```

docs/components/board/_index.md
```diff
@@ -1518,7 +1583,7 @@ interrupt.add_callback(callback_queue)
 
 **Parameters:**
 
-- `callback` [(chan Tick)](https://go.dev/tour/concurrency/2): The channel to add as a listener for when the state of the GPIO pin this interrupt is [configured for](#digital_interrupts) changes between high and low.
+- `callback` ([chan](https://go.dev/tour/concurrency/2) [Tick](https://pkg.go.dev/go.viam.com/rdk/components/board#Tick)): The channel to add as a listener for when the state of the GPIO pin this interrupt is [configured for](#digital_interrupts) changes between high and low.
```
oliviamiller, line 1586: just FYI that `Close`, `AddCallback` and `Tick` are going to be removed from the interface in the next release. They are also not part of the API so not sure that they ever should have been documented here ?

oliviamiller (approved):

docs/components/board/_index.md
sguequierre: Ok thank you!

sguequierre, line 608:
```suggestion
err = myBoard.StreamTicks(context.Background(), interrupts, ticksChan, nil)
```

docs/components/board/_index.md
sguequierre: Oh interesting. Thanks for letting me know, I can make a pr to remove them. Are there any associated tickets I can track? 

I assume you mean `Close` for the digitalInterrupt type and not for board api? I think we documented them bc they were in the interface on Go (and w/ Python for close)-- sorry for the confusion, how would commands be implemented in the interface without API endpoints (how could we differentiate)?

oliviamiller: yeah Close for the digitalInterrupt, board will still have `Close`. Theres a ticket `here` that was split into multiple PRs so Close is done and others are in progress. 

Ideally the Go interfaces would have parity with the API but we have some components (might just be board) where that is not the case and there are methods on the interface that do not have APIs. Since theres no API they can't be called client side but they do still need to be implemented when you make a driver or module for that component. (so actually maybe these methods should be documented bc of that).

oliviamiller, line 1586: just noticed this, can't comment on the actual lines but the Addcallback section above this can be removed as it won't be implemented anymore.

sguequierre: gotcha! I'll remove in new PR

sguequierre: thanks for the explanation, sounds good!

PR #2846, opened 2024-04-26
Remove setup tab screenshots

npentrel:
Noticed this in Sky's PR. I think these are probably going to change a lot and I wouldn't want people to try to copy by typing from a screenshot rather than navigate there themselves so I think probably best to not include...


sguequierre (approved):
lgtm

PR #2861, opened 2024-05-01
Fix python method and heading name on AnalogByName

sguequierre:
Noticed this while working on automation-- heading was confusing and python method name was incorrect 

AFAIK there is not a proto for this? [ReadAnalogReader](https://github.com/viamrobotics/api/blob/b39c74b5a2d320fa459ef0ec4fac6a3b36a690e3/proto/viam/component/board/v1/board.proto#L66) is what we call `Read` within the analog reader endpoints

### DIFF
--- a/docs/build/micro-rdk/board/_index.md
+++ b/docs/build/micro-rdk/board/_index.md
@@ -46,7 +46,7 @@ See [PWM signals on `esp32` pins](/build/micro-rdk/board/esp32/#pwm-signals-on-e
 
 For `Analog`s:
 
-- [`Read()`](/components/board/#read)
+- [`ReadAnalogReader()`](/components/board/#readanalogreader)
 
 For `DigitalInterrupt`s:
 
--- a/docs/components/board/_index.md
+++ b/docs/components/board/_index.md
@@ -1135,7 +1135,7 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/r
 
 ## `Analog` API
 
-### Read
+### ReadAnalogReader
 
 Read the current integer value of the digital signal output by the [ADC](#analogs).
 
--- a/docs/components/board/_index.md
+++ b/docs/components/board/_index.md
@@ -140,7 +140,7 @@ Additionally, the nested `GPIOPin`, `Analog`, and `DigitalInterrupt` interfaces 
 
 {{< readfile "/static/include/components/apis/digitalinterrupt.md" >}}
 
-### ReadAnalog
+### AnalogByName
 
 Get an [`Analog`](#analogs) pin by `name`.
 
@@ -161,7 +161,7 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 my_board = Board.from_robot(robot=robot, name="my_board")
 
 # Get the Analog "my_example_analog_pin".
-analog = await my_board.analog_by_name(name="my_example_analog")
+analog = await my_board.analog_reader_by_name(name="my_example_analog")
 ```
 
 {{% /tab %}}
--- a/static/include/components/apis/analogreader.md
+++ b/static/include/components/apis/analogreader.md
@@ -1,4 +1,4 @@
 | Method Name                         | Description                                                             |
 | ----------------------------------- | ----------------------------------------------------------------------- |
-| [`Read`](/components/board/#read)   | Read the current integer value of the digital signal output by the ADC. |
+| [`ReadAnalogReader`](/components/board/#readanalogreader)   | Read the current integer value of the digital signal output by the ADC. |
 | [`Close`](/components/board/#close) | Safely shut down the resource and prevent further use.                  |
--- a/static/include/components/apis/board.md
+++ b/static/include/components/apis/board.md
@@ -1,7 +1,7 @@
 <!-- prettier-ignore -->
 Method Name | Description
 ----------- | -----------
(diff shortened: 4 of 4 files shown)


docs/components/board/_index.md
```diff
@@ -161,7 +161,7 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 my_board = Board.from_robot(robot=robot, name="my_board")
 
 # Get the Analog "my_example_analog_pin".
-analog = await my_board.analog_by_name(name="my_example_analog")
+analog = await my_board.analog_reader_by_name(name="my_example_analog")
```
npentrel, line 164: this was changed by https://github.com/viamrobotics/docs/commit/2b889fe6f7a6a9db44f6634845f0080c2810a162 where I believe they were trying to get rid of `reader`. But I don't see `analog_by_name` in the sdk. @randhid can you let us know if that's an issue in the SDK?

docs/components/board/_index.md
```diff
@@ -140,7 +140,7 @@ Additionally, the nested `GPIOPin`, `Analog`, and `DigitalInterrupt` interfaces
 
 {{< readfile "/static/include/components/apis/digitalinterrupt.md" >}}
 
-### ReadAnalog
+### AnalogByName
```
npentrel, line 143: These headings should be the proto names, that's what we standardized on... https://github.com/viamrobotics/api/blob/main/component/board/v1/board_grpc.pb.go#L124
```suggestion
### ReadAnalogReader
```

Again interested to hear what Rand thinks

npentrel (changes requested):
this needs discussion

docs/components/board/_index.md
sguequierre: It's confusing but I believe that proto is used for [Read](https://docs.viam.com/components/board/#read) -- see [client.py](https://github.com/viamrobotics/rdk/blob/f22ff365d533c441e7e697cf76dd65cf51fa9366/components/board/client.go#L158) and the pb description. Maybe we should update that heading? 
I can't find that there is a proto for this method

npentrel: So this one should be AnalogByName and the Read one below should be `ReadAnalogReader`?

sguequierre: Yeah, I think so. I'll update Read too and we can see what Rand thinks when she's back from PTO

randhid:
It being changed by a few epics for renaming board APIs, python has a
ticket. Board is kind of special since it does not have a 1:1 rpc to sdk
wrapper relationship.

The protos pass in the board name and pin name but the sdks create a pin
object depending on what kind of pin you’re requesting then fulfill the
rpcs in the respective objects’ get and set methods.

I believe Alan is working on the python ticket, but I’ve been on vacation
for a while so that might have changed.

randhid:
> I think we can merge and then readdress once all the changes are in

The method has already been changed to AnalogByName. The protos do not match the functions in go and python.

docs/components/board/_index.md
randhid: Specifically this should be changed to match python
https://python.viam.dev/autoapi/viam/components/board/index.html#viam.components.board.Board.analog_by_name

static/include/components/apis/analogreader.md
```diff
@@ -1,4 +1,4 @@
 | Method Name                         | Description                                                             |
 | ----------------------------------- | ----------------------------------------------------------------------- |
-| [`Read`](/components/board/#read)   | Read the current integer value of the digital signal output by the ADC. |
+| [`ReadAnalogReader`](/components/board/#readanalogreader)   | Read the current integer value of the digital signal output by the ADC. |
```
randhid, line 3: This is not a fucntion anywhere in any of our sdks.

docs/components/board/_index.md
randhid: I'm sorry that board kind of breaks the proto -> method consistency because of its design. I'm fine with these being proto methods, but just clarify to the user that you can't get to the over-the-wire message in our python and go skds without invoking the `Analog` object and calling the Read method on it after.

docs/build/micro-rdk/board/_index.md
```diff
@@ -46,7 +46,7 @@ See [PWM signals on `esp32` pins](/build/micro-rdk/board/esp32/#pwm-signals-on-e
 
 For `Analog`s:
 
-- [`Read()`](/components/board/#read)
+- [`ReadAnalogReader()`](/components/board/#readanalogreader)
```
randhid, line 49: This should stay Read, becuase Write will be moved here as well. 
https://pkg.go.dev/go.viam.com/rdk@v0.26.0/components/board#Analog

docs/components/board/_index.md
```diff
@@ -1135,7 +1135,7 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/r
 
 ## `Analog` API
 
-### Read
+### ReadAnalogReader
```
randhid, line 1138: This should probably stay `Read` as well; analog now follows the gpio pin and digital interrupt pin pattern; it is no longer a reader only, but we have grouped Wrtie Analog in with it in the new board interface.

static/include/components/apis/analogreader.md
sguequierre: @npentrel is it ok to change / if proto doesn't match here?

npentrel: I think so @andf-viam tagging you for awareness re: automation

PR #2843, opened 2024-04-25
DOCS-2224: SDK ingest markdownification

sguequierre:
WIP WIP WIP

Need to clear up param types/links not being picked up from dict properly & returns, including usage string returns
Including md for reference (will not merge)

PR #2627, opened 2024-03-07
DOCS-1794: Automate Flutter SDK ingest

andf-viam:
WIP WIP WIP WIP

Add initial WIP version of `get_flutter_methods`.

Current form accomplishes:
- Fetch API proto names
- Fetch Flutter SDK docs and create dictionary(of dictionaries) containing all methods for all type:resource combinations with all their params, returns, links.

TODO:
- Take dictionary of methods and `for method in object, markdownify(method), write(method, file)`
- Troubleshoot some edge cases in fetch Flutter SDK docs methods, missing a few isolated cases in spot checking.

PR #2868, opened 2024-05-02
DOCS-2226: Add override content for autoscraping

andf-viam:
WIP WIP WIP

Add all override content to support DOCS-1794 (#2627) and DOCS-2224 (#2843).

Current State:
- All proto override files included.
	- Originally, these were only intended to supplement existing proto descriptions sourced from upstream (example: adding an additional alert), but we haven't found a consistent source for these. So now these include every proto description in full, eliminating any need to do anything: `write_markdown()` finds these and copies them in line-by-line. Proto descriptions solved.
	- This also makes it very easy for us to change any proto descriptions on our end whenever we want, and then rerun the scraper.
	- Similarly, if we do find a consistent source of upstream proto descriptions, we simply delete these to start using those. Easy.
- A few example method override files included.
	- ANDF to add the rest tomorrow.
	- These will be as needed; most methods do not currently require, so far less effort to assemble.
	- We can easily use these to add code samples that are present in docs.viam.com but aren't yet added to upstream SDKs. We are in the process of doing this presently, so want to be careful about clobbering work. Can always easily delete overrides later to default to scraped content, easy.

NOTED:
- Scraper does not presently differentiate between board sub-APIs, like GPIO, so only presents a single `Close` proto, and does not sort by sub-API. Not sure if OK or needs fixing.


andf-viam:
This PR has been merged into DOCS-2270: https://github.com/viamrobotics/docs/pull/2882.

Closing this PR in favor of that one.

PR #2889, opened 2024-05-07
Update input controller related services

JessamyT:
Probably doesn't make sense to mention frame system here for most users since input controller is for a human to control the base instead of for the motion service to control it.

### DIFF
--- a/docs/components/input-controller/_index.md
+++ b/docs/components/input-controller/_index.md
@@ -34,7 +34,6 @@ Most machines with an input controller need at least the following hardware:
 
 {{< cards >}}
 {{< relatedcard link="/mobility/base-rc/" >}}
-{{< relatedcard link="/mobility/frame-system/" >}}
 {{< /cards >}}
 
 ## Supported models


docs/components/input-controller/_index.md
```diff
@@ -34,7 +34,6 @@ Most machines with an input controller need at least the following hardware:
 
 {{< cards >}}
 {{< relatedcard link="/mobility/base-rc/" >}}
-{{< relatedcard link="/mobility/frame-system/" >}}
```
sguequierre, line 37: Can you think of any other related services that might be good to replace this with?

sguequierre (approved):
LGTM % possible replacement

docs/components/input-controller/_index.md
JessamyT: My thought is no, for the same reason that I don't think frame system should be here--I'm thinking of navigation/motion/frame system as related to each other and useful for automating motion, whereas I think of input control as almost the opposite? Can you think of any?

JessamyT: I don't think any others make sense so I'm going to merge!

PR #2891, opened 2024-05-07
DOCS-2205: Add provisioning diagrams

sguequierre:
Was not 100% sure where to add these on the page so feedback welcome


npentrel (changes requested):
diagrams have typos that we need to fix first

docs/fleet/provision.md
```diff
@@ -138,7 +140,10 @@ The Viam Agent is installed as a `systemd` service named `viam-agent`.
 With the Viam Agent installed, your machine will either connect to a local WiFi network or will create its own WiFi hotspot, depending on your configuration.
 
 - If you include a `viam-server` configuration file on your machine, located at <file>/etc/viam.json</file>, which includes a WiFi network and password to connect to, the Viam Agent will connect to the network automatically when in range.
-- If you did not include this file, or the configured WiFi network is not available when your machine comes online, the Viam Agent will create its own WiFi hotspot.
+- If you did not include this file, or the configured WiFi network is not available when your machine comes online, the Viam Agent will create its own WiFi hotspot and request WiFi credentials:
```
sguequierre, line 143: @ale7714 thought this might be more accurate placement than what I had in first commit, lmk your thoughts? Do agent docs need to be updated at all to account for this as the expected flow?

ale7714: I really like it, it has a good flow with the previous info!

No agent docs update needed.

sguequierre: Sounds good, thank you! @ale7714

npentrel (approved):
I think you're probably good to merge this as is. If Ale has further feedback we can address it in another PR

ale7714 (approved):
It looks great to me 🚀

PR #2899, opened 2024-05-08
Temporarily disable Vale: obscure rule

andf-viam:


npentrel:
Can close - per convo

PR #2900, opened 2024-05-08
Adjust support callout language

npentrel:
In these docs we do not judge 😂


andf-viam (approved):
If for some reason you want to merge this, I agree!! :joy:

LGTM & Thank you!!

PR #2901, opened 2024-05-08
Update robot.md

npentrel:
# Description

### DIFF
--- a/docs/build/program/apis/robot.md
+++ b/docs/build/program/apis/robot.md
@@ -597,6 +597,7 @@ console.log("FrameSytemConfig:", await robot.frameSystemConfig());
 Get the status of the resources on the machine.
 You can provide a list of ResourceNames for which you want statuses.
 If no names are passed in, the status of every available resource on the machine is returned.
+Status includes information relevant to the resource type, for example a servo's status includes its current position.
 
 {{< tabs >}}
 {{% tab name="Python" %}}
--- a/docs/build/program/apis/robot.md
+++ b/docs/build/program/apis/robot.md
@@ -661,7 +661,8 @@ const status = await machine.getStatus();
 
 ### Close
 
-Close the underlying connections and stop any periodic tasks across all constituent parts of the machine.
+Close the underlying client connections to the machine and stops any periodic tasks running in the client.
+
 {{< tabs >}}
 {{% tab name="Python" %}}
 
--- a/docs/build/program/apis/robot.md
+++ b/docs/build/program/apis/robot.md
@@ -596,7 +596,7 @@ console.log("FrameSytemConfig:", await robot.frameSystemConfig());
 
 Get the status of the resources on the machine.
 You can provide a list of ResourceNames for which you want statuses.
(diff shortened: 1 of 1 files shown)


docs/build/program/apis/robot.md
```diff
@@ -596,7 +596,7 @@ console.log("FrameSytemConfig:", await robot.frameSystemConfig());
 
 Get the status of the resources on the machine.
 You can provide a list of ResourceNames for which you want statuses.
-If no names are passed in, the status of every resource configured on the machine is returned.
+If no names are passed in, the status of every available resource on the machine is returned.
```
JessamyT, line 599: Isn't this less clear? What is meant by "available" if that itself is not a status?

docs/build/program/apis/robot.md
```diff
@@ -661,7 +661,8 @@ const status = await machine.getStatus();
 
 ### Close
 
-Close the underlying connections and stop any periodic tasks across all constituent parts of the machine.
+Close the underlying client connections to the machine and stops any periodic tasks running in the client.
```
JessamyT, line 664:
```suggestion
Close the underlying client connections to the machine and stop any periodic tasks running in the client.
```

docs/build/program/apis/robot.md
npentrel: It a resource is configured but does not come online it doesn't get returned

JessamyT: Hmm. So maybe this description is incorrect and it should say it returns information (not statuses) about every available resource on the machine?

npentrel: Why? I don't follow. The status isn't is the resource available on the machine or not. The method returns a list of status structs - Each object in the list contains the resource name, the last time it was reconfigured and its status. So it's a status an object may return.

Here's an example for a servo:
```
status {
  fields {
    key: "position_deg"
    value {
      number_value: 90
    }
  }
}
last_reconfigured {
  seconds: 1715015124
  nanos: 626979004
}
```

JessamyT: Ooh I see. I think of "status" as being either online or offline--we refer to the "machine part status indicator" in the Viam app which I think of as indicating the "status" of the part in that sense. I also looked at the mock robot tutorial for an example when posting my last comment, but I guess that's a bad example. Interestingly, https://python.viam.dev/autoapi/viam/robot/client/index.html#viam.robot.client.RobotClient.get_status does not show a return type at all--is that a bug? 

Is it documented anywhere what the status of each component subtype will include?

npentrel: Yes that's a bug. No that's not documented. It depends how each resource is coded

JessamyT: [opt]
```suggestion
If no names are passed in, the status of every available resource on the machine is returned.
Status includes information relevant to the resource type, for example a servo's status includes its current position.
```

JessamyT (approved):
LGTM % opt suggestion

PR #2902, opened 2024-05-09
DOCS-2300: Restructure generated includes

andf-viam:
Restructure generated include files to be by-resource instead of by-proto, cutting the number of includes down to 29 from 291. 👀 
- Also fixed some minor whitespace issues around code fences and `.after` override files.
- Also updated mapping file with recent board API changes to reader functions.
- Also omitting one more Py method until we can fix its upstream form: `write_analog`

### DIFF
--- a/.github/workflows/update_sdk_methods.py
+++ b/.github/workflows/update_sdk_methods.py
@@ -1158,10 +1158,6 @@ def write_markdown(type, names, methods):
     ## scope limited to 'type', so we don't have to loop by type:
     for resource in names:
 
-        ## Create by-resource directory structure if not already present:
-        #path_to_resource = os.path.join(path_to_generated, resource)
-        #Path(path_to_resource).mkdir(parents=True, exist_ok=True)
-
         ## Determine where to write output:
         filename = resource + '.md'
         full_path_to_file = os.path.join(path_to_generated, filename)
@@ -1170,7 +1166,7 @@ def write_markdown(type, names, methods):
         ## Switch to identify the first method encountered for each resource, to help with
         ## knowing when we are at the top of the include file, or whether to to double newline
         ## between protos:
-        is_first_method_in_this_resource = 1
+        is_first_method_in_this_resource = True
 
         ## Loop through mapping file, and determine which sdk methods to document for each proto:
         with open(proto_map_file, 'r') as f:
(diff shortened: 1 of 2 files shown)


.github/workflows/update_sdk_methods.py
```diff
@@ -1158,8 +1159,18 @@ def write_markdown(type, names, methods):
     for resource in names:
 
         ## Create by-resource directory structure if not already present:
```
npentrel, line 1161:
```suggestion
```

.github/workflows/update_sdk_methods.py
```diff
@@ -1158,8 +1159,18 @@ def write_markdown(type, names, methods):
...
-        path_to_resource = os.path.join(path_to_generated, resource)
-        Path(path_to_resource).mkdir(parents=True, exist_ok=True)
+        #path_to_resource = os.path.join(path_to_generated, resource)
+        #Path(path_to_resource).mkdir(parents=True, exist_ok=True)
```
npentrel, line 1163:
```suggestion
```

.github/workflows/update_sdk_methods.py
```diff
@@ -1589,20 +1599,25 @@ def write_markdown(type, names, methods):
...
 
+                                output_file.write('\n')
+
                             output_file.write("{{% /tab %}}\n")
                             output_file.write("{{< /tabs >}}\n")
 
+                    ## After this loop, we will be working with additional methods appended to the same {resource}.md include file.
+                    ## This switch tells us at the start of the loop for this same resource that we should double-newline the next
+                    ## proto encountered:
+                    is_first_method_in_this_resource = 0
```
npentrel, line 1620: This works but you're kind of using an integer as a boolean here
```suggestion
                    is_first_method_in_this_resource = False
```

.github/workflows/update_sdk_methods.py
```diff
@@ -1158,8 +1159,18 @@ def write_markdown(type, names, methods):
...
+
+        ## Determine where to write output:
+        filename = resource + '.md'
+        full_path_to_file = os.path.join(path_to_generated, filename)
+        output_file = open('%s' % full_path_to_file, "w")
+
+        ## Switch to identify the first method encountered for each resource, to help with
+        ## knowing when we are at the top of the include file, or whether to to double newline
+        ## between protos:
+        is_first_method_in_this_resource = 1
```
npentrel, line 1173:
```suggestion
        is_first_method_in_this_resource = True
```

npentrel (approved):
Nice! Some small nits, otherwise good to go!

.github/workflows/update_sdk_methods.py
andf-viam: You're right, this is cleaner, thanks!

.github/workflows/update_sdk_methods.py
andf-viam: Whoops, thank you!

PR #2903, opened 2024-05-10
Update data _index.md

JessamyT:


docs/data/_index.md
```diff
@@ -19,12 +19,12 @@ menuindent: true
 # SME: Alexa Greenberg
 ---
 
+## Overview
+
```
npentrel, line 23: We don't really do this anywhere else, i.e. having a headline at the top of the page under the h1. I think we should move the following two up to h2s
```suggestion
```

JessamyT: Mm I assumed the overview heading was here by request so figured we should keep it but fix the fact that the top paragraph (which is part of the overview) was divided from the rest of the "overview." I'm fine with eliminating it! Doing now.

npentrel (approved):
thanks!

PR #2909, opened 2024-05-13
DOCS-2302: Improve data type linking

andf-viam:
Improve determination of upstream data type links.
- Should fix all remaining Python SDK data type issues.
- Adds Go SDK data type link and description for `ctx`

To support the above: 
- Reduce reliance on manually populating `python_datatype_links` array, by improving scrape logic.
	- Now much more reliably getting the upstream data type links, no matter where they are defined.
	- Exception: returns with multiple return data types, where at least one is a Viam data type still require manual override via `python_datatype_links` array.
- Add new `link_data_types()` function to help with linking built-in data types, and any override links from `python_datatype_links` array.
	- I think there is still an edge cases where we might match a built-in data type to a multiple data-type return and stop looking through the return for other data types. Haven't tripped it in my testing (I think?) but a future method might.

Change to `write_markdown()`:
- No longer tracking and writing param data type and param data type link separately: Now `parse()` returns a unified `data_type` field that already includes the link(s) (for Python SDK, no changes to Go or Flutter).
- Commenting out some subtype logic for PySDK that are only necessary for Flutter. Leaving in as-commented for now though in case we want to unify code across all three, etc.

Not covered in this PR:
- Go data types other than `ctx`
- Go descriptions
- Flutter anything.

### DIFF
--- a/.github/workflows/update_sdk_methods.py
+++ b/.github/workflows/update_sdk_methods.py
@@ -1342,30 +1342,10 @@ def write_markdown(type, names, methods):
                                     param_data = methods['python'][type][resource][py_method_name]['parameters'][parameter]
 
                                     param_type = param_data.get("param_type")
-                                    #param_subtype = param_data.get("param_subtype")
-                                    #param_type_link = param_data.get("param_type_link")
-                                    #param_subtype_link = param_data.get("param_subtype_link")
                                     param_description = param_data.get("param_description")
                                     optional = param_data.get("optional")
 
                                     output_file.write(f'- `{parameter}` ({param_type})')
-
-                                    # Ideally we could update at least Python SDK with type links?
-                                    #if param_type_link:
-                                    #    # Check for subtype
-                                    #    if param_subtype:
-                                    #        output_file.write(f"({param_type_link})")
-                                    #        if param_subtype_link:
-                                    #            output_file.write(f"<[{param_subtype}]")
-                                    #            output_file.write(f"({param_subtype_link})>")
-                                    #        else:
-                                    #            output_file.write(f"<{param_subtype}>")
-                                    #    else:
-                                    #        output_file.write(f"({param_type_link})")
-                                    # SG: Haven't found any sub-types without param type links-- they are all in flutter SDK--
-                                    # could expand this logic if popped up or grabbing more subtypes?
-                                    #else:
-                                    #    output_file.write('(<INSERT PARAM TYPE LINK>)')
 
                                     if optional:
                                         output_file.write(' (optional)')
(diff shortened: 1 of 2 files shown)


.github/workflows/update_sdk_methods.py
```diff
@@ -445,6 +452,50 @@ def make_soup(url):
...
+## ListOrganizationMembers: Tuple[List[viam.proto.app.OrganizationMember], List[viam.proto.app.OrganizationInvite]
+## DESIGN DECISION: Ignore well-known, usually leading (containing) data types like List, Tuple, Dict.
+## NOTE: Only used in PySDK parsing, for now (but should work for all with minor tweak to support per-language links array):
+def link_data_types(sdk, data_type_string):
+
+    linked_data_type_string = ""
+
+    ## If the passed data_type_string matches exactly to a data type defined in python_datatype_links, use that:
+    if data_type_string in python_datatype_links.keys():
+       linked_data_type_string = '[' + data_type_string + '](' + python_datatype_links[data_type_string] + ')'
```
npentrel, line 467:
```suggestion
       return '[' + data_type_string + '](' + python_datatype_links[data_type_string] + ')'
```

.github/workflows/update_sdk_methods.py
```diff
@@ -445,6 +452,50 @@ def make_soup(url):
...
 
+## Link any matching data types to their reference links, based on {sdk}_datatype_links[] array,
+## used in parse() for both param and return data types. Handles data types syntax that includes
+## multiple data types (and therefore requires multiple data type links), such as
+## ListOrganizationMembers: Tuple[List[viam.proto.app.OrganizationMember], List[viam.proto.app.OrganizationInvite]
+## DESIGN DECISION: Ignore well-known, usually leading (containing) data types like List, Tuple, Dict.
+## NOTE: Only used in PySDK parsing, for now (but should work for all with minor tweak to support per-language links array):
+def link_data_types(sdk, data_type_string):
+
+    linked_data_type_string = ""
```
npentrel, line 463: I don't think this is needed
```suggestion
```

.github/workflows/update_sdk_methods.py
```diff
@@ -445,6 +452,50 @@ def make_soup(url):
...
+                    data_type_linked = '[' + data_type_found + '](' + python_datatype_links[data_type_found] + ')'
+                    linked_data_type_string = regex.sub(data_type_found, data_type_linked, linked_data_type_string)
+                else:
+                    ## If we get here, this data_type is actually a substring of another data type. Take no action:
+                    pass
+
+    ## If we didn't find any matching links, return an empty string so we can know to look elsewhere,
+    ## otherwise return linked data type string:
+    if linked_data_type_string == data_type_string:
+        return ""
```
npentrel, line 494: Wouldn't we want to return something that still produced a readable result and just not have a link? So if the sdk has an int but we don't know the link for it, we'll gave int in our docs just without link?
```suggestion
        return data_type_string
```

potentially we'd want to add logging I suppose

.github/workflows/update_sdk_methods.py
```diff
@@ -1255,30 +1342,30 @@ def write_markdown(type, names, methods):
...
-                                        if param_subtype:
-                                            output_file.write(f"({param_type_link})")
-                                            if param_subtype_link:
-                                                output_file.write(f"<[{param_subtype}]")
-                                                output_file.write(f"({param_subtype_link})>")
-                                            else:
-                                                output_file.write(f"<{param_subtype}>")
-                                        else:
-                                            output_file.write(f"({param_type_link})")
+                                    #if param_type_link:
```
npentrel, line 1354: why comment this out rather than remove? Do you plan to still use it?

.github/workflows/update_sdk_methods.py
npentrel: nvm - I see where it's used now

npentrel (approved):
good to merge once the commented code is cleaned up

PR #2912, opened 2024-05-14
Use TOC bot

npentrel:
With larger files getting included for SDK automation, the TOC that gets generated by Hugo becomes less useful. It essentially can't add things from an include to the TOC. Adding an external library that generates the TOC with js instead. Behaviour should be exactly the same


layouts/partials/head.html
```diff
@@ -34,6 +34,7 @@
 -}}"> {{ template "_internal/opengraph.html" . -}} {{ template
 "_internal/schema.html" . -}} {{ template "_internal/twitter_cards.html" . -}}
 {{ partialCached "head-css.html" . "asdf" -}}
+</script>
```
npentrel, line 37:
```suggestion
```

andf-viam (approved):
This is awesome! Great thinking, did not even think of this (include content not avail for TOC-generation).

This also appears to make it so that clicking a TOC entry _scrolls_ to that location on page smoothly. So clean! Love this!

PR #2913, opened 2024-05-15
Fix toc bug for headers that aren't direct content headers

npentrel:


andf-viam (approved):
Good catch! I .... think I understand what this does and why 😅 Thank you!

PR #2914, opened 2024-05-15
Update dist/main

npentrel:
Bill ran into an issue with this because dist/main didn't exist. We don't have a requirement that it is named that way so might be best not to specify


andf-viam (approved):
Change looks good, this is good feedback from Bill!

To match, should we also update `cli.md` with the same changes? There are six additional instances there. Otherwise LGTM!

PR #2915, opened 2024-05-15
Update cli to use shell background

npentrel:
# Description


andf-viam (approved):
Technically, the `$` (prompt) should only preface the first line of the command, as it's actually a one-liner split over three lines.

But that's definitely too much nit for this, so definitely LGTM here. Thank you!

PR #2922, opened 2024-05-16
DOCS-2257: Document merged Viam agent

skyleilani:
[Corresponding Slack message](https://viaminc.slack.com/archives/C03U2AY30C8/p1714431792017949).


docs/fleet/provision.md
```diff
@@ -23,7 +23,7 @@ The Viam Agent:
 - Automatically connects to a pre-configured WiFi network, or creates its own wireless hotspot if no pre-configured WiFi network is detected.
 - Installs `viam-server` as a static binary, removing the need to perform any library linking or dependency installation during first-time setup.
   You can also use a custom build of `viam-server`, if needed.
-- Provides automatic updates for `viam-server`, the agent itself, and any configured subsystems (such as the Agent Provisioning subsystem).
+- Provides automatic updates for `viam-server`, the agent itself, and any configured functionalities it controls, such as Agent Provisioning.
```
andf-viam, line 26: The sub-components that can be optionally used with the Agent are still called subsystems, both in the raw JSON (i.e.  `agent.viam-server.disable_subsystem: False` ) and on their [GH Readmes](https://github.com/viamrobotics/agent-provisioning)).

I would retain the original phrasing here.

docs/fleet/provision.md
```diff
@@ -189,21 +187,19 @@ If you did not include a `viam-server` configuration file on your machine, or no
...
+        "hotspot_password": "acme123"
         ...
-        "attributes": {
-          "hotspot_password": "acme123"
-          ...
-        }
-      }
-    }
+       }
+   }
```
andf-viam, line 202: nit: make sure these objects are intended to match; I think this should be:

```
...
"agent": {
    "agent-provisioning": {
      ...
      "attributes": {
        "hotspot_password": "acme123"
        ...
      }
    }
}
```

docs/fleet/provision.md
```diff
@@ -189,21 +187,19 @@ If you did not include a `viam-server` configuration file on your machine, or no
...
 }
 ```
```
andf-viam, line 205: There might be some additional fragment-specific changes that could be part of this PR / ticket. I'd clarify with James that:
- I think the alert below, lines 245-249 is no longer true after this change (i.e. users can indeed configure agent settings using a fragment.
- Similarly, the alert at the end of the intro section to https://docs.viam.com/fleet/configure-a-fleet/ is also no longer true: the agent is now configurable via fragments.

I wonder if there is still a limitation whereby some Agent setup steps must still be manually configured (like some Agent config might already need to be in place before using a [provisioning conf file](https://docs.viam.com/fleet/provision/#use-a-provisioning-configuration-file) to add a fragment config of the rest of agent config settings, maybe?). Not sure, James would know!

andf-viam (commented):
Content mostly looks good, just a few small comments. Let's get James' opinion on the changes to fragment configuration that I think are also a part of this ticket. Otherwise JSON changes look good and we're close here!

andf-viam (approved):
LGTM! Let's run it by James to be sure we got everything; great work!

Otterverse (approved):
Changes being made all look correct to me. If there are other places that weren't caught, I'm not sure how I'd find them, as the docs site is pretty huge now. Nothing else shows "agent_config" which was the main actual change for this specific PR.

PR #2932, opened 2024-05-18
DOCS-2011: Document webhooks offline and updated UI

JessamyT:
https://docs-test.viam.dev/2932/build/configure/webhooks/


docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
 For example, you can configure a webhook to send you a notification when your robot's sensor collects a new reading.
 Viam provides two webhook types depending on the event you want to trigger on:
 
-- `"part_data_ingested"`: trigger when a sensor on the machine part collects a new reading
-- `"part_online"`: trigger when the machine part is online
+- Data has been synced to the cloud: trigger when data from the machine is synced
+- Part is online: trigger continuously at specified interval while the {{< glossary_tooltip term_id="part" text="machine part" >}} is online
+- Part is offline: trigger continuously at specified interval while the machine part is offline
```
sguequierre, line 16:
```suggestion
- **Part is offline**: trigger continuously at specified interval while the machine part is offline
```
+ above, suggest as UI element

docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
...
+   {{<imgproc src="/build/configure/webhook-create.png" resize="x400" declaredimensions=true alt="The Create menu with Webhook at the bottom of the list of options." >}}
+
+2. Name the webhook and click **Create**.
+
+3. Select the type of event to trigger on from the **Type** dropdown.
+
+4. Follow the instructions depending on the type of webhook you want to implement:
+
+{{< tabs name="Types of Webhooks" >}}
+{{% tab name="Data synced to cloud" %}}
```
sguequierre, line 35:
```suggestion
{{% tab name="Data has been synced to the cloud" %}}
```
just to match to avoid confusion

docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
...
 
-2. Follow the instructions depending on the type of webhook you want to implement:
+5. Select the types of data you want to trigger on from the dropdown.
 
-{{% tab name="part_data_ingested" %}}
+{{% alert title="Note" color="note" %}}
+Be sure to configure [data capture](/data/capture/) and [cloud sync](/data/cloud-sync/) for the relevant components.
+For example, if you want to trigger a webhook on temperature readings, configure data capture and sync on your temperature sensor.
+Be aware that the component must return the type of data you configured in the webhook's **Data Types**.
```
sguequierre, line 42:
```suggestion
Be aware that the component must return the type of data you configure in the webhook's **Data Types**.
```
I think?
Seems slightly unclear how it matches to components-- will for example "Binary (image)" data trigger from any camera? Maybe could add more info here

docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
...
+
+6. Replace the URL value with the URL of your cloud/lambda function.
+
+   ![The webhook configured with an example URL in the Viam app.](/build/configure/webhook-configured.png)
+
+{{% /tab %}}
+{{% tab name="Raw JSON" %}}
+
+If you prefer to configure your webhook with raw JSON instead of the config builder, you can paste one of the following JSON templates into your JSON config.
+`"webhooks"` is a top-level section like `"components"`, `"services"`, or any of the other config sections.
```
sguequierre, line 68:
```suggestion
`"webhooks"` is a top-level section like `"components"` or `"services"`.
```
Last part just adds confusion I think

docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
...
 
 {{< tabs >}}
-{{% tab name="JSON Template" %}}
+{{% tab name="JSON Template: Data" %}}
```
sguequierre, line 71:
```suggestion
{{% tab name="JSON Template: Data has been synced to the cloud" %}}
```
Maybe clunky but I think they should match, or maybe you could make "type" clear in intro and use that here?

docs/build/configure/webhooks.md
```diff
@@ -11,35 +11,91 @@ Webhooks allow you to trigger actions when certain types of data are sent from y
...
-        },
-        "type": "part_data_ingested"
+        }
+      }
+    }
+  ]
+```
+
+{{% /tab %}}
+{{% tab name="JSON Template: Part Online" %}}
```
sguequierre, line 88:
```suggestion
{{% tab name="JSON Template: Part is online" %}}
```

sguequierre: Is there no template for "Part is offline"?

docs/build/configure/webhooks.md
JessamyT: I was meaning it as conceptual instead of as a UI element, since we're not actually talking about the UI just yet

JessamyT, line 16:
```suggestion
- **Data has been synced to the cloud**: trigger when data from the machine is synced
- **Part is online**: trigger continuously at specified interval while the {{< glossary_tooltip term_id="part" text="machine part" >}} is online
- **Part is offline**: trigger continuously at specified interval while the machine part is offline
```

docs/build/configure/webhooks.md
JessamyT: I don't love super long tab names since on narrow screens I think it looks bad

docs/build/configure/webhooks.md
JessamyT: Okay yeah I think configured or configure works--will change! Yes it will trigger whenever data of that type is synced to the cloud; I don't think it has anything to do with which component it's coming from?

docs/build/configure/webhooks.md
JessamyT: These tab names with the "JSON Template" are already so long I don't want to make them even longer such that they have to wrap onto a new line on narrow screens (which I think makes it look confusing). I'll add this:
```suggestion
{{% tab name="JSON Template: Data Synced" %}}
```

docs/build/configure/webhooks.md
JessamyT: It's the same, so no I think it's overkill

docs/build/configure/webhooks.md
npentrel: maybe just data sync and then more explanation can be in the tab?

docs/build/configure/webhooks.md
sguequierre: I think yeah bc they wouldn't know it's the same. Couldn't hurt too much to have another tab. Or you could point out it's the same (although type line would differ so could be tricky)

sguequierre (approved):

docs/build/configure/webhooks.md
JessamyT: Okay since I added the sentence below I think it's okay as-is

docs/build/configure/webhooks.md
JessamyT: Okay adding a tab!

PR #2942, opened 2024-05-22
DOCS-2249: Document control tab part switch

skyleilani:
I have added screenshots of the control tab populated with a new machine part, as well as new information/context on two fleet pages.

Additionally, the error messages from the subpart do not show up in the mobile app's logs. They only show up on app.viam.com and this information may help users to troubleshoot when using the mobile app. 

Screenshot 2024-05-22 at 1 21 17 PM

IMG_2243

This addition is out of scope, so it's not implemented in this PR.


docs/fleet/machines.md
```diff
@@ -78,10 +78,14 @@ If you are managing a large fleet, you can use {{< glossary_tooltip term_id="fra
 
 Once you have configured components and services for your machine, you can visually test and remotely operate them from the **CONTROL** tab in the [Viam app](https://app.viam.com) or the [Viam mobile app](/fleet/#the-viam-mobile-app).
 
-For more information, see [Control machines](/fleet/control/).
-
 {{<gif webm_src="/fleet/control.webm" mp4_src="/fleet/control.mp4" alt="Using the control tab" max-width="800px">}}
 
+You can also switch between different machine parts by selecting the part from the **Machine part** dropdown menu.
```
andf-viam, line 83: I don't see a **Machine part** drop down menu on the **Configure** tab, that seems to appear only on the **Control** tab. I think you might have the screenshots swapped as well. Otherwise: content looks good, just perhaps swapped places + some minor tweak to language to cover configure tab's machine part switching (I seem to be able to just click on them to select them, is there not actually a "switch" per se for configure like there is for control?

andf-viam (commented):
Almost there! Looks like some of this content might be swapped between configure and control tab, see comments. Otherwise, this is looking good!

andf-viam (approved):
LGTM, thanks much!

PR #2947, opened 2024-05-23
Fix ordering automation

npentrel:
based on the refactoring draft


sguequierre (approved):
I think this is correct, thanks!

PR #2952, opened 2024-05-24
DOCS-2397: Remove FUSE steps from linux install include

sguequierre:
* Removing as duplicated in app 

Figure it's probably ok to keep dependency mention at top + keep troubleshooting documentation in there in case ppl search for error in docs?

### DIFF
--- a/static/include/install/install-linux.md
+++ b/static/include/install/install-linux.md
@@ -3,32 +3,6 @@ The AppImage is a single, self-contained binary that runs on 64-bit Linux system
 
 To install `viam-server` on a Linux computer:
 
-1. Determine if FUSE version 2 is installed on your Linux system:
-
-   ```sh {class="command-line" data-prompt="$"}
-   find /usr -name libfuse.so.2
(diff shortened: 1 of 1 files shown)


JessamyT:
Hmm if this is coming out, maybe we can/should just combine all three OS tabs since there's not really anything specific to each OS besides what's in the setup tab itself.

andf-viam (approved):
I support retaining the tabs, as there is still a little OS-specific content for each: Linux arch + `uname` + AppImage context, macOS brew sourcing, etc.

This LGTM, thank you!

sguequierre:
> Hmm if this is coming out, maybe we can/should just combine all three OS tabs since there's not really anything specific to each OS besides what's in the setup tab itself.
> 
> Edit: Just noticed an issue with the images on the Windows tab and created #2963; I'll close it if you decide to take out the extra tabs!

oops sorry I didn't see this but I do agree with Andrew that we should keep tabs for now-- merging

PR #2954, opened 2024-05-25
Update Slam service returns

randhid:
# Description

We removed the component reference and extra returns from the slam service API. cleaning that up on the main page.


sguequierre (approved):
LGTM thanks Rand!

PR #2960, opened 2024-05-29
DOCS-2344: Add Windows support notice

JessamyT:


static/include/install/windows-support.md
```diff
@@ -0,0 +1,9 @@
+{{% alert title="Windows Support Notice" color="note" %}}
+
+You can install `viam-server` on Windows Subsystem for Linux (WSL) to test your code or machine configuration.
+However, WSL itself does not currently support exposing many types of Windows hardware to the embedded Linux kernel in WSL.
+This means that some hardware, such as a USB webcam connected to your Windows computer, may not be available to `viam-server` when run in WSL, even though it is fully supported for native Linux systems.
+
+You can absolutely [run Viam SDK code](/build/program/run/#run-code-remotely) on your personal computer with WSL; if you install `viam-server` on a single-board computer running Linux, or [remotely borrow a Try Viam rover](/get-started/try-viam/), you can control hardware with code running on your Windows computer.
```
npentrel, line 7:
```suggestion
We recommend you [run Viam SDK code](/build/program/run/#run-code-remotely) on your personal computer with WSL and install `viam-server` on a single-board computer running Linux, or [remotely borrow a Try Viam rover](/get-started/try-viam/). This way you can control hardware with code running on your Windows computer.
```
The absolutely is a bit odd here. What do you think about this?

JessamyT: I hesitate to say we recommend, since we don't know their use case or whether they have an SBC around. We could just as easily recommend that someone boot Linux on their PC and run viam-server there.

I wasn't sure about the "absolutely," but my objective here was to say that though you can't access viam-server's full hardware capabilities on WSL, that has nothing to do with being able to run SDK code on WSL, *if* that's of interest.

docs/build/program/_index.md
```diff
@@ -46,7 +46,8 @@ source .venv/bin/activate
 pip install viam-sdk
 ```
 
-Windows is not supported. If you are using Windows, use the [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install) and install the Python SDK using the preceding instructions for Linux.
+Windows is not supported.
```
npentrel, line 49: should we put this in an alert?

static/include/install/windows-support.md
npentrel: I think we can say that instead:
```suggestion
While you cannot access all hardware if you run `viam-server` on your personal computer with WSL, you can [run code to control a machine](/build/program/run/#run-code-remotely) on your personal computer with WSL if you install `viam-server` on a single-board computer running Linux.
```

docs/build/program/_index.md
JessamyT: I'm inclined to say no, because we already say "run this on a mac or linux computer" higher up in the page, and because there were vague murmurings that it might just work anyway??

static/include/install/windows-support.md
JessamyT: Changing to "although" because "while" can be ambiguous
```suggestion
Although you cannot access all hardware if you run `viam-server` on your personal computer with WSL, you can [run code to control a machine](/build/program/run/#run-code-remotely) on your personal computer with WSL if you install `viam-server` on a single-board computer running Linux.
```

npentrel (approved):

PR #2962, opened 2024-05-29
DOCS-2415: Remove Signaling APIs section content

skyleilani:
Per conversation with Rand:

- ”The board APIs should just be documented in [this section](https://docs.viam.com/build/program/apis/#board).” 

- “The signaling section only has board component apis, which are not signaling apis.”

Rand recommended the Signaling APIs section be deleted entirely for now or left empty.

Move current [Signaling APIs section](https://docs.viam.com/build/program/apis/#signaling-apis) content to [Board API section](https://docs.viam.com/build/program/apis/#board).

[Follow up ticket for documenting Signaling APIs](https://viam.atlassian.net/browse/DOCS-2416)

### DIFF
--- a/docs/build/program/apis/_index.md
+++ b/docs/build/program/apis/_index.md
@@ -198,24 +198,6 @@ Different [vision service](/ml/vision/) models support different methods:
 
 ## Signaling APIs
 
-### GPIO pins
-
-In addition to the [board API](#board), the [board component](/components/board/) supports the following methods for interfacing with GPIO pins on a board:
-
-{{< readfile "/static/include/components/apis/gpiopin.md" >}}
-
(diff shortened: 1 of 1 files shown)


JessamyT (changes requested):
Cool--it makes sense to me to keep this with board. Comments:
1. Shouldn't these subsections be moved up to be within the Board section then instead of just deleted?
2. I think the Signaling APIs heading should be deleted if it's becoming empty
3. It looks like SetGPIO and GetGPIO are duplicated in the current situation--we should put them either in the GPIO interface or the board interface, whichever is correct

npentrel:
As far as I can tell this is already done now. Closing this PR.

PR #2964, opened 2024-05-29
Match CLI page with style guide

JessamyT:
- Capitalize first letters
- "id" -> "ID"
- "Inclusion" column instead of "Required...true/false"

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -259,12 +259,12 @@ viam [global options] command [command options] [arguments...]
 ```
 
 <!-- prettier-ignore -->
-|        parameter     |       description      |
+| Parameter   | Description |
 | ----------- | ----------- |
-| [global options](#global-options)      | _optional_ - list of flags that apply for commands      |
-| [command](#commands)  | _required_ - the specific CLI command to run        |
-| command options   | _required for some commands_  - the operation to run for the specified command.     |
-| arguments   | _required for some commands_ - the arguments for the specified command operation. Some commands take positional arguments, some named arguments.     |
+| [Global options](#global-options)      | _optional_ - list of flags that apply for commands      |
+| [Command](#commands)  | _required_ - the specific CLI command to run        |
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -332,54 +332,54 @@ done
...
 | `filter`     | `add` or `delete` images from a dataset using a filter. See [Using the `filter` argument)](#using-the-filter-argument).|
 | `ids`     | `add` or `delete` images from a dataset by specifying one or more file ids as a comma-separated list. See [Using the `ids` argument)](#using-the-ids-argument).|
-| `--help`      | return help |
+| `--help`      | Return help |
 
 ##### Named arguments
 
 <!-- prettier-ignore -->
-|        argument     |       description | applicable commands | required |
+| Argument     | Description | Applicable commands | Inclusion |
```
andf-viam, line 358: `Inclusion` seems a bit odd to me here: it doesn't really help me to understand what to do with the column contents. You mean: I can find out whether I _must_ include or _may_ include the listed argument. Would it be more direct to state more directly? The user is already likely looking for the words 'optional' or 'required' in this context.

```suggestion
| Argument     | Description | Applicable commands | Required |
```

Or:

```suggestion
| Argument     | Description | Applicable commands | Required / optional |
```

andf-viam (approved):
LGTM, thanks for fixing up the consistency! Only comment I have is regarding the `inclusion` column header. Thank you!

docs/cli.md
JessamyT: I was just following what we do throughout the rest of the docs for consistency, and because I find clearly putting "required" or "optional" as much better than "true" and "false." However, I've always thought "Inclusion" seems a bit awk so maybe we should separately revisit for all the docs??

andf-viam: Oh my goodness, you're right it's everywhere. Nevermind: let's definitely stick with the standard here, and let's consider as a team an alternative for down the line.

This LGTM!

PR #2971, opened 2024-06-03
DOCS-2443: Add resource argument

andf-viam:
Add ability to restrict automation runs by specific resource or list of resources.

LIMITATION:
- Only supports multiple resources if they are all of the same resource type (i.e. all components or all services)

NOTE:
- Requires https://github.com/viamrobotics/docs/pull/2967 in order to use resource arg together with sdk language arg.


sguequierre (approved):
LGTM

PR #2975, opened 2024-06-04
Update Arm MoveToPosition make it clear that Pose parameter is destination pose

sguequierre:
Realized this while making SDK edits


JessamyT (approved):
Good catch!

PR #2984, opened 2024-06-05
DOCS-2462,DOCS-2463: Improve MD table generation

andf-viam:
- Updates MD table generation to conform to [new linting requirements](https://github.com/viamrobotics/docs/pull/2958).
- Strips links from proto descriptions' first sentence when used as MD table proto description.
	- Does not alter links in same proto descriptions (from proto override file) when used on API pages as full descriptions.

### DIFF
--- a/.github/workflows/update_sdk_methods.py
+++ b/.github/workflows/update_sdk_methods.py
@@ -1543,9 +1543,7 @@ def write_markdown(type, names, methods):
                             file_contents = regex.sub(r'\{\{\%.*\%\}\}.*\{\{\% \/[a-b].* \%\}\}', '', file_contents, flags=regex.DOTALL)
                             search_result = file_contents.split('.\n', 1)[0].strip().replace("\n", " ")
 
-                            ## If the proto description contains any MD links, strip them out.
-                            ## We do not wanted links in this table file:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_sdk_methods.py
```diff
@@ -1543,6 +1543,9 @@ def write_markdown(type, names, methods):
                             file_contents = regex.sub(r'\{\{\%.*\%\}\}.*\{\{\% \/[a-b].* \%\}\}', '', file_contents, flags=regex.DOTALL)
                             search_result = file_contents.split('.\n', 1)[0].strip().replace("\n", " ")
 
+                            ## If the proto description contains any MD links, strip them out:
```
npentrel, line 1546: wait so the proto link to the python docs?? Maybe we should remove that instead

npentrel: let's look into it later and merge this for now. I think it's confusing to point to python docs when someone might be looking at another language

npentrel (approved):

.github/workflows/update_sdk_methods.py
andf-viam: Oh! I see what you mean now. Forest for trees moment for me. You're absolutely right, I'll convert any SDK-specific links in proto descriptions to point to docs content instead in a follow-on. Thank you!

PR #2987, opened 2024-06-05
Remove subfolders from components

npentrel:


JessamyT:
Do we need some aliases?

npentrel:
good point, yes

npentrel:
some of them already had it but i've added the rest

JessamyT:
Looks like links still need updating

JessamyT (approved):
LGTM

PR #2999, opened 2024-06-07
Update _index.md fix return on GetImages()

sguequierre:
Realized this was tuple while looking at Python SDK docs. Code sample doesn't need to be changed


andf-viam (approved):
Thanks for catching!

PR #3006, opened 2024-06-10
DOCS-2430: Merge autogen: robot

andf-viam:
Merge `robot` (robot client) generated content.

Noticed:
- Go > Refresh not found upstream in the [usual place](https://pkg.go.dev/go.viam.com/rdk/robot), but found in docs (and [here](https://pkg.go.dev/go.viam.com/rdk/robot/client#RobotClient.Refresh)). Omitting for now until we can determine what non-interface methods should also be included.


static/include/robot/apis/generated/robot.md
```diff
@@ -0,0 +1,760 @@
...
+{{% tab name="Python" %}}
+
+**Parameters:**
+
+- `api_key` ([str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str)) (required): your API key
+- `api_key_id` ([str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str)) (required): your API key ID. Must be a valid UUID
+
+**Returns:**
+
+- ([typing_extensions.Self](https://python.viam.dev/autoapi/viam/robot/client/index.html#viam.robot.client.RobotClient)): the RobotClient.Options
```
npentrel, line 576: This is a little odd. Wonder if we should replace with `viam.robot.client.RobotClient` or just `RobotClient`

static/include/robot/apis/generated/robot.md
```diff
@@ -0,0 +1,760 @@
...
+**Returns:**
+
+- None.
+
+For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/viam/robot/client/index.html#viam.robot.client.RobotClient.log).
+
+{{% /tab %}}
+{{< /tabs >}}
+
+### GetCloudMetadata
```
npentrel, line 518: missing python code example

npentrel (approved):
2 small things - up to you if you want to merge & file tickets, or fix and then merge

static/include/robot/apis/generated/robot.md
andf-viam: There [isn't one](https://python.viam.dev/autoapi/viam/robot/client/index.html#viam.robot.client.RobotClient.get_cloud_metadata). I've added manually from what is live in docs currently.

static/include/robot/apis/generated/robot.md
andf-viam: Guess: `typing_extensions.Self` is relative to calling function, could be more than just `RobotClient` elsewhere. For now, I want to match to only case I've found (for `robot` resource), but expecting to have to walk (some of) this guess back later. For now, this is my best balance: return data type will always be correct (as-scraped), while possibility of incorrect link target exists.

If I complete all autogen merges and this return data type is only used for `RobotClient`, I will indeed replace the data type name everywhere.

PR #3009, opened 2024-06-10
DOCS-2427: Merge autogen: app

andf-viam:
Merge `app` (app client) generated content.

Noticed:
- Required manually changing all instances of `e.g.` and `i.e.` to meet our Vale lint rules.
- `create_organization_invite` ([link](https://python.viam.dev/autoapi/viam/app/app_client/index.html#viam.app.app_client.AppClient.create_organization_invite)) is missing its return section upstream. Will fix separately.


npentrel:
I think we're missing [CreateOrganizationInvite](https://docs.viam.com/appendix/apis/fleet/#createorganizationinvite)

npentrel (approved):
See comment on PR thread

andf-viam:
Thanks 👍 . I've updated with the following:

1. Fixed upstream [[1](https://github.com/viamrobotics/viam-python-sdk/pull/647), [2](https://github.com/viamrobotics/viam-python-sdk/pull/649)], and re-ran with staged latest. It's in now.
2. I have been removing deprecated functions from the mapping file (which completely omits them from the generated API reference. I think we should no longer display these, but if we want to still display deprecated methods in some altered fashion, we can tackle ways to do that afterwards, indeed.
3. Looks like we don't write out PySDK errors raised at all, currently! We get this data in `parse()` already, so I've created https://viam.atlassian.net/browse/DOCS-2506 to write it out in `write_markdown` as a follow-up ticket. Will run for all resources once complete.
4. Fixed the missing data types for string literals, and used param description override system to use the better descriptions as well.

PR #3016, opened 2024-06-12
DOCS-2423: Add generated docs for power_sensor, sensor, servo

npentrel:
## Notes

1. Servo:
- `do_command` and  `close` have no code snippet for go

2. Sensor:
- feels a bit odd that get readings isn't the first in the list
- `get_readings`, `do_command` and  `close` have no code snippet for go

3. Power sensor:
- I do like the frame link on get_geometries - I think it's gone because we removed links from all descriptions because of the python Pose one. I guess we can reconsider later to add a pose page, change everything to the pose links and re-enable


andf-viam:
1. Code samples for both are [awaiting PR review](https://github.com/viamrobotics/rdk/pull/4047) ([here](https://github.com/viamrobotics/rdk/pull/4047/files#diff-164c4b59ac9c99a99ee3e4f8c8a74218de680e980e86189ccb1cb20db8e450b0), specifically). I've been staging those same changes in a local GO SDK staged build when running automation; you can use my branch `[andf-viam:DOCS-2316-fix-comments-data-types](https://github.com/andf-viam/rdk/tree/DOCS-2316-fix-comments-data-types)` upstream to get those samples if you like.
2. You can freely re-order the protos for each resource by simply chaning their order in the [mapping file](https://github.com/viamrobotics/docs/blob/main/.github/workflows/sdk_protos_map.csv)! The order there (top to bottom) is the order the protos will be written to the MD.
3. I only removed links from proto descriptions first sentence when used in the proto list tables (api landing page and per-resource api intro section). For the proto itself, the full description, with all links retained faithfully, is used.

npentrel:
Ok updated

andf-viam (approved):
This LGTM! Thanks for the underscore fix, nice catch!!

PR #3021, opened 2024-06-14
DOCS-2264: Add Copy machine ID button

JessamyT:


docs/cli.md
```diff
@@ -229,9 +229,9 @@ To use a machine part API key to authenticate your CLI session, you must first c
    Where:
 
    - `robot-id` is your machine's ID.
-     You can find your machine ID by running `viam machines list`.
+     You can find your machine ID by running `viam machines list`, or by clicking the **...** in the upper-right corner of your machine's page in the [Viam app](https://app.viam.com).
```
andf-viam, line 232: Nit: Clarity: the ellipsis icon itself does not provide this info:

```suggestion
     You can find your machine ID by running `viam machines list`, or by clicking the **...** button in the upper-right corner of your machine's page in the [Viam app](https://app.viam.com), and selecting **Copy machine ID**.
```

docs/appendix/apis/fleet.md
```diff
@@ -1326,7 +1330,7 @@ List the {{< glossary_tooltip term_id="module" text="modules" >}} under the orga
 
 **Raises:**
 
-- `GRPCError`: This error is raised if an invalid machine ID is passed.
+- `GRPCError`: This error is raised if an invalid org ID is passed.
```
andf-viam, line 1333: Nit: full word, wherever possible:

```suggestion
- `GRPCError`: This error is raised if an invalid organization ID is passed.
```

andf-viam (approved):
LGTM, with two tiny nits. I really like the links to how to find sections that you've added to API methods that use these values - especially because part ID and machine ID could easily be conflated otherwise! Your direct links to / from each help I think to get the user to the correct steps to find the value they need. Nice work!

docs/appendix/apis/fleet.md
JessamyT: Okay will change. Sometimes I think it makes sense to leave as "org ID" because the code here and elsewhere uses a parameter named "org_id"--I don't feel strongly.

andf-viam: I use the following:
- If describing the concept: common, well-known, familiar language: "organization ID"
- If referring to the parameter: code format, explicit, specific: "`org-id`"

If we are worried about users not identifying the two as synonymous, we can do something like "organization ID (as provided by the `org-id` parameter)" etc. Or, an upstream change to use the full name if we truly find that users struggle to identify the two as same.

JessamyT: I don't think we need to add all those words I think it's fine. I just also think that due to the way that the parameters are named (which I don't think needs to be changed), it *could* be said that org ID and organization ID are equally clear in different ways.

PR #3029, opened 2024-06-17
DOCS-2422: Merge autogen: input,motor,mvmt

andf-viam:
Merge `input_controller`, `motor`, and `movement-sensor` generated content.
- Also includes script fixes, encountered when generating this content:
	- Add Accuracy PySDK data type manual linking (unliked [upstream](https://python.viam.dev/autoapi/viam/components/movement_sensor/client/index.html#viam.components.movement_sensor.client.MovementSensorClient.get_accuracy))
	- Fix issue with expanding PySDK relative links to absolute links for some param, return data types
	- Add support for `map[Control]Event`-style data types for override filenames.
	- Add support for retaining formatting in description overrides - previously all newlines were stripped; now you can specify  `<!-- preserve-formatting -->` in your override and all following lines will have their newlines preserved. Good for nesting parameter descriptions for sub data types (like Accuracy)


static/include/components/apis/generated/movement_sensor-table.md
```diff
@@ -0,0 +1,16 @@
+<!-- prettier-ignore -->
+| Method Name | Description |
+| ----------- | ----------- |
+| [`GetLinearVelocity`](/components/movement-sensor/#getlinearvelocity) | Report the current linear velocity in the x, y and z directions (as a 3D vector) in meters per second. |
+| [`GetAngularVelocity`](/components/movement-sensor/#getangularvelocity) | Report the current angular velocity about the x, y and z axes (as a 3D vector) in degrees per second. |
+| [`GetCompassHeading`](/components/movement-sensor/#getcompassheading) | Report the current [compass heading](https://en.wikipedia.org/wiki/Heading_(navigation)) in degrees. |
```
andf-viam, line 6: No idea how a link snuck in here, maybe double `))`?? Anyways, this one to an external resource is fine to leave here for, maybe even helpful. Will investigate later.

npentrel:
- Keep the movement_sensor table as manual - I think that's fine. Probably good to copy over anything that's better in the automatic one though and adjust the order of the generated part to be the same as the manual one or vice versa.
- Motor has some API descriptions where the verb has an `s`. Not merge blocking but file a ticket to fix
- Motor `ResetZeroPosition` has no code sample.
- I do wonder whether the text on return types should be shortened to just what's after the final `.`. For example `viam.components.movement_sensor.Vector3` -> `Vector3`. What do you think? (Also not for this PR ofc)

npentrel (approved):
- Keep the movement_sensor table as manual - I think that's fine. Probably good to copy over anything that's better in the automatic one though and adjust the order of the generated part to be the same as the manual one or vice versa.
- Motor has some API descriptions where the verb has an `s`. Not merge blocking but file a ticket to fix
- Motor `ResetZeroPosition` has no code sample.
- I do wonder whether the text on return types should be shortened to just what's after the final `.`. For example `viam.components.movement_sensor.Vector3` -> `Vector3`. What do you think? (Also not for this PR ofc). Possibly only if it's longer than two parts. I think we do want it in at least some instances to have two

andf-viam:
@npentrel Thank you!

1. I have restored the original `movement_sensor` table include file, and removed it from the script. Now automation will generate the methods file include, but not the table file. Added an HTML comment to the parent page to remind us of this.
2. Updated them all!
3. Found the issue upstream (my typo!) and fixed it here and [there](https://github.com/viamrobotics/rdk/pull/4104).
4. Created https://viam.atlassian.net/browse/DOCS-2530 to tackle this separately, I agree 👍

PR #3037, opened 2024-06-18
DOCS-2512: Merge autogen: camera

andf-viam:
Merge `camera` generated content.
- Also adds new Python class method `from_string` to ongoing list of APIs excluded from the PySDK scrape.


npentrel:
- `GetImage` uses third person form of verb, wheras the other don't.
- python code sample for get image does the standardframe, and the second example does it too? That's a bit odd.

npentrel (approved):
Some small things to either create PRs for or tickets
- GetImage uses third person form of verb, wheras the other don't.
- python code sample for get image does the standard frame, and the second example does it too? That's a bit odd.
- DoCommand has lost Go

andf-viam:
Thanks @npentrel 

1. I found and fixed one instance of plural, but not sure what you mean here.
2. Looks like PySDK added that bit to canonical upstream example, removed dupe in our side copy, and tweaked language a bit.
3. There is no `DoCommand` for `camera` in Go ([here](https://pkg.go.dev/go.viam.com/rdk/components/camera#VideoSource)) (neither explicitly defined, nor inherited).

npentrel:
`GetImage` saying `Returns an image from the camera.` rather than `Return`

andf-viam:
@npentrel Gotcha, thanks! Fixed it in the proto desc override [here](https://github.com/viamrobotics/docs/blob/main/static/include/components/apis/overrides/protos/camera.GetImage.md), but indeed forgot to re-run the script. Will fix separately, shortly. Thank you!

PR #3040, opened 2024-06-18
DOCS-2426: Merge autogen: nav,slam,viz

andf-viam:
Merge `navigation`, `slam`, and `vision` service generated content.
- Fixes code sample scraping bug where method names that start with other method names would be scraped incorrectly. I fixed this before but not good enoughly. This one should sort us out.

Note: This is the first merge that contains Flutter autogen content :tada:


npentrel (approved):
As usual - file tickets or PRs. But ok to merge
- For SLAM`GetPointCloudMap` and `GetInternalState` is missing Golang tab
- For SLAM `GetProperties` should we try to copy the extra parameter info into the python docs?
- For Vision `GetDetections` uses the wrong code sample (`DetectionsFromCamera`), so does `GetClassifications`
- For Vision `Reconfigure` is missing code snippet

andf-viam:
Thanks!

- Both of those methods are not for user use per Jessamy & John N on SLAM, instead they should use the *Full methods, which I wrote separate scraping for specifically because of.
- Let's hold on upstream fixes for this method until eng has finished the "add Properties everywhere" epic, then makes sense to ensure they have best version.
- Fixed but in automation that led to this. Working now 👍 
- No sample upstream. Created https://viam.atlassian.net/browse/DOCS-2600 to invent one.

PR #3042, opened 2024-06-19
Fix bug in update methods script

npentrel:


andf-viam (approved):
Ah, thanks for catching - you're right! LGTM!

PR #3045, opened 2024-06-19
DOCS-2138: Add screenshot

npentrel:


npentrel:
I'm going to err on the side of having people figure it out. Mostly because maintaining what is and isn't hidden by default would be a lot

PR #3060, opened 2024-06-21
Update check-methods.yml

npentrel:
Change the python coverage test to run once weekly and create a jira ticket instead of with every PR


andf-viam (approved):
Nice, LGTM!

PR #3064, opened 2024-06-24
DOCS-2039, DOCS-2435: Add Flutter code samples

andf-viam:
Add support for scraping Flutter code samples to automation.
- This required switching Flutter HTML scrape target from protos content to `viam_sdk` content, which should have been the original target from the beginning - Flutter was the first scrape target I wrote and I've learned a lot since then! Fixed now.
- Many fixes to mapping file now that we are considering actual Flutter methods and not their backing-protos (many of which are not implemented as methods)
- Added proto overrides for `Name` and `FromRobot`; Flutter-only for now but will expand to other SDKs after.

Note:
- Using `markdownify` here for Flutter param and return usage string conversion to markdown. Works super-well, will later convert Go usage parsing to use `markdownify`, having now seen this in action.
- No longer need to run `proto_map()` each script run, so removing it from `run()` except when using mapping mode (`-m`) (rare)

### DIFF
--- a/.github/workflows/sdk_protos_map.csv
+++ b/.github/workflows/sdk_protos_map.csv
@@ -1,17 +1,19 @@
 ## RESOURCE, PROTO, PYTHON METHOD, GO METHOD, FLUTTER METHOD
 
 ## Arm
-arm,GetEndPosition,get_end_position,EndPosition,getEndPosition
+arm,GetEndPosition,get_end_position,EndPosition,endPosition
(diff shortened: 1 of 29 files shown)


JessamyT (approved):
I think this is good! I am not 100% sure of my reviewing credentials here TBH.

andf-viam:
No, https://viam.atlassian.net/browse/DOCS-2595 for that effort. One thing (or at least just 29 things) at a time please.

PR #3077, opened 2024-06-27
DOCS-2581: Add Agent roaming mode

andf-viam:
Add docs for Viam Agent's new roaming mode
- Including changelog entry.

### DIFF
--- a/docs/appendix/changelog.md
+++ b/docs/appendix/changelog.md
@@ -21,7 +21,7 @@ outputs:
 
 {{% changelog date="2024-06-24" color="changed" title="Viam Agent roaming mode" %}}
 
-The Viam Agent now supports roaming mode. You can use the Viam Agent in roaming mode to allow it to attempt to connect to each specified WiFI network in your configuration.
+The Viam Agent now supports roaming mode. You can use the Viam Agent in roaming mode to allow it to attempt to connect to each specified WiFi network in your configuration.
 In roaming mode, the Viam Agent only connects to a WiFi network that has internet access, and will skip a configured network if it is offline.
 
 {{% /changelog %}}
--- a/docs/appendix/changelog.md
+++ b/docs/appendix/changelog.md
@@ -21,7 +21,7 @@ outputs:
 
 {{% changelog date="2024-06-24" color="changed" title="Viam Agent roaming mode" %}}
 
-The [Viam Agent](/fleet/provision) now supports [roaming mode](/fleet/provision#use-roaming-mode). You can use the Viam Agent in roaming mode to allow it to attempt to connect to each specified WiFi network in your configuration.
+The [Viam Agent](/fleet/provision/) now supports [roaming mode](/fleet/provision/#use-roaming-mode). You can use the Viam Agent in roaming mode to allow it to attempt to connect to each specified WiFi network in your configuration.
 In roaming mode, the Viam Agent only connects to a WiFi network that has internet access, and will skip a configured network if it is offline.
 
 {{% /changelog %}}
--- a/docs/appendix/changelog.md
+++ b/docs/appendix/changelog.md
@@ -19,6 +19,13 @@ outputs:
 
 <!-- If there is no concrete date for a change that makes sense, use the end of the month it was released in. -->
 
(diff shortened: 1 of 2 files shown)


andf-viam:
@Otterverse @ale7714 Updated docs for Agent's roaming mode -- LMK if you'd like to see any changes!

Note: IA pass (organization, multiple pages, etc) is not part of this ticket, but will be happening soon!

Thanks!

docs/appendix/changelog.md
```diff
@@ -19,6 +19,13 @@ outputs:
 
 <!-- If there is no concrete date for a change that makes sense, use the end of the month it was released in. -->
 
+{{% changelog date="2024-06-24" color="changed" title="Viam Agent roaming mode" %}}
+
+The [Viam Agent](/fleet/provision/) now supports [roaming mode](/fleet/provision/#use-roaming-mode). You can use the Viam Agent in roaming mode to allow it to attempt to connect to each specified WiFi network in your configuration.
```
Otterverse, line 24: What is now called "Roaming Mode" was previously the default/only mode of operation. The NEW mode is actually also the new default, single-network.

docs/fleet/provision.md
```diff
@@ -164,16 +164,54 @@ This provisioning functionality is provided by the [Viam Agent provisioning subs
 
 If you specify a WiFi network to connect to in your configuration file, the Viam Agent will automatically connect to that network when in range.
```
Otterverse, line 165: This is only true if it's the highest priority network, OR if roaming mode is enabled.

The "networks" section of the config should more correctly be called "additional networks" as the network info provided (via mobile app or web portal) during provisioning is the "primary network" in single network mode. Additional networks from the config are only really useful if roaming_mode is enabled.

docs/fleet/provision.md
```diff
@@ -164,16 +164,54 @@ This provisioning functionality is provided by the [Viam Agent provisioning subs
...
 
-You can configure one or more WiFi networks to connect to in the `agent` configuration object in your `viam-server` configuration file.
-For example, to configure SSIDs and passwords for two WiFi networks named `primaryNet` and `fallbackNet`, you can use the following configuration:
+To configure the Agent with the necessary WiFi connection information and credentials, add an `agent` configuration object to your `viam-server` configuration file that contains this information.
```
Otterverse, line 167: Not "neccessary." Additional networks are entirely optional, (and as noted) only typically useful when roaming mode is also enabled.

docs/fleet/provision.md
```diff
@@ -164,16 +164,54 @@ This provisioning functionality is provided by the [Viam Agent provisioning subs
...
+        },
+      ]
+    }
+  }
+}
+```
+
+You can add this configuration to the <file>/etc/viam.json</file> configuration file you deploy to your machine, or from the **CONFIGURE** tab in the [Viam app](https://app.viam.com/) for your machine, using **Raw JSON** mode.
+
+The Viam Agent will connect to and use the WiFi network you specify even if that network does not have internet access.
```
Otterverse, line 191: This is true ONLY for the "primary" network. Most always, that's the one configured during provisioning, not here in the config. Otherwise, it'll be whatever network has the highest priority. (A network configured via provisioning in single mode will have priority 999.) When not providing priorities, here, they'll default to 0 priority. (Range is -999 to + 999)

Otterverse (changes requested):
I think there's a bit of confusion about the network modes, and the text could be simplified. I'd suggest only mentioning the "networks" section of the config under the roaming mode section. I'm going to describe the two modes as clearly as possible below. Feel free to steal or adapt as needed.

The default (single) mode will connect to the highest priority network—typically configured during provisioning—regardless of global internet connectivity. It will start provisioning mode (hotspot) if that network is not available (or the device/robot configuration file is missing.)

Roaming mode will try all configured networks, both those configured on the device during provisioning (or manually) as well as any additional networks configured via the "networks" attribute section, until it finds one that can provide global internet connectivity. It will start provisioning mode if the internet cannot be reached (or the device/robot configuration file is missing.)

andf-viam:
Thanks, looks like significant IA changes needed after all. Will close this and fold it into the IA restructure later this week / next. Thank you.

PR #3082, opened 2024-06-28
DOCS-2599: Instruct users to set MIME type in data capture UI

JessamyT:
Additionally made some steps numbered to match the rest of that tutorial and for easier user interaction (per feedback on how-to content)


sguequierre (approved):
LGTM

PR #3086, opened 2024-06-28
Remove the module section for the micrordk board

stevebriskin:
Boards are not modular in the micrordk (at least not in a sane, predictable, tested way) so removing that note.


JessamyT (approved):
LGTM

PR #3088, opened 2024-06-29
Fix interrupt pin type and add example

stevebriskin:
The interrupt pin is an int, not a string (https://github.com/viamrobotics/micro-rdk/blob/main/micro-rdk/src/common/digital_interrupt.rs#L5). Plus added an example.
Will probably follow up with an example pr for i2c.


JessamyT (approved):
LGTM thank you!!

PR #3106, opened 2024-07-05
RSDK-7778 - Update motion service docs to include information about bounding regions + GeoGeometries

nfranczak:


docs/services/motion/_index.md
```diff
@@ -511,6 +511,7 @@ Translation in obstacles is not supported by the [navigation service](/services/
...
 - `obstacles` [(Optional[Sequence[GeoGeometry]])](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.GeoGeometry): Obstacles to consider when planning the motion of the component, with each represented as a `GeoGeometry`. <ul><li> Default: `None` </li></ul>
+- `bounding_regions` [(Optional[Sequence[GeoGeometry]])](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.GeoGeometry): Set of bounds which the robot must remain within or in collision with while navigating, with each represented as a `GeoGeometry`. If a user has defined some bounding region(s) and the robot's geometry is neither encompassed by or in collision with the bounding region(s) the call will fail. If a user has defined some bounding region(s) and the selected destination is not within the bounding region(s) the call will fail. Only if both the robot and destination are at least partly contained by the defined bounding region(s) will the call execute as normal. If we have no defined bounding regions the call will execute as normal. <ul><li> Default: `None` </li></ul>
```
npentrel, line 514:
```suggestion
- `bounding_regions` [(Optional[Sequence[GeoGeometry]])](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.GeoGeometry): Set of bounds which the robot must remain within or in collision with while navigating, with each represented as a `GeoGeometry`. The call will only execture normally if both the robot and destination are at least partly contained by the defined bounding regions or if no bounding regions are defined. If one or more bounding regions are defined and the robot's geometry is neither encompassed by or in collision with any bounding regions the call will fail. If any bounding regions are defined and the selected destination is not within the bounding regions the call will fail. <ul><li> Default: `None` </li></ul>
```

npentrel: Starting with the positive case

PR #3107, opened 2024-07-05
DOCS-2564: Add Grafana to viz howto

andf-viam:
Add Grafana tab to visualize data how-to guide, and convert existing steps to be all-others in separate tab.
- Opted for simple (how-to), so omitting example forms of various URIs and connection strings. Can re-add (from `docs/tutorials/services/visualize-data-grafana.md` if helpful)

NOTICED: `tablestep` and `tabs` seem to require a specific order, perhaps tabs with tablesteps within them are not permitted, but tablesteps with tabs in them is? If so I can try switching around, but more than one tablestep is specific to the grafana tab. I could see a solution where I declare the tabset in each tablestep, if there is no way to include tablesteps within tabs. Hoping I just did something wrong.


npentrel:
Annoyingly I still see it on staging...

npentrel (approved):
LGTM % confirmation that formatting issue is gone

andf-viam:
@npentrel Thanks so much for the help!

PR #3110, opened 2024-07-08
RSDK-8119 - delete velodyne builtin docs

seanavery:
We are removing the builtin velodyne camera component -- this PR simply removes refs from the documentation.


seanavery:
@npentrel Any tips on removing `velodyne` option return on the supported models section? It is still showing up after removing the component page. 
https://docs.viam.com/components/camera/

npentrel:
I'm not at my computer you'd need to delete the velodyne.md file and then upon deploy it should be gone. I can do it tmrw morning

npentrel (approved):
Thank you!

npentrel:
@seanavery Velodyne is now gone - thanks!

PR #3111, opened 2024-07-08
DOCS-2559:Add new headers to triggers

sguequierre:
* Demonstrate this in docs (not sure if we should start doing more to document?)


npentrel (approved):
I think there's another in flight PR that should do more documentation, so let's merge this as is and revisit

PR #3124, opened 2024-07-10
DOCS-2063: Add GetProperties API

andf-viam:
Add `GetProperties` Go and PySDK API methods. Flutter not implemented upstream yet.
- Add new data type to Python link array for `Properties`

### DIFF
--- a/static/include/services/apis/generated/vision-table.md
+++ b/static/include/services/apis/generated/vision-table.md
@@ -10,4 +10,5 @@
 | [`DoCommand`](/services/vision/#docommand) | Execute model-specific commands that are not otherwise defined by the service API. |
 | [`FromRobot`](/services/vision/#fromrobot) | Get the resource from the provided robot with the given name. |
 | [`Name`](/services/vision/#name) | Get the `ResourceName` for this instance of the Vision service with the given name. |
+| [`GetProperties`](/services/vision/#getproperties) | Get info about what vision methods the vision service provides. |
 | [`Close`](/services/vision/#close) | Safely shut down the resource and prevent further use. |
--- a/static/include/services/apis/generated/vision-table.md
+++ b/static/include/services/apis/generated/vision-table.md
@@ -10,5 +10,5 @@
 | [`DoCommand`](/services/vision/#docommand) | Execute model-specific commands that are not otherwise defined by the service API. |
 | [`FromRobot`](/services/vision/#fromrobot) | Get the resource from the provided robot with the given name. |
 | [`Name`](/services/vision/#name) | Get the `ResourceName` for this instance of the Vision service with the given name. |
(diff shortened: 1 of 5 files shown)


static/include/services/apis/generated/vision.md
```diff
@@ -615,6 +615,52 @@ For more information, see the [Flutter SDK Docs](https://flutter.viam.dev/viam_s
 {{% /tab %}}
 {{< /tabs >}}
 
+### GetProperties
+
+Fetch information about what vision methods the vision service provides.
```
JessamyT, line 620: I think this description is a bit confusing as it could be understood to mean find out which methods the vision service in general provides. What do you think about this or similar?
```suggestion
Fetch information about which vision methods a given vision service supports.
```

static/include/services/apis/generated/vision-table.md
```diff
@@ -10,4 +10,5 @@
 | [`DoCommand`](/services/vision/#docommand) | Execute model-specific commands that are not otherwise defined by the service API. |
 | [`FromRobot`](/services/vision/#fromrobot) | Get the resource from the provided robot with the given name. |
 | [`Name`](/services/vision/#name) | Get the `ResourceName` for this instance of the Vision service with the given name. |
+| [`GetProperties`](/services/vision/#getproperties) | Fetch information about what vision methods the vision service provides. |
```
JessamyT, line 13: See comment below

static/include/services/apis/overrides/protos/vision.GetProperties.md
```diff
@@ -0,0 +1 @@
+Fetch information about what vision methods the vision service provides.
```
JessamyT, line 1: See comment above; sorry I should've made that comment here instead of there since this is where we manually change it

JessamyT (approved):
LGTM % clarify description

npentrel (approved):
Lgtm % pr to upstream. Can merge as is and we'll get the changes as they merge

JessamyT:
Isn't this a change that wouldn't be done upstream since this override description isn't coming from the SDK repos?

npentrel:
> Isn't this a change that wouldn't be done upstream since this override description isn't coming from the SDK repos?

Yeah true - I did not notice that

static/include/services/apis/overrides/protos/vision.GetProperties.md
npentrel, line 1: to make this easier :)
```suggestion
Fetch information about which vision methods a given vision service supports.
```

static/include/services/apis/generated/vision-table.md
andf-viam, line 13:
```suggestion
| [`GetProperties`](/services/vision/#getproperties) | Fetch information about which vision methods a given vision service supports. |
```

PR #3125, opened 2024-07-10
DOCS-2651: Fix flutter datatype rendering

andf-viam:
Fix Flutter data type render bug whereby `>>` is rendered as `>>` but smushed. Hugo does that. We now escape earlier in the chain of tech-to-tech-to-tech to prevent it.
- Reran all auto with fix in place to pick up new syntax for gen content
- Also fixes a bunch of Py links, earlier fixed and never rerun, or recently changed upstream. Don't know.
- Also adds latest app client changes for fragment methods. We probably have a ticket for it somewhere.
- Looks like PySDK redid their DoCommand code sample.


JessamyT (approved):
LGTM

npentrel (approved):
Nice to see it catching updates 😊

PR #3126, opened 2024-07-11
DOCS-2349: Add CaptureAllFromCamera API

andf-viam:
Add `CaptureAllFromCamera` API.
- Only for PySDK and Go, not implemented in Flutter.
- Looks like PySDK has fixed / moved more data type links upstream.
- I guessed at whatever Go param `opts` and return `VisCapture` type mean. Former total guess, latter from the PySDK return. Feel free to update upstream or create a follow-up ticket if you would like to tell Eng what to put here instead!
- No code sample for Go. Please feel free to create a ticket to add one upstream if you'd like to see one!


npentrel (approved):
Yes - we do need a follow up ticket here to fix the upstream Go docs - that is our responsibility. This is ok to merge and I will create the ticket but usually that is the ticket doer's responsibility

PR #3161, opened 2024-07-28
DOCS-2641: Change more info icon

npentrel:
I've tried making it say "More info" and it doesn't look great. So I'm trying a question mark circle instead. Let me know what you think.

### DIFF
--- a/layouts/shortcodes/tablestep.html
+++ b/layouts/shortcodes/tablestep.html
@@ -1,7 +1,7 @@
 {{- $link := (.Get "link") -}}
 <div class="tablestep">
   {{ if $link }}
-  <a href="{{ $link }}" class="tablesteplink" title="Click for more information on the topic."><i class="fa fa-info-circle"></i></a>
+  <a href="{{ $link }}" class="tablesteplink" title="Click for more information on the topic."><i class="fa fa-question-circle"></i></a>
   {{ end }}
   {{ with $.Inner }}
       {{ if eq $.Page.File.Ext "md" }}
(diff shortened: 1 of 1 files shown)


JessamyT (approved):
I don't think it makes that much of a difference since it still looks like a tooltip instead of a link? So I think we should still iterate

npentrel:
If you don't think it's an improvement we don't have to do this

JessamyT:
Okay! I just meant it's fine to merge if you think it's an improvement since I don't think it makes it worse :D Unfortunately I don't think this will alleviate the feedback though so whether or not you merge this, let's keep exploring options maybe with Cristina

layouts/shortcodes/tablestep.html
```diff
@@ -1,7 +1,7 @@
 {{- $link := (.Get "link") -}}
 <div class="tablestep">
   {{ if $link }}
-  <a href="{{ $link }}" class="tablesteplink" title="Click for more information on the topic."><i class="fa fa-info-circle"></i></a>
+  <a href="{{ $link }}" class="tablesteplink" title="Click for more information on the topic."><i class="fa fa-question-circle"></i></a>
```
npentrel, line 4:
```suggestion
  <a href="{{ $link }}" class="tablesteplink" title="Click for more information on the topic."><i class="fas fa-file-alt"></i></a>
```

npentrel:
Ok I think there's a file icon which I like for this. If you want to look at other availble icons, look for font awesome version 5.15.4 and see what's available in the free icons

PR #3172, opened 2024-07-31
Draft what is Viam

JessamyT:


npentrel (commented):
I like the info the **What is Viam?** section highlights which are good points about the platform. However, I'm confused how we ended up with no examples in this draft. Maybe you were still going to do that? 
Let's sync today on this.

PR #3187, opened 2024-08-05
DOCS-2643: Address Emily H.'s feedback on person detection quickstart

sguequierre:


docs/get-started/quickstarts/detect-people.md
```diff
@@ -42,7 +42,10 @@ Click the name of a machine to go to that machine's page, where you'll find a va
 {{% /expand%}}
 {{%expand "Step 2: Install viam-server" %}}
 
-Follow the {{< glossary_tooltip term_id="setup" text="setup instructions" >}} that appear on your new machine's **CONFIGURE** page to install `viam-server` on your computer and connect it to the Viam app.
+Follow the {{< glossary_tooltip term_id="setup" text="setup instructions" >}} that appear on your new machine's **CONFIGURE** page to install `viam-server` on your computer, start running it and connect it to the Viam app.
+Select the OS that's appropriate for your computer.
+For example, **Mac** for a MacBook.
+Select **RDK** as the RDK type.
```
npentrel, line 48: Do we think that's necessary? I had just removed a lot of these because it feels like we're stating the obvious...

docs/get-started/quickstarts/detect-people.md
```diff
@@ -61,25 +64,30 @@ For more detailed information, including optional attribute configuration, see t
 {{% /expand%}}
 {{%expand "Step 4: Deploy a person detection model" %}}
 
-Now add an `ML model` service:
+Now add an `ML model` service that can detect a person:
 Click **+**, click **Service** and select the `ML model` type, then select the `TFLite CPU` model.
+Create the service.
```
npentrel, line 69:
```suggestion
```

ditto here, I think that should be clear from context, or maybe it should say "Now create an `ML model` service that can detect a person"?

npentrel (commented):
A few small things

docs/get-started/quickstarts/detect-people.md
sguequierre: I see what you mean but you do have to select create so I thought this was still helpful

docs/get-started/quickstarts/detect-people.md
sguequierre: I don't know if people will know the RDK thing if they're brand new users. Can remove the "for example" maybe?

npentrel (approved):

PR #3201, opened 2024-08-06
DOCS-2550: Add from one machine to many how-to

JessamyT:
Icons are stand-ins; could replace with more screenshots, and I'm seeing if Cristina has something similar if we want icons not screenshots

https://docs-test.viam.dev/3201/use-cases/one-to-many/

PR #3215, opened 2024-08-07
DOCS-1915: Revert changes for fragments

sguequierre:
* Revert changes for fragments

Couldn't get this to work on my own, asked question to resolve but for now going off of diagram demo


docs/fleet/fragments.md
```diff
@@ -89,12 +89,18 @@ The configuration cards of components and services included in the fragment are
 
 Likewise, when you modify the fragment itself, those changes are pushed to all machines that use that fragment.
 
-If you need to modify the config of just one machine that uses a fragment, you have two options:
+If you need to modify the config of just one machine that uses a fragment, you have three options:
 
+- Revert changes made from the fragment overwrite
```
sguequierre, line 94:
```suggestion
- Revert changes made from the fragment overwrite.
```

JessamyT (changes requested):
So, I don't think this is accurate. I think what the eng change is, is if you make fragment mods, you can revert them back to the original fragment (i.e. remove the fragment mods).

Rather than adding revert fragment mods as a separate option for overwriting a fragment, it'd be more like, once you use fragment mods using the builder UI (which this page also needs to be updated with), you can *then* revert the mods you just made.

I think it'd make more sense to update the fragments doc with the new easy (builder) way of mod-ing fragments first, before talking about reverting.

sguequierre:
> So, I don't think this is accurate. I think what the eng change is, is if you make fragment mods, you can revert them back to the original fragment (i.e. remove the fragment mods).
> 
> Rather than adding revert fragment mods as a separate option for overwriting a fragment, it'd be more like, once you use fragment mods using the builder UI (which this page also needs to be updated with), you can _then_ revert the mods you just made.
> 
> I think it'd make more sense to update the fragments doc with the new easy (builder) way of mod-ing fragments first, before talking about reverting.
> 
> image
> 
> Edit: I'm working on related content right now so if you'd like me to roll this into that lmk!

Ok, thank you, this makes sense with the confusion I had! I will update this to clarify.

JessamyT:
So sorry if you've already put time into updating this but, I think it is actually covered by https://github.com/viamrobotics/docs/pull/3201. When I went to de-duplicate content I realized it made more sense to just update in both the how-to and the existing page.

sguequierre:
> So sorry if you've already put time into updating this but, I think it is actually covered by #3201. When I went to de-duplicate content I realized it made more sense to just update in both the how-to and the existing page.

No I haven't yet, so thank you for letting me know-- I can close this if that's ok! Thank you for your work on this!

PR #3224, opened 2024-08-09
DOCS-2160: Update xarm screenshots for correct attribute requirements

sguequierre:
* more was marked as required than necessary


npentrel (approved):
as in the UI is now correct or still wrong?

sguequierre:
> as in the UI is now correct or still wrong?

UI is correct now

PR #3250, opened 2024-08-13
DOCS-2670: Make new model training gif

sguequierre:
* train pac man ghosts dataset from viam devrel demos
* goes through whole process with auto-labeling


static/services/ml/train-model.gif
sguequierre, line 1: this gif should actually be longer/show the page, I think preview is just slightly off. unsure how to fix

JessamyT (approved):
Hmm in the rendered site the GIF is stretched horizontally a little bit. Not sure what's going on there. Mind following up after merging to see if it's still doing that in prod, and if it is, try messing with the formatting?

static/services/ml/train-model.gif
npentrel: I assume you used the command to make the preview gif, which just takes the first few seconds of a video and makes a gif. Use [Kap](https://getkap.co/) or [ezgi](https://ezgif.com/) to manually make a gif that is suitable.

npentrel: Please also 
- reduce screen width so the video & gifs aren't needlessly wide
- remove needless cursor movement - re-record or cut it
- speed up the video, people can take in information faster than the recording, so generally good to increase speed to at least 150%

npentrel (changes requested):
see comments.

npentrel (approved):
Thanks looks great!

PR #3257, opened 2024-08-15
Include note about installing ffmpeg 

randhid:
Add a note to the ffmpeg camera page to tell users they need to have ffmpeg installed on their system to use this camera model. MacOS users get ffmpeg through viam-server's brew installation, but that does not apply to linux users depending on the app image they install. We are not adding an ffmpeg build dependency to viam-server. My hope is that this could be a module one day and handle its own dependencies, but until then, we add this note.


npentrel (approved):

npentrel:
Thanks!

PR #3269, opened 2024-08-18
DOCS-2822: Add provisioning guides

npentrel:
I've added two guides:

- [Prepare provisioning](https://docs-test.viam.dev/3269/how-tos/provision/)
- [Set a machine up](https://docs-test.viam.dev/3269/how-tos/provision-setup/)

### DIFF
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -115,6 +115,14 @@ sudo ./preinstall.sh
 Follow the instructions.
 If you created a <FILE>viam-provisioning.json</FILE>, specify its location when prompted.
 
+{{% expand "Using a Raspberry Pi?" %}}
+
+For Raspberry Pis, the script will automatically perform the required next steps, it will:
+
+- create a tarball
+- update `firstrun.sh`.
+- extract the tarball to the mounted root filesystem
+
 ```sh {class="command-line" data-prompt="$" data-output="2-40"}
 sudo ./preinstall.sh
 
@@ -149,7 +157,12 @@ a usr/local/lib/systemd/system/multi-user.target.wants/viam-agent.service
 Install complete! You can eject/unmount and boot the image now.
 ```
 
-If you get an error, run the script for the target system's architecture:
+{{% /expand%}}
+
+{{% expand "Error: no valid image found at mountpoints (or manually provided path)" %}}
+
+If you get this error, you can run the script for the target system's architecture.
+It will create a tarball for the system's architecture which you will then need to manually extract.
 
 {{< tabs >}}
 {{% tab name="arm64" %}}
@@ -168,11 +181,23 @@ sudo ./preinstall.sh --x86_64
 {{% /tab %}}
 {{< /tabs >}}
 
-If this command fails because the script cannot detect your mountpoint, you can specify it directly:
+To extract the tarball, run:
+
+```sh {class="command-line" data-prompt="$"}
+sudo tar -xJvpf $TARBALL -C <PATH_TO_ROOT_FS>
+```
+
+{{% /expand%}}
+
+{{% expand "Refusing to install to unknown/unset ROOTFS" %}}
+
+If your root file system cannot be detected, you can specify it directly:
 
 ```sh {class="command-line" data-prompt="$"}
 sudo ./preinstall.sh /path/to/rootfs
 ```
+
+{{% /expand%}}
 
 {{< /tablestep >}}
 {{< /table >}}
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -43,7 +43,7 @@ If you choose to have a mobile app experience, you can use the [Viam mobile app]
 If you choose to use the Viam mobile app, you must provide a {{< glossary_tooltip term_id="fragment" text="fragment" >}} for provisioning.
 If you do not yet have a fragment, follow the steps to [Create a configuration fragment](/how-tos/one-to-many/) and make a note of the fragment ID.
 
-If you choose to use the captive web portal, you can optionally create a machine already and provide its credentials configuration file at <FILE>/etc/viam.json</FILE>.
+If you choose to use the captive web portal, you can optionally create a machine in advance and provide its credentials configuration file at <FILE>/etc/viam.json</FILE>.
 
 You can get the `viam-server` machine credentials by clicking the copy icon next to **Viam server config** in the part status dropdown to the right of your machine's name on the top of the page.
 
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -1,0 +1,205 @@
+---
+title: "Complete setup for a machine"
+linkTitle: "Set a machine up"
+weight: 69
+type: "docs"
+description: "If you have received a machine that uses Viam and have been pointed to this guide, this guide will show you how to set it up."
+images: ["/platform/provisioning-demo.gif"]
+videos: ["/platform/provisioning-demo.webm", "/platform/provisioning-demo.mp4"]
+---
+
+If you have received a machine with Viam pre-installed on it, this guide will show you how to complete your device setup.
+
+{{< alert title="In this page" color="tip" >}}
+
+1. [Set up your machine using the Viam mobile app](#set-up-your-machine-using-the-viam-mobile-app).
+1. [Set up your machine using the captive portal](#set-up-your-machine-using-the-captive-portal).
+1. [Update machine configuration](#update-machine-configuration) to update hotspot details or add additional networks.
+   {{< /alert >}}
+
+## Prerequisites
+
+- Physical hardware constituting a machine
+- A laptop or mobile device if using the Viam mobile app
+
+## Set up your machine using the Viam mobile app
+
+{{<video webm_src="/platform/provisioning-demo.webm" mp4_src="/platform/provisioning-demo.mp4" alt="Using the Viam mobile app to provision a new machine with viam-agent." poster="/platform/provisioning-demo.jpg" max-width="400px" class="aligncenter imgzoom">}}
+
+{{< table >}}
+{{% tablestep %}}
+**1. Install the Viam mobile app**
+
+You can find the mobile app on the [App Store](https://apps.apple.com/vn/app/viam-robotics/id6451424162) and on [Google Play](https://play.google.com/store/apps/details?id=com.viam.viammobile&hl=en&gl=US).
+
+<a href="https://apps.apple.com/vn/app/viam-robotics/id6451424162" target="_blank">
+  <img src="https://github.com/viamrobotics/docs/assets/90707162/a470b65d-1b97-412f-9f97-daf902f2f053" width="200px" alt="apple store icon" class="center-if-small" >
+</a>
+
+<a href="https://play.google.com/store/apps/details?id=com.viam.viammobile&hl=en&gl=US" target="_blank">
+  <img src="https://github.com/viamrobotics/docs/assets/90707162/6ebd6960-08c5-41d4-81f9-42293fbfdfd4" width="200px" alt="google play store icon" class="center-if-small" >
+</a>
+
+{{% /tablestep %}}
+{{% tablestep %}}
+**2. Create a machine**
+
+Open the Viam mobile app and sign in.
+Then, select an organization and location for your machine.
+If you have already created a machine, select it.
+If you have not yet created a machine, click on **Add new smart machine** and give your machine a name.
+{{% /tablestep %}}
+{{% tablestep %}}
+**3. Connect to your machine's WiFi hotspot**
+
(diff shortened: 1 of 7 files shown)


docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+a usr/local
+a usr/local/lib
+a usr/local/lib/systemd
+a usr/local/lib/systemd/system
+a usr/local/lib/systemd/system/viam-agent.service
+a usr/local/lib/systemd/system/multi-user.target.wants
+a usr/local/lib/systemd/system/multi-user.target.wants/viam-agent.service
+
+
+Install complete! You can eject/unmount and boot the image now.
```
npentrel, line 203: If I turn the machine on, it doesn't start the hotspot.

ale7714: do you have machine credentials  in the machine?

npentrel: no, neither. I thought if I don't configure internet then it should go into hotspot mode, and that should then request machine credentials

npentrel: @Otterverse is it required that I have a working network on the device already as well as a `viam.json`? I thought the whole point was that if I didn't a hotspot would start... I guess I'll continue testing tomorrow. If we have steps for this written down anywhere else, please let me know.

Otterverse: You're correct. viam-provisioning.json is ALWAYS optional (think of it as "branding/customization" for the manufacturer) whereas viam.json is the machine credentails, and that IS exactly what provisioning is meant to help set up. So it's absolutely valid for it to be missing if you want to have a machine do provisioning at first boot.

If you pre-installed the agent on the SD card, then it should absolutely start up a hotspot and allow you to provide both network and machine credentials from there. It can take several minutes for first boot, but if after 5+ minutes nothing is happening, then something may have gone wrong. Feel free to slack me if you want some help troubleshooting.

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,205 @@
...
+
+1. [Set up your machine using the Viam mobile app](#set-up-your-machine-using-the-viam-mobile-app).
+1. [Set up your machine using the captive portal](#set-up-your-machine-using-the-captive-portal).
+1. [Update machine configuration](#update-machine-configuration) to update hotspot details or add additional networks.
+   {{< /alert >}}
+
+## Prerequisites
+
+- Physical hardware constituting a machine
+- A laptop or mobile device if using the Viam mobile app
```
Otterverse, line 23: Should mention it must have wifi. Desktops with wifi work fine too.

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Set up your machine using the Viam mobile app
```
Otterverse, line 25: Should mention this is an either/or with the Captive Portal method below. Otherwise this reads like "step number 2 of 3" in the overall page.

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,205 @@
...
+{{% tablestep %}}
+**2. Create a machine**
+
+Open the Viam mobile app and sign in.
+Then, select an organization and location for your machine.
+If you have already created a machine, select it.
+If you have not yet created a machine, click on **Add new smart machine** and give your machine a name.
+{{% /tablestep %}}
+{{% tablestep %}}
+**3. Connect to your machine's WiFi hotspot**
```
Otterverse, line 53: I'm not sure this should be documented here. The mobile app will tell users when to switch networks, and they should follow the instructions there instead I think.

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,205 @@
...
+          "priority": 10
+        }
+      ]
+    }
+  }
+}
+```
+
+{{% /tablestep %}}
+{{< /table >}}
```
Otterverse, line 196: I don't think these are needed here. Most users don't need to do these steps. Perhaps just mention them and link the existing section for these configuration options.

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
+---
+title: "Configure provisioning with viam-agent"
+linkTitle: "Prepare provisioning"
```
Otterverse, line 3: These names don't make sense to me, and are too easily confused with the other "provision-setup" page (if anything, the titles should be switch. THIS is the setup for provisioning, the other is using provisioning as an end-user.)

These instructions are for "pre-installation of Viam on a manufacturing (or other) image" So I'd suggest a name something like that. At least the phrase "pre-install" belongs, IMHO.

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Prerequisites
+
+{{% expand "One or more physical devices with supported operating system" %}}
+
+To find out more about supported systems, see [Compatibility](/installation/#compatibility).
+
+{{% /expand%}}
+
+{{% expand "A machine fragment" %}}
```
Otterverse, line 33: This is only required when using the mobile app. It's optional for the captive portal method (the portal will only display it, not do anything with it.)

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Install `viam-agent`
+
+`viam-agent` is a self-updating service manager that maintains the lifecycle for several Viam services and keeps them updated.
+If you intend to use `viam-agent` to keep your device's Viam software up-to-date or to use its provisioning plugin to let end users set up their own machines, you need to use `viam-agent` and install it on the machine before sending it to the user.
+
+You can [install `viam-agent` on live systems](/installation/) or preinstall it:
+
+{{< tabs >}}
+{{% tab name="Preinstall on an image" %}}
```
Otterverse, line 87: These instructions are (partially) duplicating the SD card tab. The preinstall script can install to a mounted FS image (that's typically an SD card) OR it can create a tarball (which is the second half of these instructions.)

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+
+```sh {class="command-line" data-prompt="$"}
+sudo ./preinstall.sh /path/to/rootfs
+```
+
+{{< /tablestep >}}
+{{< /table >}}
+
+{{% /tab %}}
+{{% tab name="Preinstall on an SD card" %}}
```
Otterverse, line 152: This should be the primary tab/option as well.
```suggestion
{{% tab name="Preinstall on an SD card (or other image)" %}}
```

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+Eject and reinsert the card to make sure it's mounted with the newly written contents. Then run the following script:
+
+```sh {class="command-line" data-prompt="$"}
+sudo /bin/sh -c "$(curl -fsSL https://storage.googleapis.com/packages.viam.com/apps/viam-agent/preinstall.sh)"
+```
+
+Follow the instructions:
+
+```sh {class="command-line" data-prompt="$" data-output="2-40"}
+coding sudo /bin/sh -c "$(curl -fsSL https://storage.googleapis.com/packages.viam.com/apps/viam-agent/preinstall.sh)"
```
Otterverse, line 173:
```suggestion
sudo /bin/sh -c "$(curl -fsSL https://storage.googleapis.com/packages.viam.com/apps/viam-agent/preinstall.sh)"
```

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+**2. Run the script**
+
+Run the preinstall script without options and it will attempt to auto-detect a mounted root filesystem (or for Raspberry Pi, bootfs) and also automatically determine the architecture.
+
+```sh {class="command-line" data-prompt="$"}
+sudo ./preinstall.sh
+```
+
+If this succeeds, continue to step 5.
+Alternatively, go to step 3.
```
Otterverse, line 117: This doesn't belong on this tab.

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+sudo ./preinstall.sh --x86_64
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+If this command fails because the script cannot detect your mountpoint, you can specify it directly:
+
+```sh {class="command-line" data-prompt="$"}
+sudo ./preinstall.sh /path/to/rootfs
```
Otterverse, line 145: This is the backup option for the "SD Card" tab... if it can't find the SD card mountpoint. It likely belongs there if this tab becomes "Create tarball"

docs/how-tos/provision.md
```diff
@@ -0,0 +1,310 @@
...
+
+{{< alert title="Note" color="note" >}}
+Use relative symlinks, especially if working on a mounted image (that is not a booted system).
+{{< /alert >}}
+
+{{< /tablestep >}}
+{{< tablestep >}}
+**4. Ensure NetworkManager is installed**
+
+Make sure NetworkManager is installed and enabled in `systemd`.
```
Otterverse, line 291: Also that it's version 1.42 or newer for full support. (May work with slightly older versions, but not fully tested.)

Otterverse (changes requested):
Issues inline!

docs/how-tos/provision-setup.md
npentrel: adding to the top of the page.

docs/how-tos/provision-setup.md
npentrel: I'll change the title. I think it's good to have for completeness. People may read through this guide without doing the actions at the time, just to get an overview.

docs/how-tos/provision.md
npentrel: I do think we should hide the password on the captive portal. It's a bit inconvenient as is

npentrel:
@Otterverse ready for re-review. Also have tested the flows now so I think we're hopefully in a good state

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,287 @@
...
+## Decide on the provisioning method
+
+You can choose to let your end users complete machine setup by using a captive web portal or a mobile app.
+
+If you choose to have a mobile app experience, you can use the [Viam mobile app](/fleet/control/#control-interface-in-the-viam-mobile-app) or create your own custom mobile app using the [Flutter SDK](https://flutter.viam.dev/viam_protos.provisioning.provisioning/ProvisioningServiceClient-class.html) or the [TypeScript SDK](https://github.com/viamrobotics/viam-typescript-sdk/blob/main/src/app/provisioning-client.ts) to connect to `viam-agent` and provision your machines.
+
+If you choose to use the Viam mobile app, you must provide a {{< glossary_tooltip term_id="fragment" text="fragment" >}} for provisioning.
+If you do not yet have a fragment, follow the steps to [Create a configuration fragment](/how-tos/one-to-many/) and make a note of the fragment ID.
+
+If you choose to use the captive web portal, you can optionally create a machine already and provide its credentials configuration file at <FILE>/etc/viam.json</FILE>.
```
Otterverse, line 46:
```suggestion
If you choose to use the captive web portal, you can optionally create a machine in advance and provide its credentials configuration file at <FILE>/etc/viam.json</FILE>.
```

docs/how-tos/provision-setup.md
```diff
@@ -0,0 +1,287 @@
...
+
+{{% /tab %}}
+{{% tab name="x86_64" %}}
+
+```sh {class="command-line" data-prompt="$"}
+sudo ./preinstall.sh --x86_64
+```
+
+{{% /tab %}}
+{{< /tabs >}}
```
Otterverse, line 169: These commands need more elaboration. They create a tarball the user needs to then manually extract into their image/filesystem as a quicker way to do a manual install. They're not just what to run if there is an error.

docs/how-tos/provision.md
```diff
@@ -0,0 +1,146 @@
...
+
+1. [Set up your machine using the Viam mobile app](#set-up-your-machine-using-the-viam-mobile-app).
+1. [Set up your machine using the captive portal](#set-up-your-machine-using-the-captive-portal).
+
+{{< /alert >}}
+
+## Prerequisites
+
+- Physical hardware constituting a machine
+- A WiFi-enabled laptop or mobile device if using the Viam mobile app
```
Otterverse, line 25:
```suggestion
- A WiFi-enabled laptop/pc/device, or mobile device (if using the Viam mobile app)
```

Otterverse (changes requested):
Looking good! Only one real thing left (and two minor things.) All inline below.

docs/how-tos/provision-setup.md
npentrel: I've added more info based on reading the script. I don't know a lot about mounting file systems so please add info if needed

Otterverse (approved):
LGTM! Thanks for the patience with all the back and forth!

PR #3277, opened 2024-08-19
IGNORE; for staging only

JessamyT:
This is just for A/B comparison for formatting decision making
https://docs-test.viam.dev/3277/how-tos/sensor-module/


JessamyT:
No longer using

PR #3288, opened 2024-08-20
DOCS-2208: Add note about overwrite failures

npentrel:
Quote from ticket: "We should indicate that if there are problems with the users mods command that result in errors in overwriting fragments, the overwrites will not be partially applied; the config will simply not be returned until the mods command is fixed. User should refer to the robot logs to see what's wrong with their mods command."

@piokasar can you check that this is enough explanation? I've tested making the fragment mods purposefully supply illegal values and it seems the config does get applied and then the machine logs errors.
What would be an example error where the overwrite fails? And does that then mean if only an overwrite command fails the existing configuration stays until further changes?

### DIFF
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -301,7 +301,7 @@ Here are the version options:
 {{< /tabs >}}
 
 After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
-If there are problems with fragment modifications, depending on the issue, your machine or some of its component may not work until the configuration is fixed.
+If there are problems with fragment modifications, depending on the issue, your machine or some of its components may not work until the configuration is fixed.
 
 If you need to restore the original fragment, click the **...** in the upper right corner of the card you modified, and click **Revert changes**.
 
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -42,7 +42,7 @@ Get started with this how-to guide:
 {{% card link="/how-tos/one-to-many/" %}}
 {{< /cards >}}
 
-## Overwrite the config of a machine that uses a fragment
+## Modify the config of a machine that uses a fragment
 
 If you need to modify the config of one or a few machines that use a fragment, you can overwrite fields of the fragment.
 
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -301,7 +301,7 @@ Here are the version options:
 {{< /tabs >}}
 
 After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
-If there are problems with overwrites to the fragment, the overwrites will not be partially applied and the configuration changes will not take effect until the configuration is fixed.
+If there are problems with overwrites to the fragment, no overwrites will be applied and the configuration changes will not take effect until the configuration is fixed.
 
 If you need to restore the original fragment, click the **...** in the upper right corner of the card you modified, and click **Revert changes**.
 
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -42,14 +42,14 @@ Get started with this how-to guide:
 {{% card link="/how-tos/one-to-many/" %}}
 {{< /cards >}}
 
-## Modify the config of a machine that uses a fragment
+## Overwrite the config of a machine that uses a fragment
 
 If you need to modify the config of one or a few machines that use a fragment, you can overwrite fields of the fragment.
 
 When you add a fragment to a machine, the configuration of components and services included in a fragment are _read-only_.
 Any changes made by overwriting them do not change the fragment itself.
 
-If you need to make changes to all machines that use a fragment, you must modify the fragment itself.
+If you need to make changes to all machines that use a fragment, modify the fragment itself instead.
 
 {{% alert title="Support Notice" color="note" %}}
 
@@ -299,6 +299,9 @@ Here are the version options:
 
 {{% /tab %}}
 {{< /tabs >}}
+
+After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
+If there are problems with overwrites to the fragment, the overwrites will not be partially applied and the configuration changes will not take effect until the configuration is fixed.
 
 If you need to restore the original fragment, click the **...** in the upper right corner of the card you modified, and click **Revert changes**.
 
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -301,7 +301,7 @@ Here are the version options:
 {{< /tabs >}}
 
 After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
-If there are problems with overwrites to the fragment, no overwrites will be applied and the configuration changes will not take effect until the configuration is fixed.
+If there are problems with fragment modifications, depending on the issue, your machine or some of its component may not work until the configuration is fixed.
 
 If you need to restore the original fragment, click the **...** in the upper right corner of the card you modified, and click **Revert changes**.
 
--- a/docs/fleet/fragments.md
+++ b/docs/fleet/fragments.md
@@ -300,7 +300,7 @@ Here are the version options:
 {{% /tab %}}
 {{< /tabs >}}
 
-After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
+After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
 If there are problems with overwrites to the fragment, no overwrites will be applied and the configuration changes will not take effect until the configuration is fixed.
 
 If you need to restore the original fragment, click the **...** in the upper right corner of the card you modified, and click **Revert changes**.
--- a/docs/how-tos/one-to-many.md
+++ b/docs/how-tos/one-to-many.md
@@ -104,7 +104,7 @@ If some of your machines have slight differences, you can still add the fragment
 
 ## Modify a fragment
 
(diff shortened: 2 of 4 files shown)


docs/fleet/fragments.md
```diff
@@ -300,6 +300,9 @@ Here are the version options:
 {{% /tab %}}
 {{< /tabs >}}
 
+After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
+If there are problems with overwrites to the fragment, the overwrites will not be partially applied and the configuration changes will not take effect until the configuration is fixed.
```
JessamyT, line 304: "will not be partially applied" is a confusing phrase. Will they be fully applied? Not applied at all?

JessamyT (commented):
To me, "overwrite the config" implies that you're overwriting the whole thing, as opposed to overwriting portions of it. I think modify doesn't have the same connotation.

docs/fleet/fragments.md
npentrel, line 304:
```suggestion
If there are problems with overwrites to the fragment, no overwrites will be applied and the configuration changes will not take effect until the configuration is fixed.
```

docs/fleet/fragments.md
```diff
@@ -42,14 +42,14 @@ Get started with this how-to guide:
 {{% card link="/how-tos/one-to-many/" %}}
 {{< /cards >}}
 
-## Modify the config of a machine that uses a fragment
+## Overwrite the config of a machine that uses a fragment
```
npentrel, line 45:
```suggestion
## Modify the config of a machine that uses a fragment
```

docs/how-tos/one-to-many.md
```diff
@@ -104,15 +104,15 @@ If some of your machines have slight differences, you can still add the fragment
...
+If your machines are similar but not identical, you can use a fragment with all of them, and then [overwrite parts of it](/fleet/fragments/#overwrite-the-config-of-a-machine-that-uses-a-fragment) to customize select fields of the configuration without modifying the upstream fragment.
 
 {{% alert title="Note" color="note" %}}
 If you modify fields within a fragment, your modifications will act as overwrites.
 If you later update the upstream fragment, the updates to the overwritten fields will effectively not be updated.
 {{% /alert %}}
 
 {{< table >}}
-{{% tablestep link="/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment" %}}
+{{% tablestep link="/fleet/fragments/#overwrite-the-config-of-a-machine-that-uses-a-fragment" %}}
```
npentrel, line 115:
```suggestion
{{% tablestep link="/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment" %}}
```

docs/tutorials/control/air-quality-fleet.md
```diff
@@ -312,11 +312,11 @@ For each machine:
    {{< expand "Click here for usb_interface troubleshooting help" >}}
 
 If you only have one USB device plugged into each of your boards, the `usb_interface` value you configured in the sensor config is likely (conveniently) the same for all of your machines.
-If not, you can use [fragment overwrite](/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment) to modify the value on any machine for which it is different:
+If not, you can use [fragment overwrite](/fleet/fragments/#overwrite-the-config-of-a-machine-that-uses-a-fragment) to modify the value on any machine for which it is different:
```
npentrel, line 315:
```suggestion
If not, you can use [fragment overwrite](/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment) to modify the value on any machine for which it is different:
```

docs/tutorials/control/air-quality-fleet.md
```diff
@@ -312,11 +312,11 @@ For each machine:
...
 
 1. If you're not getting sensor readings from a given machine, check the path of the USB port using the same [process by which you found the first USB path](#usb-path).
 2. If the path to your sensor on one machine is different from the one you configured in the fragment, add a fragment overwrite to the config of that machine to change the path without needing to remove the entire fragment.
-   Follow the [instructions to add a fragment overwrite](/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment) to your machine's config, using the following JSON template:
+   Follow the [instructions to add a fragment overwrite](/fleet/fragments/#overwrite-the-config-of-a-machine-that-uses-a-fragment) to your machine's config, using the following JSON template:
```
npentrel, line 319:
```suggestion
   Follow the [instructions to add a fragment overwrite](/fleet/fragments/#modify-the-config-of-a-machine-that-uses-a-fragment) to your machine's config, using the following JSON template:
```

docs/how-tos/one-to-many.md
```diff
@@ -125,10 +125,17 @@ You can click the **{}** button to switch to advanced view and see the changes.
 
 Don't forget to **Save**.
 
+{{% /tablestep %}}
+{{% tablestep %}}
+**2. Check your machine's logs**
+
+After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```
npentrel, line 132:
```suggestion
After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```

docs/fleet/fragments.md
```diff
@@ -300,6 +300,9 @@ Here are the version options:
 {{% /tab %}}
 {{< /tabs >}}
 
+After configuring overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```
npentrel, line 303:
```suggestion
After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```

docs/fleet/fragments.md
```diff
@@ -300,6 +300,9 @@ Here are the version options:
...
+If there are problems with overwrites to the fragment, no overwrites will be applied and the configuration changes will not take effect until the configuration is fixed.
```
npentrel, line 304:
```suggestion
If there are problems with fragment modifications, no configuration changes will take effect until the configuration is fixed.
```

docs/fleet/fragments.md
```diff
@@ -300,6 +300,9 @@ Here are the version options:
 {{% /tab %}}
 {{< /tabs >}}
 
+After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
+If there are problems with fragment modifications, depending on the issue, your machine or some of its component may not work until the configuration is fixed.
```
npentrel, line 304:
```suggestion
If there are problems with fragment modifications, depending on the issue, your machine or some of its components may not work until the configuration is fixed.
```

docs/fleet/fragments.md
```diff
@@ -300,6 +300,9 @@ Here are the version options:
 {{% /tab %}}
 {{< /tabs >}}
 
+After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
+If there are problems with fragment modifications, depending on the issue, your machine or some of its components may not work until the configuration is fixed.
```
npentrel, line 304:
```suggestion
If there are problems with fragment modifications, depending on the issue, your machine or some of its components may not work until the configuration is fixed.
```

docs/how-tos/one-to-many.md
```diff
@@ -125,10 +125,17 @@ You can click the **{}** button to switch to advanced view and see the changes.
 
 Don't forget to **Save**.
 
+{{% /tablestep %}}
+{{% tablestep %}}
+**2. Check your machine's logs**
+
+After configuring fragment modifications, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```
JessamyT, line 132: [opt] This is a place where I think either makes sense. What I meant with my other comment is that I think it makes sense to either day "modify a fragment" or "overwrite a portion of a fragment." I just think the confusing thing is if we say "overwrite a fragment."
```suggestion
After configuring fragment overwrites, check your machine's [**LOGS** tab](/cloud/machines/#logs).
```

JessamyT (approved):
LGTM

docs/fleet/fragments.md
piokasar: You could say instead "the overwrite will fail and your config will not be returned to your machine; your robot will instead rely on its cached config until you fix the overwrite error and the next config fetch is successful".

piokasar:
> @piokasar can you check that this is enough explanation? I've tested making the fragment mods purposefully supply illegal values and it seems the config does get applied and then the machine logs errors.
> What would be an example error where the overwrite fails? And does that then mean if only an overwrite command fails the existing configuration stays until further changes?

- The overwrite fails either if the overwrite is invalid (example: deleting a field that doesn't exist)
- This is a user error, so the logs will indicate that 
- If the error is internal, i.e. on our end, the logs will indicate that too. The user can't really do anything in this case.
- If the overwrite command fails, the existing configuration, i.e. the cached config, stays until the next config fetch is successful

docs/fleet/fragments.md
piokasar, line 304: I would say "may not work as expected" instead of "may not work". This is because if the machine was previously working with the old config, even if the overwrite fails, it will continue to work. It won't be the new config with the overwrite applied though.

npentrel: I overwrote a pin value with a negative value for a motor I think. The overwrite succeeded but the resulting change causes an illegal config for the component. So it in facet may not continue to work. I suppose you can argue that the overwrite succeeded but you still get logged errors and the component is no longer appearing on the control tab if I do that

piokasar: Ah I see. yeah the overwrite succeeded in that case because the field existed; we can't tell if the pin value itself is wrong - that happens on the robot

piokasar (approved):
this lgtm! thanks for explaining your reasoning

PR #3289, opened 2024-08-20
DOCS-2828: Split Upload a Module into a separate how-to

JessamyT:


docs/how-tos/upload-module.md
```diff
@@ -0,0 +1,276 @@
...
+images: ["/registry/module-puzzle-piece.svg"]
+tags: ["modular resources", "components", "services", "registry"]
+description: "Make your module available to others by uploading it to the modular resource registry."
+aliases:
+  - /registry/upload/
+  - /extend/modular-resources/upload/
+  - /modular-resources/upload/
+---
+
+After [writing a module](/how-tos/create-module/), you can upload your module to the Viam registry to:
```
npentrel, line 15: Missing prerequisites section and missing in this page card

docs/how-tos/upload-module.md
```diff
@@ -0,0 +1,276 @@
...
+You can choose to upload it to the Viam registry as a _public module_ that is shared with other Viam users, or as a _private module_ that is shared only within your [organization](/cloud/organizations/).
+
+To upload your custom module to the [Viam registry](https://app.viam.com/registry), either as a public or private module, use the Viam CLI commands `create` and `upload` following these instructions:
+
+{{< table >}}
+{{% tablestep %}}
+**1. Install the CLI**
+
+First, [install the Viam CLI](/cli/#install) and [authenticate](/cli/#authenticate) to Viam, from the same machine that you intend to upload your module from.
+
```
npentrel, line 29: we should propably make the instructions an include so that they can be included rather than linking out

docs/how-tos/upload-module.md
```diff
@@ -0,0 +1,276 @@
...
+This command creates a new `meta.json` metadata file in your current working directory, which serves as a template.
+
+{{% /tablestep %}}
+{{% tablestep %}}
+**3. Edit the metadata file**
+
+Edit the newly-created `meta.json` file, and provide the required configuration information for your custom module by filling in the following fields.
+
+<table class="table table-striped">
```
npentrel, line 73: this is fine for now but we probably want this table to live in reference docs in the future. just pointing it out to ensure we're on the same page

npentrel (approved):

docs/how-tos/upload-module.md
npentrel: Actually this should probably also be a prerequisite. 

And you can use this:

```
{{< readfile "/static/include/how-to/install-cli.md" >}}
```

There's currently an extra sentence in it but I am removing that in DOCS-2794

npentrel: merging that here https://github.com/viamrobotics/docs/pull/3291 for you

docs/how-tos/upload-module.md
JessamyT: Okay will add prereqs but I temporarily added an in this page card and then took it away again because it's only one item, with nothing to link to, so it just looks weird and distracts from the intro paragraph.

JessamyT: Adding it anyway--lmk if you don't like it

PR #3302, opened 2024-08-22
DOCS-2785: Update installation guide for new UI

npentrel:
@loopDelicious you had said the installation for esp32 was confusing, could you review 
https://docs-test.viam.dev/3302/installation/ and tell me what you think?

### DIFF
--- a/docs/installation/_index.md
+++ b/docs/installation/_index.md
@@ -29,7 +29,7 @@ aliases:
 
 To use Viam, you need to install either the `viam-server` binary or `viam-micro-server`.
 
-[`viam-server`](/get-started/#viam-server) is the binary built from the [Robot Development Kit](https://github.com/viamrobotics/rdk) that contains and manages communications between all Viam's built-in hardware drivers ({{< glossary_tooltip term_id="component" text="components" >}}) and software {{< glossary_tooltip term_id="service" text="services" >}}, connects your machine to the cloud, manages machine configuration, and manages dependencies including {{< glossary_tooltip term_id="module" text="modules" >}}.
+[`viam-server`](/architecture/rdk/) is the binary built from the [Robot Development Kit](https://github.com/viamrobotics/rdk) that contains and manages communications between all Viam's built-in hardware drivers ({{< glossary_tooltip term_id="component" text="components" >}}) and software {{< glossary_tooltip term_id="service" text="services" >}}, connects your machine to the cloud, manages machine configuration, and manages dependencies including {{< glossary_tooltip term_id="module" text="modules" >}}.
 
 `viam-micro-server` is a lightweight version of `viam-server` which can run on resource-limited embedded systems that cannot run the fully-featured `viam-server`.
 `viam-micro-server` is built from the [micro-RDK](https://github.com/viamrobotics/micro-rdk/tree/main).
--- a/docs/installation/_index.md
+++ b/docs/installation/_index.md
@@ -169,7 +169,7 @@ If you have a [compatible microcontroller](/installation/#compatibility), follow
 
 1. Select the platform you want to use from which to flash the target system.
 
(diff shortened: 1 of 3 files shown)


docs/installation/_index.md
```diff
@@ -165,12 +165,11 @@ If you have a [compatible microcontroller](/installation/#compatibility), follow
...
 
-1. Select your system's operating system and **Micro-RDK** as your RDK type.
+1. Select **ESP32** as the platform you want to run on.
 
-1. Follow the instructions to flash `viam-micro-server` directly to an ESP32 connected to your computer through a data cable.
+1. Select the platform you want to use from which to flash the target system.
 
-   To see the `viam-micro-server` logs through the serial connection, add `--monitor` to the command in step 3.
-   If the program cannot auto-detect the serial port to which your ESP32 is connected, you may be prompted to select the correct one among a list.
+1. Follow the instructions to flash `viam-micro-server` directly to an ESP32 connected to your computer through a data cable.
```
npentrel, line 172:
```suggestion
1. Follow the instructions to flash `viam-micro-server` to an ESP32 connected to your computer through a data cable.
```

loopDelicious:
- **Pages setup**: If someone lands directly on the `viam-server`, `viam-micro-server`, or `viam-agent` page (not from the Installation Guide page), include a one-liner at the top to explain the difference between the 3. I thought there was previously a callout on `viam-server` to say if you're working w/microcontroller, use the microRDK instead. If I land on `viam-agent`, when would I need to install that instead of the other options.
- **MicroRDK page flow**: I'm curious why setting up your dev env comes before installing `viam-micro-server`. From my perspective, having something running would come prior to customization.
- **MicroRDK installation**: An extra step was required for me to configure current shell, source corresponding env file under `$HOME/.cargo` before I could use `cargo`. So I needed to run `. "$HOME/.cargo/env"`
- **MicroRDK compatibility**: I didn't know a plain ESP32 was different than other varieties of ESP32 (like ESP32-S3, thought S3 referred to the kit), so I chose a board that was part of WROOM series that had the following resources as listed in the docs. At the time, it wasn't supported, would it be reasonable to list model numbers, or check if something is supported?

npentrel:
> * **Pages setup**: If someone lands directly on the `viam-server`, `viam-micro-server`, or `viam-agent` page (not from the Installation Guide page), include a one-liner at the top to explain the difference between the 3. I thought there was previously a callout on `viam-server` to say if you're working w/microcontroller, use the microRDK instead. If I land on `viam-agent`, when would I need to install that instead of the other options.

Added to viam-server. I think the agent page already has this. And viam-micro-server differentiates between viam-server and itself. 

> * **MicroRDK page flow**: I'm curious why setting up your dev env comes before installing `viam-micro-server`. From my perspective, having something running would come prior to customization.

It isn't supposed to... I think you landed on the dev setup rather than the [primary installation page](https://docs.viam.com/installation/#install-viam-micro-server). I'll add a note at the top of the dev install page that directs people there.

> * **MicroRDK installation**: An extra step was required for me to configure current shell, source corresponding env file under `$HOME/.cargo` before I could use `cargo`. So I needed to run `. "$HOME/.cargo/env"`

Added this in a separate PR, thanks!

> * **MicroRDK compatibility**: I didn't know a plain ESP32 was different than other varieties of ESP32 (like ESP32-S3, thought S3 referred to the kit), so I chose a board that was part of WROOM series that had the following resources as listed in the docs. At the time, it wasn't supported, would it be reasonable to list model numbers, or check if something is supported?

We do list the series here. ESP32-S3 is not yet supported but will be coming relatively soon, at which point it'll be added here. https://docs.viam.com/installation/#compatibility

npentrel:
Merging for now further changes can happen in subsequent PRs

PR #3350, opened 2024-08-27
DOCS-2832: Move datasets info into existing how-to

JessamyT:
Add GIFs into existing datasets how-to in expanders, and delete some seemingly pretty intuitive instructions such as how to delete an image from a dataset


docs/how-tos/deploy-ml.md
```diff
@@ -113,6 +157,8 @@ Test your ML model classifier with [existing images in the Viam app](/services/v
 
 ## Next steps
 
+To work with datasets programmatically, see the [data API](/appendix/apis/data-client/#api) which includes several methods to work with datasets.
```
npentrel, line 160: to make more visible
```suggestion
To work with datasets programmatically, see the [data API](/appendix/apis/data-client/#api) which includes several methods to work with datasets:

{{< cards >}}
  {{% card link="/appendix/apis/data-client/" %}}
{{< /cards >}}
```

docs/services/data/dataset.md
```diff
@@ -1,153 +0,0 @@
----
```
npentrel, line 1: I think we do want to keep a dataset page to have reference information. I am thinking something like this:

```
title: "Dataset"
---

On the [**DATA** tab](https://app.viam.com/data/view) in the Viam app, you can create and manage datasets.
A dataset is a grouping of image data that you can add labels to.
You can use datasets to traing Machine Learning models on. 

{{< alert title="Info" color="info" >}}
Filtered datasets are views and not materialized.
That means the data you are viewing may change as you label and train on the dataset.
If the underlying data matching the filter changes because data is deleted or more data is added, the dataset will also change.
{{< /alert >}}

## Labels

You can label your data with:

- [Bounding boxes](#bounding-boxes), which you can use to [train object detection models](/services/ml/train-model/#train-a-model).
- [Image tags](#image-tags), which you can use to train [image classification models](/services/ml/train-model/#train-a-model).

## API

To interact with datasets programmatically, you use the Dataset API, which supports the following methods:

{{< readfile "/static/include/app/apis/generated/dataset-table.md" >}}

The data client API supports a separate set of methods that allow you to upload and export data to and from the Viam app.
For information about that API, see [Data Client API](/appendix/apis/data-client/).

## Next steps

Link to how-tos.
...



```

npentrel (changes requested):
Looks good for the most part but I do think we want to keep a dataset page, just very very minimal.

npentrel (approved):

PR #3373, opened 2024-08-30
Update sensor-module.md

npentrel:


docs/how-tos/sensor-module.md
```diff
@@ -292,27 +293,32 @@ class meteo_PM(Sensor, Reconfigurable):
...
-            longitude = -121
+        # Check that configured fields are floats
+        if "latitude" in config.attributes.fields and
+            if not config.attributes.fields["latitude"].HasField("number_value"):
+                raise Exception("Latitude must be a float.")
+        else:
+            self.default_lat = 45
+        if "longitude" in config.attributes.fields:
+            if not config.attributes.fields["latitude"].HasField("number_value"):
+                raise Exception("Latitude must be a float.")
```
HipsterBrown, line 304:
```suggestion

        if "longitude" in config.attributes.fields:
            if not config.attributes.fields["longitude"].HasField("number_value"):
                raise Exception("Longitude must be a float.")
```

HipsterBrown (changes requested):
There is an existing import of `viam.utils` for the `SensorReading` class, `struct_to_dict` can be tacked onto that line rather than add a new line:

```python
from viam.utils import SensorReading, struct_to_dict
```

Also, the `default_lat` and `default_long` class parameters should be set with the `latitude` and `longitude` class parameters outside of this method. If they don't get set because the fields exist in the config.attributes, using them in the `reconfigure` method will cause an error:

```python
# Class parameters
latitude: float  # Latitude at which to get data
longitude: float # Longitude at which to get data
default_lat: float = 45 # Default latitude for configuration
default_long = -121 # Default longitude for configuration
```

npentrel:
Good points - thanks for testing @HipsterBrown

HipsterBrown (approved):

docs/how-tos/sensor-module.md
```diff
@@ -292,27 +294,33 @@ class meteo_PM(Sensor, Reconfigurable):
...
-        latitude = config.attributes.fields["latitude"].number_value
-        if latitude == "":
-            # Set a default
-            latitude = 45
-        longitude = config.attributes.fields["longitude"].number_value
-        if longitude == "":
-            # Set a default
-            longitude = -121
+        # Check that configured fields are floats
+        if "latitude" in config.attributes.fields and
```
npentrel, line 298:
```suggestion
        if "latitude" in config.attributes.fields:
```

docs/how-tos/sensor-module.md
```diff
@@ -277,7 +277,9 @@ class meteo_PM(Sensor, Reconfigurable):
 
     # Class parameters
     latitude: float  # Latitude at which to get data
-    longitude: float  # Longitude at which to get data
+    longitude: float # Longitude at which to get data
+    default_lat: float = 45 # Default latitude for configuration
+    default_long = -121 # Default longitude for configuration
```
npentrel, line 282:
```suggestion
    longitude: float  # Longitude at which to get data
    default_lat: float = 45  # Default latitude for configuration
    default_long = -121  # Default longitude for configuration
```

PR #3383, opened 2024-09-02
Improve trigger sync guide

npentrel:
Minor update to the performance data guide.
Structural change to the trigger sync guide. The guide is supposed to teach people how to add their own logic, I think it does a better job at that now.


npentrel:
Going to merge since the change is rather small. If anyone has feedback we can address separately.

PR #3447, opened 2024-09-12
Apply architecture feedback

JessamyT:


docs/architecture/_index.md
```diff
@@ -169,11 +170,15 @@ Many simple {{< glossary_tooltip term_id="machine" text="machines" >}} consist o
...
-- Multiple individual machines, linked by a {{< glossary_tooltip term_id="remote-part" text="remote" >}} connection.
+- Multiple individual machines (each made up of one or more parts), linked by a {{< glossary_tooltip term_id="remote-part" text="remote" >}} connection.
 
 These two options are very similar: in both cases, the parts communicate with each other securely and directly using gRPC/{{< glossary_tooltip term_id="webrtc" text="WebRTC" >}}.
+Any given part can be a remote part of multiple machines, whereas a part can only be a sub-part of one machine.
+In other words, remote connections allow sharing of resources across multiple machines, whereas main parts and sub-parts are a way to hierarchically organize one machine.
 
-Because the parts are interconnected, you can write SDK code that establishes a connection with one of them and controls them all in a coordinated way.
+Connecting parts (either as main part and sub-part, or as part and remote part) means that you can write SDK code that establishes a connection with one part and controls them all in a coordinated way.
```
npentrel, line 179: would you have to connect to the main part though?

docs/architecture/parts.md
```diff
@@ -15,26 +15,43 @@ aliases:
...
+## Machine parts
 
-Machines are organized into _parts_, where each part represents a computer (a single-board computer, desktop, laptop, or other computer) running [`viam-server`](/installation/), the hardware [components](/components/) attached to it, and any [services](/services/) or other resources running on it.
+Machines are organized into _parts_, where each part represents a computer (a single-board computer, desktop, laptop, or other computer) running [`viam-server`](/installation/), the hardware [components](/components/) attached to it, and any [services](/services/) or other {{< glossary_tooltip term_id="resource" text="resources" >}} running on it.
 
 Every smart machine has a main part which is automatically created when you create the machine.
 Multi-part machines also have one or more _sub-parts_ representing additional computers running `viam-server`.
 
-There are two ways to link machine parts:
+There are two ways to link machine parts, depending on whether you want to think of your parts as all part of one machine, or be able to share resources across multiple machines:
```
npentrel, line 25: I think this change makes it more confusing to me. I think either stick with how it was and explain the rest as it is in the bullets or maybe something like this?

```suggestion
You can think of parts as machine parts that together make up a whole machine where each part can also be shared as a resource across multiple machines:
```

docs/architecture/parts.md
```diff
@@ -15,26 +15,43 @@ aliases:
...
+  Any given part can only be a sub-part of one main part.
 
-- **Remote part**: To connect two computers that are parts of _different machines_ in the same or different organizations, [add one machine part as a remote part of the other machine](#configure-a-remote-part).
+  <details>
+    <summary>Click for an example.</summary>
+   Imagine you have a system of five cameras in different places along an assembly line, each attached to its own single-board computer, and you want to run an object detector on the streams from all of them.
+   You have one main computer with greater compute power set up as the main part.
+   You set up each of the single-board computers as a sub-part.
+   This allows the main part to access all the camera streams and run object detection on all of them.<br><br>
+   You could also set this up with each single-board computer being a remote part instead of a sub-part, but if you consider them all to be part of one system, it can be easier to configure the connections this way.
```
npentrel, line 36: what does "can be easier" mean - is it easier? do you have to do less? That needs to be clearer

docs/architecture/_index.md
JessamyT: Yes, but since with remotes you don't talk about "main part" in the same way I left it vague. However, maybe can gloss over and make this clearer:
```suggestion
Connecting parts (either as main part and sub-part, or as part and remote part) means that you can write SDK code that establishes a connection with the main part and controls all parts in a coordinated way.
```

docs/architecture/parts.md
JessamyT: What I was trying to get across is that you could just use remotes for all cases, but making a multi-part machine is just kind of an organizational concept.

docs/architecture/parts.md
JessamyT: It is ever so slightly easier:

```suggestion
   You could also set this up with each single-board computer being a remote part instead of a sub-part, but it is slightly easier to configure sub-parts because you do not need to add the address of each part to your machine's config.
   Additionally, configuring a discrete system of parts as one multi-part machine helps keep your fleet more clearly organized in the Viam app.
```

docs/architecture/parts.md
JessamyT:
```suggestion
There are two ways to link machine parts:
```

docs/architecture/parts.md
npentrel: yeah that is better

npentrel (approved):
LGTM % the last suggestion

PR #3460, opened 2024-09-13
DOCS-2832: Move export data to how-to

JessamyT:
Also adds `icon: true` to how-to frontmatter where applicable


docs/how-tos/export-data.md
```diff
@@ -0,0 +1,97 @@
...
+
+- [Export data with the Viam CLI](#export-data-with-the-viam-cli)
+- [Manage data with the data client API](#manage-data-with-the-data-client-api)
+
+{{% /alert %}}
+
+## Prerequisites
+
+{{< expand "Install the Viam CLI and authenticate." >}}
+[Install the Viam CLI](/cli/#install), then [authenticate your session with Viam](/cli/#authenticate).
```
npentrel, line 34: Please follow the pattern we use in other guides. We want to provide the information here so they don't have to switch pages.

docs/how-tos/export-data.md
```diff
@@ -0,0 +1,97 @@
...
+{{< /expand >}}
+
+## Export data with the Viam CLI
+
+To export your synced data using the Viam CLI:
```
npentrel, line 39: It could also be data they uploaded. So I think let's not say synced data
```suggestion
To export your data from the cloud using the Viam CLI:
```

docs/how-tos/export-data.md
```diff
@@ -0,0 +1,97 @@
...
+{{% tablestep link="/cli/#data" %}}
+**3. Run the command**
+
+Run the copied command in a terminal:
+
+```sh {class="command-line" data-prompt="$"}
+viam data export --org-ids=<org-id> --data-type=<binary|tabular> --mime-types=<mime types> --destination=.
+```
+
+This command uses the Viam CLI to download the data locally onto your computer based on the search criteria you select in the Viam app.
```
npentrel, line 68: (not needed/duplicate)
```suggestion
This command uses the Viam CLI to download the data onto your computer based on the search criteria you select in the Viam app.
```

docs/how-tos/export-data.md
```diff
@@ -0,0 +1,97 @@
...
+
+Since data is downloaded in parallel, the order is not guaranteed.
+Sort your folder by filename in order to see them in chronological order.
+
+{{% /tablestep %}}
+{{< /table >}}<br>
+
+You can see more information about exporting data in the [Viam CLI documentation](/cli/#data).
+
+## Manage data with the data client API
```
npentrel, line 83: i feel like having this as a section when it has no content is somewhat misleading. I think work this information into the intro more and then get rid of this section.

docs/how-tos/export-data.md
```diff
@@ -0,0 +1,97 @@
...
+This API includes a set of methods for managing data, including export, batch delete, tag, upload, and many more.
+
+## Next steps
+
+Other how-to guides for using and querying data include:
+
+{{< cards >}}
+{{% card link="/how-tos/image-data/" %}}
+{{% card link="/how-tos/deploy-ml/" %}}
+{{% card link="/how-tos/sensor-data-query-sdk/" %}}
```
npentrel, line 95: I think when you're here you will have already synced data most of the time so the image-data guide seems unnecessary. And the sdk one will hopefully soon go away in favour of more examples moving to the app. So I think these are probably better to link to

```suggestion
{{% card link="/how-tos/deploy-ml/" %}}
{{% card link="/how-tos/sensor-data-visualize/" %}}
```

npentrel (approved):
LGTM % small change

PR #3506, opened 2024-09-26
DOCS-2608: Add emailing to triggers

sguequierre:
* noticed that this came up with contrasts with the ticket saying emailing was added for all trigger types: should we report? a bit confused 
image

* already mentioned emails in this docs so I think all that's left is to add it to the attribute table as done here


npentrel (changes requested):
You missed the UI changes, those instruct you to add a URL when it should be either or

sguequierre:
> You missed the UI changes, those instruct you to add a URL when it should be either or

sorry, what do you mean? I did put it as either or

npentrel:
Screenshot 2024-09-27 at 16 41 21

docs/configure/triggers.md
```diff
@@ -46,22 +46,28 @@ For example, if you want to trigger a trigger on temperature readings, configure
 {{% /alert %}}
 
 {{% /tab %}}
```
npentrel, line 48: So I think the create menu option doesn't need the tabs here because it's now just one item. I am wondering if it wouldn't be better to get rid of the `Builder mode: Create menu` tab entirely since it's a shortcut for the other one and hopefully the rest should explain itself? I think the copy for adding webhooks & emails is better here though so merging that with the other tab would be good

sguequierre: I was thinking about the other webhook option that's about to be documented, then there'll be a second tab. And i'm fine removing that, will do now

sguequierre: but also wait are you sure you want to remove that? in my understanding those are their own types of triggers and you can't configure them otherwise/with plain builder mode, its not super self explanatory that you can do that. maybe if the conditional data ingested is the same thing but I'm not sure

npentrel: You can show me tomorrow in the UI and we can see if you can create the same trigger with just the add trigger ui

sguequierre: removed!

PR #3508, opened 2024-09-26
Update controls-package.md

npentrel:
@randhid do you think this is used for anything or should we delete this?

### DIFF
--- a/docs/internals/controls-package.md
+++ b/docs/internals/controls-package.md
@@ -5,6 +5,7 @@ weight: 10
 type: "docs"
 description: "The Control Package implements feedback control on an endpoint."
 # SME: Nicolas Menard, Rand, Martha
+draft: true
 ---
 
 [Encoded motors](/components/motor/encoded-motor/) and [sensor controlled bases](/components/base/sensor-controlled/) use a control loop that is implemented by `viam-server`.


randhid:
> @randhid do you think this is used for anything or should we delete this?

It is used in all our controlled components, and if you want to configure a loop in a go module to use our control logic, this shows you how. Is internals getting a face-lift, is that why we're converting it to draft?

If so, we can add it as a link to the components' pages, I did want to link to this page in the readme when we modularize the controlled components, which will happen in the next 4-5 weeks.

stevebriskin:
@randhid I told Naomi that this info is extremely internal and no one external would use it. Sounds like you disagree, which is fine.
Maybe we need a place where we document utilities like this (along with goutils type stuff). Or integrate it more natively into component config docs. "Viam Internals" isn't the right place for this if we expect people to use it.

randhid:
> @randhid I told Naomi that this info is extremely internal and no one external would use it. Sounds like you disagree, which is fine. Maybe we need a place where we document utilities like this (along with goutils type stuff). Or integrate it more natively into component config docs. "Viam Internals" isn't the right place for this if we expect people to use it.

I see. We can put it under motor and base same as we have "calibrate a camera" under camera, "cbirrt and tp-space" under motion, and "set up an rtk base" under movement sensor. It's in the same bucket as those things - which are specific ways of making some models of that api type work and are also libraries in some way (rimage, motionplan, gpsutils)

PR #3509, opened 2024-09-27
Replace laptop thumbnails

JessamyT:


npentrel:
Look for things like https://unsplash.com or freepik in the future.
These may still need cropping but I think are a bit better
dell-ocAuPlfZXEc-unsplash_1_-removebg-preview
dlxmedia-hu-UkpTGYox6RM-unsplash-removebg-preview

npentrel (approved):
Thanks!

PR #3514, opened 2024-09-30
DOCS-2981: Remove fromrobot methods

npentrel:
Also adds other updates from api docs re-generation


npentrel:
We do have them in the connection information for each resource - and if you look at what's getting removed they actually were only there for flutter.

PR #3527, opened 2024-10-03
DOCS-1847: Document conditional data ingested trigger

sguequierre:
* adds new type of trigger


docs/configure/triggers.md
```diff
@@ -204,8 +255,113 @@ The following attributes are available for triggers:
...
+  ]
+}
+```
+
+This eval would trigger for the following sensor reading:
+
+```json {class="line-numbers linkable-line-numbers"}
+{
+  "readings": {
+    "key": 10
```
sguequierre, line 302: no example so not sure if this is accurate (not sure whether there would be a key or not)-- can anyone confirm?

sguequierre: all the examples I have been able to configure in the ui do have keys

docs/configure/triggers.md
```diff
@@ -12,11 +12,12 @@ aliases:
 
 Triggers allow you to trigger webhooks when certain types of data are sent from your machine to the cloud, or when the your machine parts connect to Viam.
 For example, you can configure a trigger to send you a notification when your robot's sensor collects a new reading.
-Viam provides three trigger types depending on the event you want to trigger on:
+Viam provides four trigger types depending on the event you want to trigger on:
 
 - **Data has been synced to the cloud**: trigger when data from the machine is synced
 - **Part is online**: trigger continuously at a specified interval while the {{< glossary_tooltip term_id="part" text="machine part" >}} is online
 - **Part is offline**: trigger continuously at a specified interval while the machine part is offline
+- **Conditional data ingestion**: trigger any time data is captured from a specified component with a specified method
```
RoxyFarhad, line 20: I would add with specific method "and condition"

RoxyFarhad (changes requested):
Some comments based on the preview env: 

1. Need to add the Conditional Data Ingested Trigger here?
Screenshot 2024-10-04 at 2 09 13 PM

2. Need to add Trigger Condition here:
Screenshot 2024-10-04 at 2 10 09 PM

docs/configure/triggers.md
sguequierre: @RoxyFarhad

docs/configure/triggers.md
```diff
@@ -245,6 +402,7 @@ def trigger():
         "Max-Time-Received": data.get('max_time_received', 'no value'),
         "Data-Type": data.get('data_type', 'no value'),
         "File-Id": data.get('file_id', 'no value'),
+        "Trigger-Condition": data.get("trigger_condition", 'no value'),
```
sguequierre, line 405: is this accurate? @RoxyFarhad and the below

docs/configure/triggers.md
RoxyFarhad: yeah I think they all have the reading keys right now

RoxyFarhad (approved):

docs/configure/triggers.md
```diff
@@ -12,11 +12,12 @@ aliases:
 
 Triggers allow you to trigger webhooks when certain types of data are sent from your machine to the cloud, or when the your machine parts connect to Viam.
 For example, you can configure a trigger to send you a notification when your robot's sensor collects a new reading.
-Viam provides three trigger types depending on the event you want to trigger on:
+Viam provides four trigger types depending on the event you want to trigger on:
```
npentrel, line 15: good catch, maybe 
```suggestion
Viam provides the following trigger types depending on the event you want to trigger on:
```

docs/configure/triggers.md
```diff
@@ -134,6 +152,39 @@ To configure your trigger by using **JSON** mode instead of **Builder** mode, pa
...
+    {
+      "name": "<trigger name>",
+      "event": {
+        "type": "conditional_data_ingested",
+        "conditional": {
+          "data_capture_method": "<component>:<name-of-component>:<method>",
+          "condition": {
+            "evals": [
+              {
+                "operator": "<operator>",
```
npentrel, line 169: Please list the operators, so readers don't have to guess (I know they're further down but this makes it easier):
```suggestion
                "operator": "<lt|gt|lte|gte|eq|neq>",
```

docs/configure/triggers.md
```diff
@@ -39,6 +40,23 @@ To configure a trigger:
 Select the data types for which the Trigger should send requests.
```
npentrel, line 40: on line 35, change `   For the respective type, configure the respective attributes:` to `   For some types you can configure additional attributes:`. Otherwise people may wonder why their type isn't listed

docs/configure/triggers.md
```diff
@@ -134,6 +152,39 @@ To configure your trigger by using **JSON** mode instead of **Builder** mode, pa
...
+      "name": "<trigger name>",
+      "event": {
+        "type": "conditional_data_ingested",
+        "conditional": {
+          "data_capture_method": "<component>:<name-of-component>:<method>",
+          "condition": {
+            "evals": [
+              {
+                "operator": "<operator>",
+                "value": <object, string, or int>
```
npentrel, line 170: I don't actually see an example where this is string or int. Do you have an example for that? All the ones I tested just now use an object.

docs/configure/triggers.md
```diff
@@ -204,8 +255,114 @@ The following attributes are available for triggers:
...
+
+#### Conditions
+
+The `conditional` object for the `conditional_data_ingested` trigger includes the following options:
+
+<!-- prettier-ignore -->
+| Name | Type | Required? | Description |
+| ---- | ---- | --------- | ----------- |
+| `data_capture_method` | string | **Required** | The method of data capture to trigger on. <br> Example: `sensor:<name-of-component>:Readings`. |
+| `condition` | object | Optional | Any additional conditions for the method to fire the trigger. Leave out this object for the trigger to fire any time there is data synced. <br> Options: <ul><li>`evals`:<ul><li>`operator`: Logical operator for the condition. </li><li>`value`: Can be a number, string or another object like a key-value pair that specifies the shape of the readings. </li></ul></li></ul> |
```
npentrel, line 269: See above about value and whether that can be a string or int. Also this doesn't just work on readings, it is also available for a range of other components. So mentioning readings is confusing. We do need to explain what the key and value are a bit more. Especially for components/conditions where the key doesn't have a popoulated drop down. We need to explain what the values should be.

docs/configure/triggers.md
```diff
@@ -204,8 +255,114 @@ The following attributes are available for triggers:
...
+{{% tab name="2 levels of nesting" %}}
+
+```json {class="line-numbers linkable-line-numbers"}
+"condition": {
+  "evals": [
+    {
+      "operator": "lt",
+      "value": {
+        “level_1_of_object”: {
+          “level_2_of_object”: 50
```
npentrel, line 344: This uses smart quotes - you must've copied it from somewhere that's why the formatting is weird (and code would probably not run)
```suggestion
        "level_1_of_object": {
          "level_2_of_object": 50
```

docs/configure/triggers.md
npentrel: @sguequierre please actually test this

npentrel (changes requested):
I still need to review the examples fully - @sguequierre please confirm once you have tested this thoroughly

docs/configure/triggers.md
sguequierre: I'm grabbing this from the scope doc:
image
if you think it should just be object fine to change it but I would assume there are cases where this could be a string or int, like lt 20. I guess all of the examples we have use keys and I'm not sure how to construct one without them either

npentrel: If you, based on the docs you've written, cannot construct a conditional trigger with this info, then this info will likely not help our readers either. So I think that's a good indication that this isn't good enough documentation.

It could be that this changes from when the scope document was written. If you don't know of an example, and we have now identified this potential issue, what have you done to investigate this? And if you haven't investigated further than that, please have a think what you could do to try to investigate this on your own. Once you've done that, if you cannot figure it out, I am happy to help, but in the first instance it's the authoring writer's job to understand and double check what they're writing about.

docs/configure/triggers.md
npentrel: You can use the Flask code on https://docs-test.viam.dev/3527/configure/triggers/ alongside `ngrok`. Install ngrok, which allows you to make your localhost web app accessible to the internet temporarily. You can start the "tunnel" with `ngrok http 8080` if the webapp is running on port 8080. Then the link it gives you you can put in the webhooks field

docs/configure/triggers.md
sguequierre: I just looked into all the triggers possible on my kitchen sink machine and I can't find anything that doesn't operate like a key value pair (+ besides option for no conditions). I also asked Roxy and she said that everything has keys for now, so I am going to proceed as such and edit these parts to treat the value as an object or nested object that provides a key-value pair for conditions.

docs/configure/triggers.md
npentrel: I did just test this and it does work.

docs/configure/triggers.md
npentrel: I think the no nesting example is actually an example for this but it's unclear what method this would be used with. i thought maybe voltage for power sensor would make sense but even there you have to specify a key ("volts") and it seems to possibly be usable for power sensors, for voltage for example.

docs/configure/triggers.md
sguequierre: thank you so much!

docs/configure/triggers.md
sguequierre: oh thank you! I tried a similar thing and it didn't send me any notifications so I'm glad you found this.

npentrel: I also think conditional data triggers do not work on Try Viam. Can you ask Roxy about this and otherwise ask in the octagon slack?

sguequierre: Asked her over dm, will let you know

PR #3528, opened 2024-10-04
DOCS-2972: Add DoCommand methods for xArms

sguequierre:
* add set_speed and set_acceleration

### DIFF
--- a/docs/components/arm/lite6.md
+++ b/docs/components/arm/lite6.md
@@ -102,11 +102,54 @@ See [the frame system service](/services/frame-system/) for more information on 
 
 {{< readfile "/static/include/components/test-control/arm-control.md" >}}
 
-### DoCommand
+### Additional commands
 
-Execute the model-specific commands `set_speed` and `set_acceleration` that are not otherwise defined by the component API.
+In addition to the [Arm API](/appendix/apis/components/arm/), the `lite6` arm supports some model-specific commands that allow you to set the speed and the acceleration of the arm.
+You can invoke these commands by passing the following JSON document to the [`DoCommand()`](/appendix/apis/components/arm/#docommand) method:
+
+```json
+{
+  "set_speed": 45.0,
+  "set_acceleration": 10.0
+}
+```
+
+| Key                | Type  | Description                                    |
+| ------------------ | ----- | ---------------------------------------------- |
+| `set_speed`        | float | Speed in degrees per second.                   |
+| `set_acceleration` | float | Acceleration in degrees per second per second. |
+
+For example:
 
 {{< tabs >}}
+{{% tab name="Python" %}}
+
+**Parameters:**
+
+- `command` (Mapping[[str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str), ValueTypes]) (required): The command to execute.
+- `timeout` ([float](https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex)) (optional): An option to set how long to wait (in seconds) before calling a time-out and closing the underlying RPC call.
+
+**Returns:**
+
+- (Mapping[[str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str), viam.utils.ValueTypes]): Result of the executed command.
+
+**Raises:**
+
+- (NotImplementedError): Raised if the Resource does not support arbitrary commands.
+
+**Example:**
+
+```python {class="line-numbers linkable-line-numbers"}
+command = {
+  "set_speed": 45.0, # Set speed to 45.0 degrees per second
+  "set_acceleration": 10.0 # Set acceleration to 10.0 degrees per second squared
+}
+result = await my_arm.do_command(command)
+```
+
+For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/viam/components/arm/client/index.html#viam.components.arm.client.ArmClient.do_command).
+
+{{% /tab %}}
 {{% tab name="Go" %}}
 
 **Parameters:**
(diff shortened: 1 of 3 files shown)


docs/components/arm/lite6.md
```diff
@@ -102,6 +102,41 @@ See [the frame system service](/services/frame-system/) for more information on
 
 {{< readfile "/static/include/components/test-control/arm-control.md" >}}
 
+### DoCommand
+
+Execute the model-specific commands `set_speed` and `set_acceleration` that are not otherwise defined by the component API.
+
+{{< tabs >}}
+{{% tab name="Go" %}}
```
sguequierre, line 110: @nfranczak want to confirm-- this is only implemented in the Go SDK? from what I could tell in the linked pr to the ticket there were no python or flutter changes

nfranczak: Correct this is only implemented in the Go SDK for only the xArm models.

nfranczak (approved):
🚀

docs/components/arm/lite6.md
npentrel: I don't think you two are talking about the same thing. I think Sierra wants to know if we can call the model-specific commands from Python and flutter code as well. Like this for python for example:

```python
command = { "set_speed": 45.0, "set_acceleration": 10.0}
result = await my_arm.do_command(command)
```

I'm reasonably certain that should be possible given that the language a model is implemented in broadly doesn't matter. You can write code in _any_ language when you're using it. Whether a method is implemented in a specific SDK only matters for the method (so it matters that `DoCommand` exists in Flutter and Go for the arm component) - the command passed to the DoCommand method will just be passed on and eventually be executed by the go model code. But that does not mean we can't use python code to invoke it.

@nfranczak please confirm that this isn't some weird edge case with how DoCommand works?

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,41 @@ See [the frame system service](/services/frame-system/) for more information on
 
 {{< readfile "/static/include/components/test-control/arm-control.md" >}}
 
+### DoCommand
+
+Execute the model-specific commands `set_speed` and `set_acceleration` that are not otherwise defined by the component API.
```
npentrel, line 107: This doesn't explain what this is or why it's here.

```suggestion
### Additional commands

In addition to the [Arm API](/appendix/apis/components/arm/), the `lite6` arm supports some model-specific commands that allow you to set the speed and the acceleration of the arm.

You can invoke these commands by passing the following JSON document to the [`DoCommand()`](/appendix/apis/components/arm/#docommand) method:

```json
{
    "set_speed": 45.0, // Set speed to 45.0 degrees per second
    "set_acceleration": 10.0  // Set acceleration to 10.0 degrees per second per second
}
\```

TODO(Add table of possible arguments)

```

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,41 @@ See [the frame system service](/services/frame-system/) for more information on
...
+
+{{< tabs >}}
```
npentrel, line 109: these tabs should go.

npentrel (changes requested):
see comments

docs/components/arm/lite6.md
nfranczak: > I think Sierra wants to know if we can call the model-specific commands from Python and flutter code as well.

I understand, thank you for the clarification.
I just tested out the doCommand on hardware using the Python SDK and it seems to work. I am guessing that it will also work through flutter then.

npentrel: Thank you for testing 😊

docs/components/arm/lite6.md
sguequierre: Is it ok to keep them since we are adding examples for the three sdks now?

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,109 @@ See [the frame system service](/services/frame-system/) for more information on
...
+
+**Returns:**
+
+- (Mapping[[str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str), viam.utils.ValueTypes]): Result of the executed command.
+
+**Raises:**
+
+- (NotImplementedError): Raised if the Resource does not support arbitrary commands.
+
+**Example:**
```
npentrel, line 140: All of this should go
```suggestion
```

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,109 @@ See [the frame system service](/services/frame-system/) for more information on
...
+command = {
+  "set_speed": 45.0,  # Set speed to 45.0 degrees per second
+  "set_acceleration": 10.0  # Set acceleration to 10.0 degrees
+                            # per second squared
+}
+result = await my_arm.do_command(command)
+```
+
+For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/viam/components/arm/client/index.html#viam.components.arm.client.ArmClient.do_command).
+
```
npentrel, line 152:
```suggestion
```

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,109 @@ See [the frame system service](/services/frame-system/) for more information on
...
+- `ctx` [(Context)](https://pkg.go.dev/context#Context): A Context carries a deadline, a cancellation signal, and other values across API boundaries.
+- `cmd` [(map[string]interface{})](https://go.dev/blog/maps): The command to execute.
+
+**Returns:**
+
+- [(map[string]interface{})](https://pkg.go.dev/builtin#string): The command response.
+- [(error)](https://pkg.go.dev/builtin#error): An error, if one occurred.
+
+**Example:**
+
```
npentrel, line 167:
```suggestion
```

docs/components/arm/lite6.md
```diff
@@ -102,6 +102,109 @@ See [the frame system service](/services/frame-system/) for more information on
...
+cmd := map[string]interface{}{
+    "set_speed": 45.0, // Set speed to 45.0 degrees per second
+    "set_acceleration": 10.0  // Set acceleration to 10.0 degrees per second squared
+}
+
+result, err := myArm.DoCommand(context.Background(), cmd)
+```
+
+For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/resource#Resource).
```
npentrel, line 179:
```suggestion
```

docs/components/arm/lite6.md
npentrel: We really should not be duplicating method docs here. Since you've already created the code samples keep them, but we probably shouldn't take the time to do them in the future

sguequierre: Ok, sorry for the misunderstanding

npentrel (approved):

npentrel:
Ready to merge @sguequierre ?

PR #3529, opened 2024-10-04
DOCS-2993: Add pipeline field object to transform camera json

sguequierre:


docs/components/camera/transform.md
```diff
@@ -181,14 +184,16 @@ The Crop transform crops takes an image and crops it to a rectangular area speci
 The Depth Edges transform creates a canny edge detector to detect edges on an input depth map.
 
 ```json {class="line-numbers linkable-line-numbers"}
-{
-  "type": "depth_edges",
-  "attributes": {
-    "high_threshold_pct": <float>,
-    "low_threshold_pct": <float>,
-    "blur_radius_px": <float>
+"pipeline": [
```
npentrel, line 187: I think this still would work if you posted it into the advanced field. So it's still confusing. I think we need to add the entire json component config object

sguequierre: adding source camera-- I think this is what you mean right? making it look like the attributes do in the advanced view when configured fully

docs/components/camera/transform.md
```diff
@@ -84,7 +84,7 @@ Then remove and fill in the attributes as applicable to your camera, according t
   "type": "camera",
   "namespace": "rdk",
   "attributes" : {
-    "source" : "<your-camera-name>",
+    "source" : "<your-source-camera-name>",
```
sguequierre, line 87: figured this was confusing with your-camera-name also being prompted above for the transform camera itself's name

npentrel (approved):
Looks good and I have filed https://viam.atlassian.net/browse/APP-6554 to ask for them to bring back the input field

PR #3534, opened 2024-10-07
DOCS-2704: Edit installation guide

sguequierre:
* suggestions from esha 
* make viam-micro-server and viam-server separate pages 
* make installation guide empty page
* edit wording on viam-micro-server installation steps slightly based off [this document](https://docs.google.com/document/d/1Lj6VsWlQbx9D99kd8QHcHvJlDHQS7vApznNnIdsmPqc/edit)


docs/installation/viam-server-setup.md
```diff
@@ -0,0 +1,125 @@
...
+---
+
+To use Viam, you need to install either the `viam-server` binary or `viam-micro-server`.
+
+[`viam-server`](/architecture/rdk/) is the binary built from the [Robot Development Kit](https://github.com/viamrobotics/rdk) that contains and manages communications between all Viam's built-in hardware drivers ({{< glossary_tooltip term_id="component" text="components" >}}) and software {{< glossary_tooltip term_id="service" text="services" >}}, connects your machine to the cloud, manages machine configuration, and manages dependencies including {{< glossary_tooltip term_id="module" text="modules" >}}.
+
+`viam-micro-server` is a lightweight version of `viam-server` which can run on resource-limited embedded systems that cannot run the fully-featured `viam-server`.
+To install `viam-micro-server`, see [`viam-micro-server` Installation Guide](/installation/viam-micro-server-setup/).
+
+For an overview of the Viam software platform, see [Learn about Viam](/get-started/).
```
EshaMaharishi, line 34: Really excited that these changes are going in! I just wanted to make one suggestion - I don't think it makes sense to mention viam-micro-server on this page which is otherwise about viam-server.

We could potentially keep the top-level "Installation Guide" page and just have it contain this text. (We could add a sentence at the end of the second pragraph "viam-server is the binary..." that links to this page, similar to the sentence at the end of the third paragraph "vim-micro-server is a lightweight...")

EshaMaharishi: Or we could just remove the part about viam-micro-server from here, I just saw it's repeated on the two viam-micro-server pages

npentrel: We can limit the explanation and really only mention the other pages. However, they do need to both mention each other. Because it's likely to be confusing for anyone unfamiliar with viam-server and viam-micro-server and they may end up on either page and then think we only support microcontrollers or SBCs. That's a reason to not separate the pages by the way. We may be hurting ourselves and I honestly still haven't decided whether this is the right thing to do here.

I think you were one of the people to point out that maybe people don't even click on the top level page. So I don't think it would be a good idea to put vital information for understanding on that page.

docs/installation/viam-server-setup.md
```diff
@@ -0,0 +1,125 @@
+---
+title: "viam-server Installation Guide"
+linkTitle: "viam-server setup"
+childTitleEndOverwrite: "Try Viam"
+weight: 9
+no_list: true
+type: docs
+images: ["/installation/thumbnails/install.png"]
+imageAlt: "Install Viam"
+description: "To use Viam, install and run viam-server or viam-micro-server on the computer or microcontroller that will run your machine and is connected to your hardware."
```
npentrel, line 10: remove viam-micro-server here

docs/installation/viam-server-setup.md
npentrel: after re-reading it just now, I do actually think this works as is

docs/installation/viam-server-setup.md
```diff
@@ -0,0 +1,125 @@
...
+WSL itself does not currently support exposing many types of Windows hardware to the embedded Linux kernel.
+This means that some hardware, such as a connected webcam, may not be available to `viam-server` with WSL, even though it is fully supported for native Linux systems.
+
+Although you cannot access all hardware if you run `viam-server` on your personal computer with WSL, you can [run code to control a machine](/sdks/#run-code) on your personal computer with WSL if you install `viam-server` on a single-board computer running Linux.
+
+{{% /alert %}}
+
+### Operating System installation
+
+`viam-server` can run on **any computer that runs a [supported operating system](#platform-requirements)**, including single-board computers (SBCs) running 64-bit Linux.
```
npentrel, line 59: I think the link back is otherwise confusing since the heading doesn't match
```suggestion
`viam-server` can run on **any computer that runs on a [supported platform](#platform-requirements)**, including single-board computers (SBCs) running 64-bit Linux.
```

docs/installation/viam-micro-server-setup.md
```diff
@@ -0,0 +1,217 @@
+---
+title: "viam-micro-server Installation Guide"
+linkTitle: "viam-micro-server quick setup"
```
npentrel, line 3:
```suggestion
linkTitle: "viam-micro-server setup"
```

docs/installation/viam-micro-server-setup.md
```diff
@@ -0,0 +1,217 @@
...
+childTitleEndOverwrite: "Try Viam"
```
npentrel, line 4:
```suggestion
```

docs/installation/viam-server-setup.md
npentrel: What do you both think about changing the titles to `Installation (Microcontrollers)` and `Installation`?

EshaMaharishi: Sure, that works - I think with that name change it makes sense to have a link to the microcontrollers page from the main installation page

sguequierre: Made this title change!

EshaMaharishi (approved):
Thanks for making these updates!

npentrel (approved):
LGTM, thanks!

PR #3566, opened 2024-10-14
DOCS-3001: Add physical assembly instructions to quickstarts from learn about viam

sguequierre:
I think these are the right former quickstarts? I checked the how-tos that were moved in the [new front page PR](https://github.com/viamrobotics/docs/commit/980f53b99841d8b3ade08a913a39f5b7ab1aa8f8#diff-ba86e80f5b7483551756f2e4c925571f360b5f945b0abe41b3f42ab3c1994863). Please LMK if I am mistaken.

### DIFF
--- a/docs/how-tos/collect-data.md
+++ b/docs/how-tos/collect-data.md
@@ -54,7 +54,7 @@ The rover already has `viam-server` installed and is configured with some compon
 
 Once you have borrowed a rover, go to the **CONFIGURE** tab of the machine, find the cameras and click on the **Test** panel at the bottom of each camera's configuration panel to test the camera stream.
 You should have a front-facing camera and an overhead view of your rover.
-Now you know what the rover can _perveive_.
+Now you know what the rover can _perceive_.
 
 If your rover is facing a wall, find the base configuration panel and click on its **Test** panel.
 Use the controls to drive your rover to a different location.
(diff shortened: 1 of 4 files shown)


docs/how-tos/collect-data.md
```diff
@@ -64,6 +64,8 @@ Now that you have seen that the cameras on your Try Viam rover work, **continue
 If you have a computer that can run `viam-server` but no physical sensor, you can use the [`viam:viam-sensor:telegrafsensor`](https://app.viam.com/module/viam/viam-telegraf-sensor) model which measures computer performance metrics.
 {{% /expand%}}
 
+Make sure to connect your sensor or webcam to your computer.
```
JessamyT, line 67: I think move this to the bullet about cameras and sensors, before the "no hardware" expander. If it's here, it feels like then the people with no hardware need to do it after they've already done it.

docs/how-tos/control-motor.md
```diff
@@ -51,6 +51,8 @@ If you do not have both a board and motor, install `viam-server` on your laptop
 
 ## Instructions
 
+Make sure to wire your motor to your board before starting.
+Power the board on if you want to test your machine while configuring it.
```
JessamyT, line 55: I think this makes just as much sense as putting it in requirements, but I'd suggest picking one or the other for all of these for consistency. Or is there a reason this one is here instead?

JessamyT:
Should https://docs.viam.com/how-tos/drive-rover/ be part of this PR?

sguequierre:
> Should https://docs.viam.com/how-tos/drive-rover/ be part of this PR?

I checked that one and found it already had the appropriate instructions on line 78.

JessamyT (approved):
Ah you're right sorry I missed it! LGTM

PR #3568, opened 2024-10-15
DOCS-2961: Add How-to path expanders

JessamyT:


docs/how-tos/_index.md
```diff
@@ -14,5 +14,79 @@ sitemap:
...
+<div class="search-panel__results card-container lozad">
+    <div id="hits" class="row-no-margin"></div>
+    <div id="pagination"></div>
+</div>
+
+<div id="how-to-paths">
+
+{{< how-to-expand title="Get started with Viam basics" tasks="4" level="BEGINNER-FRIENDLY" >}}
+{{< cards >}}
+{{% card link="/how-tos/drive-rover/" %}}
```
npentrel, line 41: can you add the no image attribute to all the how to cards please
```suggestion
{{% card link="/how-tos/drive-rover/" noimage="true" %}}
```

layouts/shortcodes/how-to-expand.html
npentrel, line 1: this can go away now, right?

JessamyT: No? static/include/how-to/howtopaths.html can but this shortcode is used.

npentrel (approved):

PR #3594, opened 2024-10-17
update commands in agent 'manual install' path

abe-winter:
## What changed
- updates to the 'manual install' path
## Why
Ran into trouble when trying it
## Manual test
Followed steps in rendered docs on a pi
## Questions
1. Do we actually need this path? Like is the install script not sufficient. (I think the answer is yes, but LMK if not).
2. Should I nudge people to the 'install viam-agent on live systems' link at the top of the manual install tab? For many users that is actually the better choice

### DIFF
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -175,10 +175,13 @@ Create a file called <FILE>viam-provisioning.json</FILE> with the following form
 `viam-agent` is a self-updating service manager that maintains the lifecycle for several Viam services and keeps them updated.
 If you intend to use `viam-agent` to keep your device's Viam software up-to-date or to use its provisioning plugin to let end users set up their own machines, you need to use `viam-agent` and install it on the machine before sending it to the user.
 
-You can [install `viam-agent` on live systems](/installation/viam-server-setup/) or preinstall it.
-The following instructions will preinstall it using a preinstall script that adds `viam-agent` to images.
+The following instructions will preinstall `viam-agent` into an image.
 
-**Only use the following method for offline pre-installs.**
+**Only use the following method for offline pre-installs with images. For live systems, follow the instructions on a machine's setup tab to [install `viam-server` with `viam-agent`](/installation/viam-server-setup/).**
+
+```suggestion
+**Only use the following method for offline pre-installs. For live systems, **
+```
 
 {{< alert title="Support notice" color="note" >}}
 Please note this script works only under POSIX (MacOS and Linux) at the moment.
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -179,10 +179,6 @@ The following instructions will preinstall `viam-agent` into an image.
 
 **Only use the following method for offline pre-installs with images. For live systems, follow the instructions on a machine's setup tab to [install `viam-server` with `viam-agent`](/installation/viam-server-setup/).**
 
-```suggestion
-**Only use the following method for offline pre-installs. For live systems, **
-```
-
 {{< alert title="Support notice" color="note" >}}
 Please note this script works only under POSIX (MacOS and Linux) at the moment.
 {{< /alert >}}
--- a/docs/how-tos/provision-setup.md
+++ b/docs/how-tos/provision-setup.md
@@ -176,7 +176,8 @@ Create a file called <FILE>viam-provisioning.json</FILE> with the following form
 If you intend to use `viam-agent` to keep your device's Viam software up-to-date or to use its provisioning plugin to let end users set up their own machines, you need to use `viam-agent` and install it on the machine before sending it to the user.
 
 You can [install `viam-agent` on live systems](/installation/viam-server-setup/) or preinstall it.
-The following instructions will preinstall it using a preinstall script that adds `viam-agent` to images:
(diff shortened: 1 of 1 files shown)


docs/how-tos/provision-setup.md
```diff
@@ -314,15 +314,15 @@ sudo ./preinstall.sh /path/to/rootfs
 {{< tablestep >}}
 **1. Create and download files**
 
-Run the following commands to create directories for the provisioning binaries, then download the binaries and make them executable:
+As root on the target device, run the following commands to create directories for the provisioning binaries, then download the binaries and make them executable:
```
Otterverse, line 317: So the issue is that these instructions are supposed to be when you're NOT on the target device. This is pre-install to a factory image that's not currently running. On a live device, the normal install script should just be used.

Otterverse (commented):
Mostly fine, in that it correccts the commands but the one noted issue is that it's not clear this should only be used for offline pre-installs.

Honestly, the manual steps may not even be needed. If the scripted tarball generation isn't enough for a particular client, they can talk to us about why, IMHO. Also, with the unification happening in agent, these instructions will change (and become simpler with only one binary involved) after that releases (monday.)

I'm fine either way though. They've been broken this long and no one else has said anything, so I'm fine with this merging as-is too if you prefer.

docs/how-tos/provision-setup.md
abe-winter: oh that makes my experience make more sense. this is the manual version of the **preinstall** flow

I'm guessing these commands are wrong if you're not on the target device bc of abs paths

abe-winter:
@Otterverse if your call is to remove this path entirely, and only support the preinstall script, I'm for it. I can make that change in this branch

Otterverse:
> @Otterverse if your call is to remove this 'manual install' path entirely, and only support the preinstall script, I'm for it. I can make that change in this branch

I think that makes sense. This was orginally deep-dive, last-resort stuff meant for other viam devs trying to understand the provisioning preinstall process. We should instead just focus on improving/fix the script if it ever fails to work for someone (or advanced devs could just read the script to figure out how to do the manual things if they absolutely had to.)

docs/how-tos/provision-setup.md
npentrel: So do any of these comments apply?

abe-winter: nope I'm totally wrong

but I think these would still fail in 'editing an sd card' context because of the mix of abs and relative paths

see discussion below about removing the 'manual install' tab entirely -- @npentrel are you okay with that change?

npentrel:
That sounds good to me, I'll make the change

abe-winter:
@npentrel thanks, all yours

docs/how-tos/provision-setup.md
```diff
@@ -154,7 +154,7 @@ if __name__ == '__main__':
 If you are using the captive portal, this step is optional.
 If you are using a mobile app, you must create a provisioning configuration file, specifying at least a `fragment_id`.
 
-Create a file called <FILE>viam-provisioning.json</FILE> with the following format and customize the [attributes](/fleet/provision/#configuration):
+Create a file called <FILE>/etc/viam-provisioning.json</FILE> with the following format and customize the [attributes](/fleet/provision/#configuration):
```
npentrel, line 157: Is this change still needed? @abe-winter and @Otterverse? I think I tested this and the script asks you to enter where the json is anyway, doesn't it?

abe-winter: guessing not -- would discard all my changes, I didn't understand the context for these instructions

npentrel:
```suggestion
Create a file called <FILE>viam-provisioning.json</FILE> with the following format and customize the [attributes](/fleet/provision/#configuration):
```

docs/how-tos/provision-setup.md
```diff
@@ -175,19 +175,16 @@ Create a file called <FILE>viam-provisioning.json</FILE> with the following form
...
 
-You can [install `viam-agent` on live systems](/installation/viam-server-setup/) or preinstall it:
-
-{{< tabs >}}
-{{% tab name="Preinstall on an SD card (or other image)" %}}
-
-Viam provides a preinstall script that works with images to add `viam-agent`.
+You can [install `viam-agent` on live systems](/installation/viam-server-setup/) or preinstall it.
+The following instructions will preinstall it using a preinstall script that adds `viam-agent` to images.
+Only use the following method for offline pre-installs.
```
npentrel, line 180:
```suggestion

**Only use the following method for offline pre-installs.**
```

npentrel:
Screenshot 2024-10-18 at 19 03 49

There are some caching issues so providing a screenshot for https://docs-test.viam.dev/3594/how-tos/provision-setup/index.html#install-viam-agent 

Does this add better context for the script @abe-winter ?

abe-winter:
yup makes sense now

I think the reason I got it wrong before was that I thought the 'manual install' path was for installing on *device* not on SD card

abe-winter:
I can't approve bc github thinks it's my PR, but LGTM

Otterverse (approved):
LGTM

npentrel:
Ok - thanks both! merging!

PR #3596, opened 2024-10-18
docs(service/base-rc): fix control_mode instructions

HipsterBrown:
While helping a Discord community member configure a base remote control service to use the input controller's joysticks to control their base, following the current documentation around the `control_mode` attribute for the service did not work. After digging into [the code](https://github.com/viamrobotics/rdk/blob/main/services/baseremotecontrol/builtin/builtin.go#L130-L142) for the service, it appears to be a spelling discrepancy between the RDK and the docs. The docs specify "joyStickControl" as the expected mode name, while the RDK is looking for "joystickControl". 

I'm choosing to fix this in the docs as the quicker solution, however this could be seen as a bug in the RDK instead. If the latter is the better long term solution, please let me know and I will open the corresponding PR in that repo. In any case, I will probably still propose some validation logic to be added to the base-rc service to help catch these types of mistakes in the future.

### DIFF
--- a/docs/services/base-rc/_index.md
+++ b/docs/services/base-rc/_index.md
@@ -19,12 +19,12 @@ This uses the [`input` API](/components/input-controller/#api) to make it easy t
 
 Add the base remote control service after configuring your machine with a base and input controller to control the linear and angular velocity of the base with the controller's button or joystick controls.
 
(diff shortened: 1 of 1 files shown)


npentrel:
Thank you will merge as is - It's possible the naming changed

npentrel (approved):

PR #3600, opened 2024-10-18
Update air-quality-fleet to use new preferred pm_2_5 key

zaporter-work:
I changed the `pm_2.5` key to `pm_2_5` for my module.

`pm_2.5` still works (it is not backwards-breaking), but `pm_2_5` is the preferred new way to access this key.

(periods in the keys are painful)


npentrel (approved):
Thanks!

PR #3614, opened 2024-10-22
DOCS-3024: Fix flutter code sample control motor

sguequierre:
* fixes sample to do what other samples do (set power to .35 percent, wait, stop)
* tested and confirmed it works


docs/how-tos/control-motor.md
```diff
@@ -312,7 +307,7 @@ class MotorScreen extends StatelessWidget {
 }
 ```
 
-This code creates a screen with a power widget to adjust the power and two buttons that, when pressed, call on the `goFor()` method to make the motor either go forwards 10 revolutions or go backwards 10 revolutions.
+This code creates a screen with a power widget to adjust the power and a button that, when pressed, calls on the `setPower()` method to set the power of the motor to 35% spinning the motor forwards, waits three seconds, and then stops the motor.
```
sguequierre, line 310:
```suggestion
This code creates a screen with a power widget to adjust the power and a button that, when pressed, calls on the `setPower()` method to spin the motor forwards at 35% power, waits three seconds to let it spin, and then stops the motor.
```

npentrel (approved):
awesome thanks!

PR #3623, opened 2024-10-23
DOCS-2001: Update viam-server page slightly

npentrel:


JessamyT (approved):
LGTM. Only opt thought is, should we mention RDK anywhere since it's still visible in GitHub repos etc?

JessamyT:
Title should be DOCS-2001 though

PR #3637, opened 2024-10-28
DOCS-2654: Allow users to determine when to reconfigure

npentrel:


docs/architecture/viam-server.md
```diff
@@ -57,6 +57,28 @@ This means you can add, modify, and remove a modular resource instance from a ru
...
+   "sensor_name" : string,
+   "maintenance_allowed_key" : string
+}
+```
+
+<!-- prettier-ignore -->
+| Attribute | Type | Required? | Description |
+| --------- | ---- | --------- | ----------- |
+| `sensor_name` | string | Optional | The name of the sensor that provides the information if it is safe to update a machine's configuration. |
+| `maintenance_allowed_key` | string | Optional | The key of the key value pair for the reading returned by the sensor. |
```
cheukt, line 80: both keys are required if you want to use the feature

docs/architecture/viam-server.md
```diff
@@ -57,6 +57,28 @@ This means you can add, modify, and remove a modular resource instance from a ru
...
+maintenance : {
+   "sensor_name" : string,
+   "maintenance_allowed_key" : string
+}
+```
+
+<!-- prettier-ignore -->
+| Attribute | Type | Required? | Description |
+| --------- | ---- | --------- | ----------- |
+| `sensor_name` | string | Optional | The name of the sensor that provides the information if it is safe to update a machine's configuration. |
```
cheukt, line 79: I don't know if we should be a bit more specific - a `sensor_name` here should be the full name in the form `rdk:component:sensor/sensor1`, `sensor1` by itself will not work

npentrel, line 79:
```suggestion
| `sensor_name` | string | **Required** | The full name of the sensor that provides the information if it is safe to update a machine's configuration. For example `rdk:component:sensor/sensor1`. |
```

docs/architecture/viam-server.md
npentrel, line 80:
```suggestion
| `maintenance_allowed_key` | string | **Required** | The key of the key value pair for the reading returned by the sensor. |
```

cheukt (approved):
LGTM

PR #3640, opened 2024-10-28
DOCS-2588: Document generate module CLI command

JessamyT:


npentrel (approved):

docs/cli.md
```diff
@@ -616,23 +616,25 @@ viam logout
...
 - Building a module locally and running it on a target device. Rebuilding & restarting if already running.
 
-See [Upload a custom module](/how-tos/create-module/#upload-your-module-to-the-modular-resource-registry) and [Update an existing module](/how-tos/manage-modules/#update-an-existing-module) for a detailed walkthrough of the `viam module` commands.
+See [Upload a custom module](/how-tos/upload-module/) and [Update an existing module](/how-tos/manage-modules/#update-an-existing-module) for a detailed walkthrough of the `viam module` commands.
 
 If you update and release your module as part of a continuous integration (CI) workflow, you can also
 [automatically upload new versions of your module on release](/how-tos/manage-modules/#update-an-existing-module-using-a-github-action) using a GitHub Action.
 
 ```sh {class="command-line" data-prompt="$"}
+viam module generate --resource-type=<component|service> --resource-subtype=<subtype>
```
njooma, line 637: the args are optional, and the preferred path is to actually not include them. i would remove them from this particular example

njooma (changes requested):

docs/cli.md
JessamyT: I believe this section is meant to illustrate all possible options, whereas the next section (line 652) is *examples*. I will add a note to the table below suggesting they not use the options. Also note that, in a different currently-open PR, we do guide people through creating a module using this command, and we do not use options.

docs/cli.md
```diff
@@ -704,6 +710,8 @@ viam module upload --version=1.0.0 --platform=darwin/arm64 packaged-module.tar.g
 | `--force` | Skip local validation of the packaged module, which may result in an unusable module if the contents of the packaged module are not correct | `upload` | Optional |
 | `--id` | The build ID to list or show logs for, as returned from `build start` | `build list`, `build logs` | Optional |
 | `--module` | The path to the [`meta.json` file](#the-metajson-file) for the custom module, if not in the current directory | `update`, `upload`, `build` | Optional |
+| `--resource-subtype` | The API to implement with the modular resource. For example, `motor`. | `generate` | Optional |
+| `--resource-type` | Whether the new resource is a component or a service. For example, `component`. | `generate` | Optional |
```
JessamyT, line 714:
```suggestion
| `--resource-subtype` | The API to implement with the modular resource. For example, `motor`. We recommend *not* using this option and instead following the prompts after running the command. | `generate` | Optional |
| `--resource-type` | Whether the new resource is a component or a service. For example, `component`. We recommend *not* using this option and instead following the prompts. | `generate` | Optional |
```

docs/cli.md
```diff
@@ -616,23 +616,25 @@ viam logout
 
 ### `module`
 
-The `module` command allows to you to manage custom {{< glossary_tooltip term_id="module" text="modules" >}}
+The `module` command allows to you to manage custom {{< glossary_tooltip term_id="module" text="modules" >}}.
```
npentrel, line 619:
```suggestion
The `module` command allows to you to work with {{< glossary_tooltip term_id="module" text="modules" >}}.
```

docs/cli.md
```diff
@@ -616,23 +616,25 @@ viam logout
...
 This includes:
 
-- Creating a new custom {{< glossary_tooltip term_id="resource" text="modular resource" >}}
+- Generating stub files for a new module
+- Creating metadata for a new custom {{< glossary_tooltip term_id="resource" text="modular resource" >}}
```
npentrel, line 623:
```suggestion
- Creating metadata for a {{< glossary_tooltip term_id="resource" text="modular resource" >}}
```

docs/cli.md
```diff
@@ -616,23 +616,25 @@ viam logout
...
 - Uploading a new module to the [Viam registry](https://app.viam.com/registry)
 - Uploading a new version of your module to the [Viam registry](https://app.viam.com/registry)
 - Updating an existing module in the Viam registry
 - Updating a module's metadata file based on models it provides
 - Building your module for different architectures using cloud runners
 - Building a module locally and running it on a target device. Rebuilding & restarting if already running.
 
-See [Upload a custom module](/how-tos/create-module/#upload-your-module-to-the-modular-resource-registry) and [Update an existing module](/how-tos/manage-modules/#update-an-existing-module) for a detailed walkthrough of the `viam module` commands.
+See [Upload a custom module](/how-tos/upload-module/) and [Update an existing module](/how-tos/manage-modules/#update-an-existing-module) for a detailed walkthrough of the `viam module` commands.
```
npentrel, line 631: It's not a walkthrough of the commands - more a walkthrough of the process the commands are used for
```suggestion
See [Upload a custom module](/how-tos/upload-module/) and [Update an existing module](/how-tos/manage-modules/#update-an-existing-module) for more information.
```

docs/cli.md
npentrel: We can make an exception if we're that opinionated about it  
```suggestion
viam module generate
```

docs/cli.md
```diff
@@ -683,6 +688,7 @@ viam module upload --version=1.0.0 --platform=darwin/arm64 packaged-module.tar.g
 <!-- prettier-ignore -->
 | Command option | Description | Positional arguments |
 | -------------- | ----------- | -------------------- |
+| `generate` | Auto-generate stub files for a new Python module by following prompts. | - |
```
npentrel, line 691: This will have Go soon, no need to specifiy Python here
```suggestion
| `generate` | Auto-generate stub files for a new module by following prompts. | - |
```

npentrel: In fact golang is already there and supported

docs/cli.md
JessamyT: Okay. I'll leave the notes I added in the table to explain this, unless you think that's overkill now that this is coming out.

docs/cli.md
JessamyT: Hooray!

docs/cli.md
njooma: Thanks. The CLI flags are still not fully baked yet so I don't want people relying on them yet

njooma (approved):

PR #3644, opened 2024-10-30
Update cli.md to fix incorrect command

bhaney:
The CLI `viam upload` command does not allow you to select a directory as a `path`. You must point to a specific file. In addition, the file should be zipped with tar, and have the .tar.gz extension


sguequierre (approved):
thanks bijan!

npentrel:
@bhaney @sguequierre added an additional fix because the descriptions for those arguments were swapped.

PR #3646, opened 2024-10-31
Update configure.md

npentrel:
https://docs-test.viam.dev/3646/how-tos/configure/

### DIFF
--- a/docs/how-tos/configure.md
+++ b/docs/how-tos/configure.md
@@ -55,7 +55,7 @@ Create a {{< glossary_tooltip term_id="machine" text="machine" >}} in any locati
 
 Follow the {{< glossary_tooltip term_id="setup" text="setup instructions" >}} on your machine page to install `viam-server`.
 If you are using a microcontroller, install `viam-micro-server` instead.
-Follow tge setup instructions until your machine connects and appear as **Live** in the Viam app.
+Follow the setup instructions until your machine connects and appear as **Live** in the Viam app.
 
 {{% /tablestep %}}
 {{% tablestep link="/configure/" %}}
@@ -96,7 +96,7 @@ If a component you want to use for your project is not natively supported, you c
     {{< /alert >}}
 
 4.  When you add a component, it will create a panel in the configuration builder tool. Fill in any required attributes, following the documentation for the specific model.
-    You can find this documentation by following the link to the model's documentation from the available resources clicking on the respective model in each component's documentation
+    You can find this documentation by following the link to the model's documentation from the available resources.
 
 5.  Click the **TEST** area of the configuration panel to test your component.
 6.  If any problems occur check the **LOGS** tab or check the component page for component-specific troubleshooting.
--- a/docs/how-tos/configure.md
+++ b/docs/how-tos/configure.md
@@ -19,7 +19,7 @@ date: "2024-08-02"
 cost: "0"
 ---
 
-You can get a smart machine running with Viam in just a few steps.
+You can get any device running with Viam in just a few steps.
 
 Viam's modular system of {{< glossary_tooltip term_id="component" text="components" >}} and {{< glossary_tooltip term_id="service" text="services" >}} means that you can start doing interesting things with your machine without writing much or any code.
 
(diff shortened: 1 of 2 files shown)


docs/how-tos/configure.md
```diff
@@ -77,25 +73,33 @@ Click on the **CONFIGURE** tab of your machine's page in the Viam app to navigat
...
+You will now create a configuration that describes any connected hardware as well as any software your device uses.
+The {{% glossary_tooltip term_id="component" text="_components_" %}} in your configuration can describe:
+
+- Physical hardware, such as arms, motors, or cameras.
+- Wrappers that provide additional functionality for other configured components, for example a failover motor that switches from one motor to another if the primary motor fails.
+- Software that maps to the API calls that you would use for existing hardware types. For example, a temperature sensor may provide temperature readings based on an online service instead of hardware readings. Another such type of sensor that is frequently used returns _readings_ confirming it is safe to sync data or reconfigure a machine.
+- Software that does not map well to the existing APIs, such as a switch that you'd like to be able to turn on an off. For these types of components, you use the generic component API.
```
npentrel, line 82: @IanWhalen I am rewriting this based on yesterday's discussion to see how we can trim/clarify this. While I've trimmed the steps, I think there is actually vital information missing. The physical hardware part is fairly obvious. The other three aren't. Especially the latter two are things that once people get it, get folks excited - based on our SEs.

We need to also consider teachable moments as we shorten content. At some point we do need to explain what's possible with the platform. I think we can skip over other parts of this but I'd assume this isn't one. Something to run by the SEs though.

docs/how-tos/configure.md
```diff
@@ -43,29 +43,25 @@ See [`viam-server` Platform requirements](/installation/viam-server-setup/#platf
...
 
-All of the software that runs your machine on a computer is packaged into a binary called `viam-server`.
-If you are using a microcontroller, use `viam-micro-server` instead.
-Install it on the computer controlling your smart machine by following the {{< glossary_tooltip term_id="setup" text="setup instructions" >}} in the [Viam app](https://app.viam.com/).
-
-You can install `viam-server` on your personal computer, or on a single-board computer (SBC).
-After following the setup instructions, your machine should connect and appear as **Live** on the Viam app.
+Follow the {{< glossary_tooltip term_id="setup" text="setup instructions" >}} on your machine page to install `viam-server`.
+If you are using a microcontroller, install `viam-micro-server` instead.
+Follow tge setup instructions until your machine connects and appear as **Live** in the Viam app.
```
sguequierre, line 58:
```suggestion
Follow the setup instructions until your machine connects and appear as **Live** in the Viam app.
```

docs/how-tos/configure.md
```diff
@@ -77,25 +73,33 @@ Click on the **CONFIGURE** tab of your machine's page in the Viam app to navigat
...
-3. You need to [_configure_](/configure/) your machine so that `viam-server` can interact with its hardware.
-   Use the configuration builder tool in the Viam app to create a file that describes what hardware you are using and how it is connected.
-   For example, if you have a DC motor, follow the [corresponding configuration instructions](/components/motor/gpio/) to tell the software which pins it is connected to.
-4. Add a suitable model to your machine on the **CONFIGURE** page:
+    {{< /alert >}}
 
-   - Click the + icon next to your machine part in the left-hand menu and select Component.
-   - Choose any component type (example: `camera`) and model (example: `webcam`). If a component you want to use for your project is not natively supported, you can [build your own modular resource](/how-tos/create-module/).
+4.  When you add a component, it will create a panel in the configuration builder tool. Fill in any required attributes, following the documentation for the specific model.
+    You can find this documentation by following the link to the model's documentation from the available resources clicking on the respective model in each component's documentation
```
sguequierre, line 99:
```suggestion
    You can find this documentation by following the link to the model's documentation from the available resources clicking on the respective model in each component's documentation.
```
This sentence seems odd- is it describing two ways or the same way? could trim down second half if same thing

sguequierre (changes requested):

docs/how-tos/configure.md
npentrel:
```suggestion
    You can find this documentation by following the link to the model's documentation from the available resources.
```

npentrel: good catch

docs/how-tos/configure.md
IanWhalen: i don't know if I want to think about this in terms of a diff.  I expect I would like everything to be written from scratch as much as possible given the new IA we come up with.

npentrel: ok - I will merge this for now, since it removes some weirdness and we can see how a rewrite turns out

PR #3649, opened 2024-11-01
DOCS-2705: Add username and env attributes for processes

npentrel:


docs/configure/processes.md
```diff
@@ -1,78 +1,55 @@
...
-
-Click **Save** in the upper right corner of the screen to save your config.
+| Attribute | Type | Required? | Description |
+| --------- | ---- | --------- | ----------- |
+| Executable (`name`) | string  | **Required** | The command you want to execute when your machine connects to the server. On many operating systems, you can find the executable path of commands by running `which <command-name>`. |
+| Arguments (`args`) | string  | Optional | Arguments to follow the command. |
+| Working directory (`cwd`) | string  | Optional | Where you want the process to execute. Defaults to the directory where `viam-server` executes. |
+| `username` (not available in builder mode) | string | Optional | Example: `"username": "ubuntu"`. |
```
npentrel, line 35: @zaporter-work I was not able to confirm that the `username` attribute does anything. My assumption is that it would allow us to run the process as a user?

On my pi I added a process with this config:
```
{
      "env": {
        "SVC_API_KEY": "VALUE"
      },
      "username": "pi",
      "id": "process-2",
      "name": "python",
      "args": [
        "username.py"
      ],
      "cwd": "/home/pi",
      "log": true,
      "one_shot": false
    }
```
and this little python program:

```
import os
import getpass
import time

while True:
  print(1)
  print(time.sleep(1))
  raise ValueError(getpass.getuser())
```

I can't see anytihng printed at all and what gets raised always indicates that the process runs as `root`. 

Does `username` do somethign else?

docs/configure/processes.md
```diff
@@ -1,78 +1,55 @@
...
+| `env` (not available in builder mode) | string | Optional | Environment variables for the process. Example: `"environment": { "SVC_API_KEY":"VALUE" }`. |
```
npentrel, line 36: @zaporter-work could you confirm for me how the config for a machine reaches the machine? That's all e2e encrypted right? I think we should probably mention that here.

docs/configure/processes.md
zaporter-work: Thank you so much for testing this! 

I have just confirmed that with the newest RDK, this now works.

A slightly easier way to test that this works is via the `whoami` command:
```json
  "processes": [
    {
      "username": "pi",
      "id": "process-1",
      "name": "whoami",
      "log": true
    }
  ],
```

docs/configure/processes.md
```diff
@@ -1,78 +1,55 @@
...
-| Execute once             | `one_shot`           | boolean | Optional     | Toggle whether to execute the command just once or keep running the process indefinitely.<ul><li>If true, the process executes once at `viam-server` startup. Until the process completes, `viam-server` startup is blocked and the robot appears offline in the [Viam app](https://app.viam.com), so this should only be used for quick processes.</li><li>If false, the process continues to run. If the process crashes, it automatically restarts. It does not block `viam-server`. Default: `false`.</li></ul> |
-
-Click **Save** in the upper right corner of the screen to save your config.
+| Attribute | Type | Required? | Description |
+| --------- | ---- | --------- | ----------- |
+| Executable (`name`) | string  | **Required** | The command you want to execute when your machine connects to the server. On many operating systems, you can find the executable path of commands by running `which <command-name>`. |
+| Arguments (`args`) | string  | Optional | Arguments to follow the command. |
```
zaporter-work, line 33:
```suggestion
| Arguments (`args`) | string[]  | Optional | Arguments to follow the command. |
```

docs/configure/processes.md
zaporter-work, line 35:
```suggestion
| `username` (not available in builder mode) | string | Optional | Run this process as a different user. Example: `"username": "ubuntu"`. |
```

docs/configure/processes.md
zaporter-work, line 36:
```suggestion
| `env` (not available in builder mode) | Map<string, string> | Optional | Environment variables for the process. Example: `"environment": { "SVC_API_KEY":"VALUE" }`. |
```

(or however you write nice types in docs)

docs/configure/processes.md
```diff
@@ -87,3 +64,17 @@ The corresponding JSON looks like this:
     }
   ]
 ```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+## Set up dependencies
```
zaporter-work, line 71: Seems better down here +1

zaporter-work (approved):
Really appreciate this. Thank you!

docs/configure/processes.md
```diff
@@ -1,78 +1,55 @@
...
-
-Click **Save** in the upper right corner of the screen to save your config.
+| Attribute | Type | Required? | Description |
+| --------- | ---- | --------- | ----------- |
+| Executable (`name`) | string  | **Required** | The command you want to execute when your machine connects to the server. On many operating systems, you can find the executable path of commands by running `which <command-name>`. |
+| Arguments (`args`) | string[]  | Optional | Arguments to follow the command. |
+| Working directory (`cwd`) | string  | Optional | Where you want the process to execute. Defaults to the directory where `viam-server` executes. |
+| `username` (not available in builder mode) | string | Optional | Example: `"username": "ubuntu"`. |
+| `env` (not available in builder mode) | Map<string, string> | Optional | Environment variables for the process. Example: `"environment": { "SVC_API_KEY":"VALUE" }`. |
```
npentrel, line 36:
```suggestion
| `env` (not available in builder mode) | Map<string, string> | Optional | Environment variables for the process. Environment variables are encrypted in transit but are stored in plain text in the configuration file that resides on the machine. Example: `"environment": { "SVC_API_KEY":"VALUE" }`. |
```

PR #3651, opened 2024-11-02
DOCS-2949: Fix mismapped methods

JessamyT:
- Fixes mixup of value vs name, ex. AnalogByName != analogReaderValue, same issue with interrupt.
- The actual proto names for "Write," "Read," and "Value" are WriteAnalog, ReadAnalogReader and GetDigitalInterruptValue, and using those consistently may help avoid the above confusion ^ in the future
- Don't merge until https://viaminc.slack.com/archives/C0653ULNNAX/p1730506244310229 is resolved
- Move GetGeometries further down the page because it's not board-specific and thus belongs at the bottom near the other inherited methods


.github/workflows/sdk_protos_map.csv
```diff
@@ -38,18 +38,18 @@ board,GetPWM,Yes,get_pwm,PWM,pwm
 board,SetPWM,Yes,set_pwm,SetPWM,setPwm
 board,PWMFrequency,Yes,get_pwm_frequency,PWMFreq,pwmFrequency
 board,SetPWMFrequency,Yes,set_pwm_frequency,SetPWMFreq,setPwmFrequency
-board,AnalogByName,No,analog_by_name,AnalogByName,analogReaderValue
-## HACK: Omitting PySDK: write_analog, currently borked: https://python.viam.dev/autoapi/viam/components/board/client/index.html#viam.components.board.client.BoardClient.write_analog
-board,Write,Yes,,Write,writeAnalog
-board,GetDigitalInterruptValue,No,digital_interrupt_by_name,DigitalInterruptByName,,
+board,GetDigitalInterruptValue,No,value,Value,digitalInterruptValue
+board,ReadAnalogReader,Yes,read,Read,analogReaderValue
```
npentrel, line 42: This is supported by micro-rdk? You're sure?

JessamyT: Was just going by what the old “read” entry said, but changing the proto name to the actual proto name

npentrel (approved):

PR #3677, opened 2024-11-15
Add log config for services

mcvella:
My understanding is that this works for any resource, not just components


npentrel (approved):

npentrel:
thanks!

PR #3692, opened 2024-11-19
DOCS-3130: Add MoveThroughJointPositions method to arm interface

JessamyT:


sguequierre (approved):
This LGTM, thanks for making a changelog entry! Did you check the SDKs too to see what content will be pulled in and if it could be improved?

JessamyT:
Yep I checked the SDKs and it seems only Go has implemented it so far (as reflected in the CSV), and does have an example, but I just noticed the example is mis-labeled there so fixing here https://github.com/viamrobotics/rdk/pull/4575

JessamyT:
@npentrel I figure the link to the method details can be changed once the automation creates something to change the link to. Would you rather I not add this to the changelog until then though?

PR #3702, opened 2024-11-26
DOCS-3162: Add disable wifi power save option

npentrel:
Hey Ale, can you confirm this is the only thing that changed?

https://docs-test.viam.dev/3702/configure/agent/#agent-provisioning


ale7714 (approved):
yes that's the only thing that change, thank you!

PR #3704, opened 2024-11-26
DOCS-3152: Clarify info about preinstall script

npentrel:


docs/how-tos/provision-setup.md
```diff
@@ -213,6 +213,7 @@ If you created a <FILE>viam-provisioning.json</FILE>, specify its location when
 
 You must have flashed a 64-bit image to your SD card and customized at least the hostname when prompted by the Raspberry Pi Imager.
 
+Any customizations for the image must be applied in the RPi Imager tool when creating the SD card.
```
npentrel, line 216: @Otterverse I don't know if this is even needed. Are there other customizations aside from what we say in the next line?

PR #3712, opened 2024-11-29
DOCS-3177: Clarify reconfiguration details

npentrel:
@cheukt could you confirm these details are correct? 

@hexbabe I should've probably asked before putting this up but why did you need to know this and is this something we think users will actually need to know?


hexbabe (commented):
I asked this since it might be important for module writers. It was important since different OAK cameras' reconfigures can sometimes depend on another OAK camera when both are on the same machine

hexbabe (approved):

npentrel:
Thanks!

PR #3730, opened 2024-12-04
DOCS-3103: Add Go to Hello World Module how-to

JessamyT:


docs/how-tos/hello-world-module.md
```diff
@@ -230,8 +258,87 @@ You need to add some sensor-specific code to support the sensor component.
...
+3. Edit the file structure:<br><br>
+
+   1. Change the name of <file>hello-world/models/module.go</file> to <file>hello-camera.go</file> and change the name of the <file>/models/</file> folder to <file>hellocamera</file> so you have <file>hello-world/hellocamera/hello-camera.go</file>.<br><br>
+
+   1. In the first line of <file>hello-camera.go</file>, change `package models` to `package hellocamera`.<br><br>
+
+   1. Change the name of <file>temporary/models/module.go</file> to <file>hello-sensor.go</file> and change the name of its folder to <file>hellosensor</file>.
+      Move the <file>hellosensor</file> folder from <file>temporary</file> to <file>hello-world</file> so you have <file>hello-world/hellosensor/hello-sensor.go</file>.<br><br>
+
+   1. In the first line of <file>hello-sensor.go</file>, change `package models` to `package hellosensor`.<br><br>
```
njooma, line 273: We probably shouldn't tell them to update the directory names, but just the file names. So the file structure should be `hello-world/models/hello-camera.go` and `hello-world/models/hello-sensor.go`, both with the package name `package models`.

docs/how-tos/hello-world-module.md
```diff
@@ -230,8 +258,87 @@ You need to add some sensor-specific code to support the sensor component.
...
+            return err
+        }
+
+        if err = helloWorld.AddModelFromRegistry(ctx, camera.API, hellocamera.HelloCamera); err != nil {
+            return err
+        }
+
+        if err = helloWorld.AddModelFromRegistry(ctx, sensor.API, hellosensor.HelloSensor); err != nil {
+            return err
+        }
```
njooma, line 310: Then here, we add `models.HelloCamera` and `models.HelloSensor`

njooma (changes requested):

docs/how-tos/hello-world-module.md
JessamyT: What will the generator do when multi-model modules are added to its capabilities? Will it put them all in one models folder?

docs/how-tos/hello-world-module.md
JessamyT: And we also change Config to sensorConfig, and we take out the declaration of `errUnimplemented = errors.New("unimplemented")` from hello-sensor.go

docs/how-tos/hello-world-module.md
JessamyT: I made all the necessary changes to get it to work but this way feels like more work for the user than making a directory per model, with the current way the files are generated. I think it'd be great if Config was named `mysensorConfig` or something by default.

docs/how-tos/hello-world-module.md
```diff
@@ -230,8 +258,82 @@ You need to add some sensor-specific code to support the sensor component.
...
+
+        <-ctx.Done()
+        return nil
+    }
+   ```
+
+   Save the file.<br><br>
+
+1. Change all instances of "temporary" in <file>hello-world/models/hello-sensor.go</file>:<br><br>
```
npentrel, line 320: I think it's just one instance?
```suggestion
1. Change all instances of `"temporary"` in <file>hello-world/models/hello-sensor.go</file>:<br><br>
```

docs/how-tos/hello-world-module.md
```diff
@@ -278,6 +380,8 @@ You need to add some sensor-specific code to support the sensor component.
 
 Edit the stub files to add the logic from your test script in a way that works with the camera and sensor APIs:
```
npentrel, line 381:
```suggestion
Edit the stub files to add the logic from your test script in a way that works with the camera and sensor APIs.
```

docs/how-tos/hello-world-module.md
```diff
@@ -390,13 +494,197 @@ You don't need to edit any of the validate or configuration methods because you'
...
+   ```go {class="line-numbers" data-start="33"}
+   type Config struct {
+       resource.AlwaysRebuild
+       ImagePath string `json:"image_path"`
+   }
+   ```
+
+   This adds the `image_path` attribute and causes the resource to rebuild each time the configuration is changed.
+
+1. Delete the `Reconfigure()` function entirely, since the camera will rebuild instead of reconfiguring.
```
npentrel, line 536: People at this point haven't encountered these concepts so I think we need to provide some context
```suggestion
1. Delete the `Reconfigure()` function.
   Some models use it to reconfigure attributes but in this case we'll have the camera rebuild instead of reconfiguring.
```

docs/how-tos/hello-world-module.md
```diff
@@ -390,13 +494,197 @@ You don't need to edit any of the validate or configuration methods because you'
...
+
+1. We are not providing a default image but rely on the end user to supply a valid path to an image when configuring the resource.
+   This means `image_path` is a required attribute.
+   Add the following code to the `Validate` function to throw an error if `image_path` isn't configured or isn't a string:
```
npentrel, line 540:
```suggestion
   Replace the `Validate` function with the following function to throw an error if `image_path` isn't configured or isn't a string:
```

docs/how-tos/hello-world-module.md
```diff
@@ -390,13 +494,197 @@ You don't need to edit any of the validate or configuration methods because you'
...
+   imgFile, err := os.Open(imagePath)
+   if err != nil {
+     return nil, camera.ImageMetadata{}, errors.New("Error opening image.")
+   }
+   defer imgFile.Close()
+   imgByte, err := ioutil.ReadFile(imagePath)
+   return imgByte, camera.ImageMetadata{}, nil
+   ```
+
+1. Delete the `SubscribeRTP` and `Unsubscribe` methods, since they are not applicable to this camera.
```
npentrel, line 576: does it matter if we leave them in?

docs/how-tos/hello-world-module.md
```diff
@@ -390,13 +494,197 @@ You don't need to edit any of the validate or configuration methods because you'
...
+{{< expand "Click if you are also creating a sensor component" >}}
+
+Now edit the sensor class definition to implement the sensor API.
+You don't need to edit any of the validate or configuration methods because you're not adding any configurable attributes for the sensor model.
+
+1. Add `"math/rand"` to the list of imports in <file>hello-sensor.go</file> for the random number generation.<br><br>
+
+1. Since `errUnimplemented` and `Config` are defined in <file>hello-camera.go</file>, you need to change <file>hello-sensor.go</file> to avoid redeclaring them:<br><br>
+
+   - Delete line 20, `errUnimplemented = errors.New("unimplemented")` from <file>hello-sensor.go</file>.<br><br>
```
npentrel, line 623: this is line 16 for me. I think it may be impractical to add the line numbers while things with the generator are still changing a lot

docs/how-tos/hello-world-module.md
```diff
@@ -406,7 +694,23 @@ With the implementation written, it's time to test your module locally:
...
 
-1. Make sure your machine's instance of `viam-server` is live and connected to the [Viam app](https://app.viam.com).
+{{% /tab %}}
+{{% tab name="Go" %}}
+
+1. Use Go to compile your module into a single executable:
+
+   - Navigate to your <file>hello-world</file> module directory in your terminal.
+   - Run `make` and then `make setup` to install dependencies.
+   - Run `go build` to compile your entry point (main program) file <file>main.go</file> and all other <file>.go</file> files in the directory, building your module and all dependencies into a single executable file.
```
npentrel, line 704: We can simplify this and make it copy pastable. I don't want to have to read through text to skim for commands I need to run.
```suggestion
1. From within the <file>hello-world</file> directory, compile your module into a single executable:

   ```sh {class="command-line" data-prompt="$" data-output="5-10"}
   make
   make setup
   go build
   ```
```

docs/how-tos/hello-world-module.md
```diff
@@ -406,7 +694,23 @@ With the implementation written, it's time to test your module locally:
...
+
+   Compiling your Go module also generates the `go.mod` and `go.sum` files that define dependency resolution in Go.
+
+   See the [Go compilation documentation](https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies) for more information.
```
npentrel, line 708: I think it's fine to assume that people here will know how to use the programming language they choose
```suggestion
```

docs/how-tos/hello-world-module.md
npentrel: Are you missing the step where you uncomment "resource.TriviallyReconfigurable" ? Because this alone does not work

docs/how-tos/hello-world-module.md
npentrel: Shouldn't we run 
```
make setup
make
```
It looks like the make command already runs go build?

npentrel (changes requested):
Overall good - I was able to get it to follow the instructions and get it to work mostly, a few issues. I think it's probably wise to remove the line numbers. That's probably wasted effort since they're likely to change still.

docs/how-tos/hello-world-module.md
JessamyT: It's marginally easier to delete than copy paste more lines

docs/how-tos/hello-world-module.md
JessamyT: I only got build errors when I ran go build, not make or make setup, so I used that to debug. Will ask SME.

docs/how-tos/hello-world-module.md
JessamyT, line 536:
```suggestion
```
Ooh actually can just delete this step

JessamyT: Oop I'll just delete this step.

docs/how-tos/hello-world-module.md
JessamyT:
```suggestion
   Replace the `Validate` function with the following code to throw an error if `image_path` isn't configured or isn't a string:
```
Okay yeah. Changing to "code" so as not to imply it's a totally different function.

docs/how-tos/hello-world-module.md
JessamyT: Oh you're right it should be 16 or 17 but it depends how many newlines they added to imports. Yes it's hard to maintain but it does make it much  easier to follow if it's right....and if it's off by only a line or two I think it'll be okay. I will change the *Config *sensorConfig to a search-replace but for the others since they're one-offs I don't think it makes sense to?

JessamyT, line 623:
```suggestion
   - Delete line 16, `errUnimplemented = errors.New("unimplemented")` from <file>hello-sensor.go</file>.<br><br>
```

docs/how-tos/hello-world-module.md
```diff
@@ -390,13 +494,197 @@ You don't need to edit any of the validate or configuration methods because you'
...
+   ```
+
+   ```go {class="line-numbers linkable-line-numbers" data-start="48" }
+   cfg    *sensorConfig
+   ```
+
+   ```go {class="line-numbers linkable-line-numbers" data-start="63" }
+   conf, err := resource.NativeConfig[*sensorConfig](rawConf)
+   ```
+
```
JessamyT, line 644:
```suggestion
   - Search for all instances of `*Config` in <file>hello-sensor.go</file> and change them to `*sensorConfig`.

```

docs/how-tos/hello-world-module.md
JessamyT:
```suggestion
1. From within the <file>hello-world</file> directory, compile your module into a single executable:
	
	   ```sh {class="command-line" data-prompt="$" data-output="5-10"}
	   make setup
	   make build
```

docs/how-tos/hello-world-module.md
JessamyT:
```suggestion
1. Change all instances of `temporary` in <file>hello-world/models/hello-sensor.go</file>:<br><br>
```

JessamyT:
>  I think it's probably wise to remove the line numbers. That's probably wasted effort since they're likely to change still.
@npentrel I am cognizant of this but, the module generator will not just change line numbers but will actually result in many of these steps being eliminated entirely, so the line numbers won't have to be maintained in the future. I've already put them there, and I think if they're close enough they're useful (and most if not all are correct at present).

docs/how-tos/hello-world-module.md
njooma: Likely when the generator can do multiple, it'll all be in the `models` directory

njooma (approved):

docs/how-tos/hello-world-module.md
npentrel:
```suggestion
1. Delete the `SubscribeRTP` and `Unsubscribe` methods. They are not applicable to this camera.
```

npentrel (approved):

PR #3736, opened 2024-12-06
DOCS-3147: Use resource-level logging for Python modules

JessamyT:


npentrel (approved):
This looks good to me - was that the only occurrance of using the global logger in our docs?

docs/how-tos/create-module.md
```diff
@@ -1119,7 +1119,24 @@ The example code shown above under [Write your new resource model definition](#w
...
+self.logger.debug("debug info")
+self.logger.info("info")
+self.logger.warn("warning info")
+self.logger.error("error info")
+self.logger.exception("error info", exc_info=True)
+self.logger.critical("critical info")
+```
+
+Resource-level logs are recommended instead of global logs for modular resources, because they make it easier to determine which component or service an error is coming from.
+Resource-level logs appear in the **ERROR LOGS** section of each resource's configuration card in the app.
```
JessamyT, line 1135:
```suggestion
Resource-level logs appear in the **ERROR LOGS** section of each resource's configuration card in the app.

{{% alert title="Note" color="note" %}}
In order to see resource-level debug logs when using your modular resource, you'll either need to run `viam-server` with the `-debug` option or [configure your machine or individual resource to display debug logs](/architecture/viam-server/#logging).
{{% /alert %}}
```

stuqdog (approved):
lgtm, thanks for updating these!

JessamyT:
> This looks good to me - was that the only occurrance of using the global logger in our docs?

This and the sensor module page which also needs updating were the only places where I see it for Python

docs/how-tos/create-module.md
```diff
@@ -1119,7 +1119,28 @@ The example code shown above under [Write your new resource model definition](#w
...
```
JessamyT, line 1135:
```suggestion
Resource-level error logs appear in the **ERROR LOGS** section of each resource's configuration card in the app.
```

npentrel (approved):

PR #3744, opened 2024-12-09
Add missing `viam module create` step in hello world module docs

Abalaxy:
Was going through the [Create a Hello World module](https://docs.viam.com/how-tos/hello-world-module/#package-and-upload-the-module) doc and discovered that the `viam module create` step was missing, but that it was present in the dedicated [Upload a module to the Viam Registry](https://docs.viam.com/how-tos/upload-module/#upload-a-module) doc. Added in that information, for both Python and Go, and updated the language in the existing steps to be consistent with the change.

**Screenshots:**
Screenshot 2024-12-09 at 5 25 31 PM
_Python_

Screenshot 2024-12-09 at 5 25 37 PM
_Go_

### DIFF
--- a/docs/how-tos/hello-world-module.md
+++ b/docs/how-tos/hello-world-module.md
@@ -766,11 +766,11 @@ To create the module in the registry, package your code (for Python), and upload
 
 1. Create the module in the registry, from within your <file>hello-world</file> directory, run the `viam module create` CLI command:
 
-    ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
-    viam module create --name <your-module-name> --org-id <your-org-id>
-    ```
+   ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
+   viam module create --name <your-module-name> --org-id <your-org-id>
+   ```
 
-    Get the `org-id` for your {{< glossary_tooltip term_id="organization" text="organization" >}} from your organization's **Settings** page in the [Viam app](https://app.viam.com/).
+   Get the `org-id` for your {{< glossary_tooltip term_id="organization" text="organization" >}} from your organization's **Settings** page in the [Viam app](https://app.viam.com/).
 
 2. Package the module as an archive, run the following command:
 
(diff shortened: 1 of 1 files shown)


michaellee1019:
Hey @Abalaxy I see there is `viam module generate` on https://docs.viam.com/how-tos/hello-world-module/#generate-stub-files. I think that is why there is no create below. Behind the scenes generate will also run create. Unless something isn't working right.

michaellee1019:
In my exploring I also saw there were two similar how-tos for creating a module. Should there be only one?

JessamyT:
My understanding is that using `Register module: Yes` when prompted by the CLI module generator would indeed register the module. If a user said `Register module: No`, that's when they'd need to run `viam module create`, but that's outside the scope of this particular doc.

> In my exploring I also saw there were two similar how-tos for creating a module. Should there be only one? https://docs.viam.com/how-tos/hello-world-module/ https://docs.viam.com/how-tos/create-module/

These two docs serve two different purposes: The hello world guide is more prescriptive and provides less information. It's also only for Python and Go. The more general create a module guide provides more details and generalized guidance, and includes C++ information. Currently, there's also the Create a Python Sensor Module guide, which has overlap as well. The docs IA is changing, and with those changes we'll be modifying some of this content.

Abalaxy:
> Hey @Abalaxy I see there is `viam module generate` instructions on https://docs.viam.com/how-tos/hello-world-module/#generate-stub-files. I think that is why there is no create below. Behind the scenes generate will also run create. Unless something isn't working right.

Aggghhh, I missed switching that to `Yes`! I had the "Generate the sensor code" section open as well and read that and conflated the two, because that section says to set it to `No` (which is correct for that call, but not the first one). Good catch, nevermind then 🤦

JessamyT:
I’m sorry! I’m realizing the instructions to set it to Yes may have been updated only after you first encountered this page, because I did indeed erroneously have it set to No in a previous version. You were actually onto something before my other PR merged. Thank you so much for your efforts in any case.

Abalaxy:
> I’m sorry! I’m realizing the instructions to set it to Yes may have been updated only after you first encountered this page, because I did indeed erroneously have it set to No in a previous version. You were actually onto something before my other PR merged. Thank you so much for your efforts in any case.

No worries at all! Now I know how to make changes to the docs, so it was a useful exercise!

PR #3751, opened 2024-12-10
Update transform.md

randhid:
remove debug attribute from transform


randhid:
You got there faster i see will close

PR #3769, opened 2024-12-16
Document new Rust 1.83 requirement and add a new troubleshooting hint

acmorrow:
- The change to document the new rust requirement is per the changes here: https://github.com/viamrobotics/micro-rdk/pull/354. Note that the micro-rdk PR says "1.82", but what was ultimately merged was in fact 1.83.
- The new troubleshooting tip is per https://viam.atlassian.net/browse/RSDK-9134, which has an open PR in the micro-rdk repo (see https://github.com/viamrobotics/micro-rdk/pull/364), so that should merge before this.


npentrel (approved):
Thank you Andrew!

PR #3773, opened 2024-12-19
DOCS-3168: Add two new features for crop transform camera

sguequierre:
* proportional cropping and overlay crop box


npentrel (approved):
LGTM - It may be worth considering taking all the possibel transformations out of the table (that is the classification, crop, depth edges etc) and instead having them be individual h3s with the respective content. One of the consistent pieces of feedback has been that the amount of tables makes the content harder to navigate.

PR #3774, opened 2024-12-20
Update the C++ module creation docs to reflect new data type

lia-viam:
Edits the module how-to to reflect the replacement of `AttributeMap` with `ProtoStruct`


JessamyT (approved):
Assuming this has been tested, go ahead and merge. TYSM for updating!

lia-viam:
@JessamyT what does testing entail? I checked the `docs-test` version and it looks good but not sure if there's other checks I missed

JessamyT:
@lia-viam I just mean from an eng perspective, if you're sure this code works go for it.

PR #3775, opened 2024-12-26
Remove some transform pipelines

randhid:
We have not tested most of the transform camera's pipelines in over two years, and some of them remain extremely low use. I'm being a bit cautious and removing these attributes in preparation for a full deprecation of these pipelines and perhaps this model - tbd.

The models removed are configured on fewer than 5 machines total over Viam's four-year span, and 0 configured machines in the last year.


sguequierre (approved):
thanks rand!

PR #3779, opened 2025-01-02
Update stepper_delay_usec for gpiostepper

michaellee1019:
When using the gpiostepper component I noticed that the attribute name was changed to `stepper_delay_usec`. Also wanted to add a note that without this attribute defined you cannot call SetPower and get an exception message saying that it needs to be set.

Review process:

- No need to request review from any specific docs team members; we will see your PR and one of us will review.
- If you need technical review from engineering, please request review from the relevant person.
- Docs team might commit styling nit fixes to save you the trouble :)
- You can merge after approval from one docs team member (as well as any necessary technical review).


npentrel (approved):
Thank you for fixing this @michaellee1019 !

PR #3829, opened 2025-01-09
DOCS-3273: Split out data export into data export binary and data export tabular

sguequierre:
* flags are different for each

### DIFF
--- a/docs/dev/reference/apis/data-client.md
+++ b/docs/dev/reference/apis/data-client.md
@@ -20,7 +20,7 @@ tags:
 aliases:
   - /program/apis/data-client/
   - /build/program/apis/data-client/
-  - /appendix/apis/data-client/  
+  - /appendix/apis/data-client/
 
 date: "2024-09-19"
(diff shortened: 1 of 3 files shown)


docs/dev/reference/apis/services/motion.md
```diff
@@ -9,7 +9,6 @@ images: ["/icons/components/arm.svg"]
 date: "2022-01-01"
 aliases:
   - /appendix/apis/services/motion/
-
```
sguequierre, line 12: prettier fix the check was catching

docs/dev/reference/apis/data-client.md
```diff
@@ -20,7 +20,7 @@ tags:
 aliases:
   - /program/apis/data-client/
   - /build/program/apis/data-client/
-  - /appendix/apis/data-client/  
+  - /appendix/apis/data-client/
```
sguequierre, line 24: prettier fix the check was catching

npentrel (approved):

PR #3838, opened 2025-01-10
Update mlmodel.md

bhaney:
Add a missing step where the user needs to convert a PIL image into a Viam image before feeding it into a get_detections method


sguequierre (approved):
thanks!

PR #3845, opened 2025-01-13
Add API key page

npentrel:


docs/operate/control/api-keys.md
```diff
@@ -0,0 +1,40 @@
...
+If you select location, your key grants access to all of the machines contained within that location.
+
+For **Role**, assign either an **Owner** or **Operator** role.
+See [Permissions](/manage/manage/rbac/) for information about the privilege each role entails at each resource level.
+
+## Change an API key's access
+
+To edit an API key, click on **Show details** in the key's row of the key table's **Resources** column.
+
+{{<imgproc src="/fleet/additional-details.png" resize="700x" declaredimensions=true alt="Additional details for a key">}}
```
JessamyT, line 33: Do we care that this shows the ID without a box over it? Not sure why that would matter without the key but just wondering how much we care about setting precedents and whether readers will wonder whether we meant to include it.

JessamyT (approved):
LGTM % Instead of adding this as a true new IA heading, another option would be putting it in the build reference section or in dev tools and linking to it where relevant. My thought is that "API keys" isn't a JTBD/imperative mood thing like all the other non-reference headings here (exception being SDKs which is just a link).

npentrel:
> LGTM % Instead of adding this as a true new IA heading, another option would be putting it in the build reference section or in dev tools and linking to it where relevant. My thought is that "API keys" isn't a JTBD/imperative mood thing like all the other non-reference headings here (exception being SDKs which is just a link).

Yes, I mentioned this option on slack, too. We can move it there if it feels like to much where it is now. I placed it with building apps as that is where it is relevant and it may be that that is where people would search for that information. If you want it to be in reference you can move it there, just look out for people who struggle to find that information, if that is the case move it back here. 

I will merge this for now to remove the broken links

docs/operate/control/api-keys.md
JessamyT: I meant a redaction box--a solid grey box over the ID. Fine to leave if you're okay with it; just wanted to point it out in case we don't want to do this.

JessamyT:
> Yes, I mentioned this option on slack, too. We can move it there if it feels like to much where it is now. I placed it with building apps as that is where it is relevant and it may be that that is where people would search for that information. If you want it to be in reference you can move it there, just look out for people who struggle to find that information, if that is the case move it back here.
> 
> I will merge this for now to remove the broken links (after fixing the remaining ones)

Ah that's right, you did--sorry I should've phrased differently. Sounds good--will leave where it is for now and will change if it becomes clear that it is one too many headings.

PR #3853, opened 2025-01-15
Clarify micrordk connection stability troubleshooting

JessamyT:
Also fixes typo because codespell was failing


stevebriskin (approved):

npentrel (approved):
LGTM but it seems like there's an unrelated codespell issue

PR #3858, opened 2025-01-16
DOCS-3341: Add missing python methods and fix check

npentrel:
Adding a variety of methods that have been added to the python SDK docs. Also fixing the check that checks whether there is anything in the python sdk docs we're not covering


sguequierre (approved):

sguequierre (approved):
thanks naomi!

PR #3864, opened 2025-01-17
Generate Netlify redirects file

npentrel:


config.toml
```diff
@@ -49,10 +49,18 @@ weight = 1
     ordered = false
     startLevel = 2
 
+disableAliases = true
```
npentrel, line 52:
```suggestion
disableAliases = false
```

npentrel: Let's do that part separately

PR #3873, opened 2025-01-20
DOCS-3234: Add CLI commands for billing service

npentrel:


sguequierre (approved):
Might help to add examples especially with --address

PR #3882, opened 2025-01-20
Use 404 plugin

npentrel:


npentrel:
It works - it detected a missing alias! 🥹

PR #3884, opened 2025-01-21
DOCS-3412: Show left nav for tutorials and hidden pages

npentrel:
This cleans up the sidenav for hidden pages and tutorials. Previously this would be either empty or not show the active path. Now it will show something sensible:

- For tutorials it will show a link to the tutorials page and the table of contents.
- For hidden pages, it will show the path to the page if we're on the page and hide it otherwise.


JessamyT (approved):
Staged site LGTM

PR #3918, opened 2025-01-27
DOCS-3453: Explicitly state there's no model available to mirror the camera image

sguequierre:
We got this q a couple times and the ai bot suggested the transform camera with no specific model so I figured it might be best to explicitly state it


docs/operate/reference/components/camera/camera-facts.md
```diff
@@ -0,0 +1,17 @@
...
+weight: 60
+type: "docs"
+description: "There's no model of transform camera available to mirror the camera image."
+images: ["/icons/components/camera.svg"]
+tags: ["camera", "components"]
+toc_hide: true
+hidden: true
+_build:
+  render: false
+  list: false
```
npentrel, line 13: what do these do? Are they needed?

sguequierre: Re. [slack thread[(https://viaminc.slack.com/archives/C055G9S1W3Y/p1738005179634089?thread_ts=1738003197.913079&cid=C055G9S1W3Y)-- they make the link a 404. I thought it was needed so someone couldn't accidentally go to the page just in case

npentrel: Sorry I needed to test this and just have. I don't see this generate a page for this now at all. So the 404 is coming from there just being no page here. Is there something I am missing? How is the AI supposed to now ingest this?

I think an alternative would be to create a shortcode that generates a div that has css on it to hide it that has the content. And then you'd just put the shortcode on the camera page. Then you don't need to have another page. But with the page approach the page needs to actually exist I think.

```
{{% hiddencontent %}}
There's no model of transform camera available to mirror the camera image.
{{% /hiddencontent%}}
```

npentrel (changes requested):
see comment

docs/operate/reference/components/camera/camera-facts.md
sguequierre: Thanks for the suggestion!

docs/operate/reference/components/camera/_index.md
```diff
@@ -96,6 +96,10 @@ For general configuration, development, and usage info, see:
 
 You can also use the camera component with the following services:
 
-- [Data management service](/data-ai/capture-data/capture-sync/: To capture and sync the camera's data
+- [Data management service](/data-ai/capture-data/capture-sync/): To capture and sync the camera's data
```
sguequierre, line 99: found a fixup

npentrel (approved):
Great!

PR #3922, opened 2025-01-27
DOCS-2671: Make first remotes example full example

sguequierre:
- ai was getting confused and thought that the first example was the full example, it was unclear


docs/data-ai/capture-data/capture-sync.md
```diff
@@ -472,7 +472,6 @@ To add them to your JSON configuration you must explicitly add the remote resour
 {{< expand "Click to view example JSON configuration for an ESP32 board" >}}
 
 The following example shows the configuration of the remote part, in this case an [ESP32 board](/operate/reference/components/board/esp32/).
-This config is just like that of a non-remote part; the remote connection is established by the main part (in the next expandable example).
```
sguequierre, line 475: find this sentence adds to confusion

JessamyT (commented):
I think this is still confusing--isn't the esp32 config meant to be just the remote side of the config, and then the second config example would be a Pi part that establishes the esp32 as a remote of it? in which case, the esp32 wouldn't have any remotes in its config?

PR #3935, opened 2025-01-28
DOCS-3466: Remove stream from hello world module guide

JessamyT:


sguequierre (commented):
Very rusty on modules but you might need to add `Image` now that that's been added to the interface?

JessamyT:
See lines above; image is implemented in the previous step. The diff in this case is just what isn't implemented.

sguequierre:
> See lines above; image is implemented in the previous step. The diff in this case is just what isn't implemented.

oh sorry I assumed it would be in the same step. should have checked more carefully.

sguequierre (approved):

JessamyT:
> > See lines above; image is implemented in the previous step. The diff in this case is just what isn't implemented.
> 
> oh sorry I assumed it would be in the same step. should have checked more carefully.

No worries--would rather you mention it than risk image falling through the cracks completely!

PR #3942, opened 2025-01-29
DOCS-3471: Update camera.md remove stream from Go camera interface

sguequierre:
* API has already been removed but thought this needed to go? double checking with sean

### DIFF
--- a/docs/dev/_index.md
+++ b/docs/dev/_index.md
@@ -478,9 +478,7 @@ for d in detections:
 ```go
 // Get image from camera stream on construction site
 myCamera, err := camera.FromRobot(machine, "construction-site-cam")
-camStream, err := myCamera.Stream(context.Background())
-img, release, err := camStream.Next(context.Background())
-defer release()
+img, err = camera.DecodeImageFromCamera(context.Background(), utils.MimeTypeJPEG, nil, myCamera)
 
 // Use machine learning model to gather information from the image
 visService, err := vision.FromRobot(machine, "hardhat_detector")
@@ -493,6 +491,8 @@ for i := 0; i < len(detections); i++ {
   }
 }
 ```
+
+Be sure to import `"go.viam.com/rdk/utils"` at the beginning of your file.
 
 {{% /tab %}}
 {{< /tabs >}}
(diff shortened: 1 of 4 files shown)


hexbabe (approved):
Nice! Are there any other docs changes that mention Stream and Stream.Next to get image?

sguequierre:
> Nice! Are there any other docs changes that mention Stream and Stream.Next to get image?

there are a few more, going through now-- thanks for reminding me! I had already changed on the vision and camera APIs but looking now there are a few more instances where this is used.

JessamyT (approved):
LGTM though I'm curious how the video feed on a camera test panel works and if we want to elude to that?

sguequierre:
> LGTM though I'm curious how the video feed on a camera test panel works and if we want to elude to that?

it seems to function just the same as before, although I am unsure how exactly it works. @hexbabe do you think we need to add any additional information here and if so what would it be?

hexbabe:
> > LGTM though I'm curious how the video feed on a camera test panel works and if we want to elude to that?
> 
> it seems to function just the same as before, although I am unsure how exactly it works. @hexbabe do you think we need to add any additional information here and if so what would it be?



> > LGTM though I'm curious how the video feed on a camera test panel works and if we want to elude to that?
> 
> it seems to function just the same as before, although I am unsure how exactly it works. @hexbabe do you think we need to add any additional information here and if so what would it be?

I think behavior on app should be exactly the same as before. If it isn't then it's a bug

JessamyT:
What I mean is that, if it's possible to stream video from a camera, should we be telling people how so they don't think they can only get static images and not video?

hexbabe:
> What I mean is that, if it's possible to stream video from a camera, should we be telling people how so they don't think they can only get static images and not video?

In the context of this doc change PR, it is not possible to stream video from a camera using our API

hexbabe:
The now-removed `Stream` is a bit of a misnomer which is a big reason why it was removed. TS and Flutter use `RenderFrame` and some webrtc magic to achieve true streams.

hexbabe:
Sorry I think I'm being unclear. For the *camera component* docs, I don't think we should mention streaming because streaming cannot be achieved with just camera API methods.

JessamyT:
Ooh ty for the explanation

PR #3952, opened 2025-01-31
DOCS-2695: Document improved bounding box annotation experience

sguequierre:
* add new gifs for labeling and tagging images
* remove third gif bc should be self explanatory


npentrel (approved):
Nice examples!

PR #3956, opened 2025-01-31
DOCS-3499: Improve capture edge data 

sguequierre:
Working with Jack/Devin

### DIFF
--- a/docs/data-ai/capture-data/capture-sync.md
+++ b/docs/data-ai/capture-data/capture-sync.md
@@ -56,6 +56,8 @@ The data management service writes data from your configured Viam resources to l
 - Data is synced to the Viam cloud at a configured sync interval using encrypted gRPC calls and deleted from the disk once synced.
 - You can capture and sync data independently, one can run without the other.
 
+For more information, see [How sync works](data-ai/capture-data/advanced/how-sync-works/).
+
 ## Configure the data management service for your machine
 
 {{< tabs >}}
@@ -656,20 +658,23 @@ The following components and services support data capture and cloud sync:
 
 {{< readfile "/static/include/data/capture-supported.md" >}}
 
-## Stop data capture
-
-If you don't need to capture data, for instance in a test scenario, you can turn off data capture to reduce unnecessary storage.
-Alternatively, see [advanced data capture and sync configurations](/data-ai/capture-data/capture-sync/#advanced-data-capture-and-sync-configurations) for other ways to control data usage, such as conditional sync or retention policies.
-
-In the **Data capture** section of your resource's configuration card, toggle the switch to **Off**.
-
-Click the **Save** button in the top right corner of the page to save your config.
-
 ## View captured data
 
 To view all the captured data you have access to, go to the [**DATA** tab](https://app.viam.com/data/view) where you can filter by location, type of data, and more.
 
 You can also access data from a resource or machine part menu.
+
+## Stop data capture
+
+If you don't need to capture data, for instance in a test scenario, you can turn off data capture to reduce unnecessary storage.
+Alternatively, see [advanced data capture and sync configurations](/data-ai/capture-data/capture-sync/#advanced-data-capture-and-sync-configurations) for other ways to control data usage, such as conditional sync or retention policies.
+
+To turn off data capture for a specific resource's capture method (for example, a camera component capturing via the `GetImage` capture method) navigate to the **Data capture** section of your resource's configuration card and toggle the configured capture method's switch to **Off**.
+You can also globally turn off data capture on the `data_manager` service configuration card by toggling the **Capturing** switch to **Off**.
+
+To turn off data sync, navigate to the `data_manager` service configuration card and toggle the **Syncing** switch to **Off**.
+
+Click the **Save** button in the top right corner of the page to save your config.
 
 ## Advanced data capture and sync configurations
 
@@ -758,7 +763,7 @@ See [Conditional cloud sync](/data-ai/capture-data/conditional-sync/) for how to
 
 Configure how long your synced data remains stored in the cloud:
 
-- **Retain data up to a certain size or for a specific length of time:** Set `retention_policies` at the resource level.
+- **Retain data up to a certain size (for example, 100GB) or for a specific length of time (for example, 14 days):** Set `retention_policies` at the resource level.
   See the `retention_policy` field in [data capture configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-capture-attributes).
 - **Delete data captured by a machine when you delete the machine:** Control whether your cloud data is deleted when a machine or machine part is removed.
   See the `delete_data_on_part_deletion` field in the [data management service configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-management-attributes).
--- a/docs/data-ai/capture-data/capture-sync.md
+++ b/docs/data-ai/capture-data/capture-sync.md
@@ -669,7 +669,7 @@ You can also access data from a resource or machine part menu.
 If you don't need to capture data, for instance in a test scenario, you can turn off data capture to reduce unnecessary storage.
 Alternatively, see [advanced data capture and sync configurations](/data-ai/capture-data/capture-sync/#advanced-data-capture-and-sync-configurations) for other ways to control data usage, such as conditional sync or retention policies.
 
-To turn off data capture for a specific resource's capture method (for example, a camera component capturing via the `GetImage` capture method) navigate to the **Data capture** section of your resource's configuration card and toggle the configured capture method's switch to **Off**.
+To turn off data capture for a specific resource's capture method (for example, a camera component capturing through the `GetImage` capture method) navigate to the **Data capture** section of your resource's configuration card and toggle the configured capture method's switch to **Off**.
 You can also globally turn off data capture on the `data_manager` service configuration card by toggling the **Capturing** switch to **Off**.
 
 To turn off data sync, navigate to the `data_manager` service configuration card and toggle the **Syncing** switch to **Off**.
--- a/docs/data-ai/capture-data/capture-sync.md
+++ b/docs/data-ai/capture-data/capture-sync.md
@@ -43,9 +43,27 @@ aliases:
 You can use the data management service to capture data from [supported components and services](#supported-resources), then sync it to the cloud.
 You can also sync data from arbitrary folders on your machine.
 
-## Configure the data management service
+## Overview
 
-To start, configure the data management service to capture and sync data.
+Data capture and sync in Viam involves two key pieces:
+
+- A data management {{< glossary_tooltip term_id="service" text="service" >}} that writes captured data to local edge device storage and syncs that data with the cloud.
+- Individual {{< glossary_tooltip term_id="resource" text="resource" >}} configurations that specify what data to capture and how often.
+
+## How data capture and data sync works
+
+- The data management service stores captured data locally in <file>~/.viam/capture<file> by default.
+- Data is synced to the Viam cloud at a configured sync interval using encrypted gRPC calls and deleted from the disk once synced.
+- You can capture and sync data independently, one can run without the other.
+
+## Set up data capture and data sync
+
+### Configure a data management service for your machine
+
+The data management service is responsible for:
+
+- Writing data from your configured Viam resources to local storage on your edge device
+- Syncing data from the edge device to the cloud
 
 {{< tabs >}}
 {{% tab name="Config Builder" %}}
@@ -117,6 +135,8 @@ Click the **Save** button in the top right corner of the page to save your confi
 
 The following attributes are available for the data management service:
 
+{{< expand "Click to view data management attributes" >}}
+
 <!-- prettier-ignore -->
 | Name               | Type   | Required? | Description | `viam-micro-server` Support |
 | ------------------ | ------ | --------- | ----------- | ------------------- |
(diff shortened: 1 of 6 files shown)


docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,27 @@ aliases:
...
-## Configure the data management service
+## Overview
 
-To start, configure the data management service to capture and sync data.
+Data capture and sync in Viam involves two key pieces:
+
+- A data management {{< glossary_tooltip term_id="service" text="service" >}} that writes captured data to local edge device storage and syncs that data with the cloud.
+- Individual {{< glossary_tooltip term_id="resource" text="resource" >}} configurations that specify what data to capture and how often.
+
+## How data capture and data sync works
```
npentrel, line 53: Please make sure there is at least an intro sentence when you have bullet points. 

Worth considering if the explanation of how it works is actually what's needed this early in the page. Do we expect people to come here primarily for setup or for explanation?

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,27 @@ aliases:
 You can use the data management service to capture data from [supported components and services](#supported-resources), then sync it to the cloud.
 You can also sync data from arbitrary folders on your machine.
 
-## Configure the data management service
+## Overview
```
npentrel, line 46: We don't use headings like this anywhere in the docs. It's an explicit choice to not do this and instead place this type of content elsewhere - often at the top of the page, here it should go under the headline `Set up data capture and data sync` where there's no content before the next heading.

```suggestion
```

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,27 @@ aliases:
...
+- A data management {{< glossary_tooltip term_id="service" text="service" >}} that writes captured data to local edge device storage and syncs that data with the cloud.
+- Individual {{< glossary_tooltip term_id="resource" text="resource" >}} configurations that specify what data to capture and how often.
+
+## How data capture and data sync works
+
+- The data management service stores captured data locally in <file>~/.viam/capture<file> by default.
+- Data is synced to the Viam cloud at a configured sync interval using encrypted gRPC calls and deleted from the disk once synced.
+- You can capture and sync data independently, one can run without the other.
+
+## Set up data capture and data sync
```
npentrel, line 59: Headings should never be followed by another heading like this without text in between. This generally indicates a content design issue. 

In this case I wouldn't be surprised it you'd also find that people accidentally skip the step to add the data mgmt service. There is nothing helping the reader to avoid this from happening so when they're just skimming, they could easily miss it.

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,27 @@ aliases:
...
+
+### Configure a data management service for your machine
+
+The data management service is responsible for:
+
+- Writing data from your configured Viam resources to local storage on your edge device
+- Syncing data from the edge device to the cloud
```
npentrel, line 66: This is explanatory content that you've put with a heading where readers would expect pragmatic content. This info should likely be with the "How data capture and data sync works" section.

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,27 @@ aliases:
 You can use the data management service to capture data from [supported components and services](#supported-resources), then sync it to the cloud.
 You can also sync data from arbitrary folders on your machine.
 
-## Configure the data management service
+## Overview
 
-To start, configure the data management service to capture and sync data.
+Data capture and sync in Viam involves two key pieces:
+
+- A data management {{< glossary_tooltip term_id="service" text="service" >}} that writes captured data to local edge device storage and syncs that data with the cloud.
```
npentrel, line 50: Unless this has changed, there can only be one data management service, that's why it should be a definite article.

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -135,74 +155,16 @@ The following attributes are available for the data management service:
...
-  This can happen in cases such as MongoDB being inaccessible to `viam-server` or writes timing out.
-- Capturing directly to MongoDB may reduce the maximum frequency that data capture can capture data due to the added latency of writing to MongoDB.
-  If your use case needs to support very high capture rates, this feature may not be appropriate.
+### Configure data capture for individual resources
 
-{{< /alert >}}
+You can capture data for any {{< glossary_tooltip term_id="resource" text="resource" >}} that supports it, including resources on {{< glossary_tooltip term_id="remote-part" text="remote parts" >}}.
 
-## Configure data capture
+Configure data capture for individual resources by:
```
npentrel, line 164: I think you need to add where they'd do this though.

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -705,25 +678,101 @@ To view all the captured data you have access to, go to the [**DATA** tab](https
 
 You can also access data from a resource or machine part menu.
 
-## Supported resources
+## Advanced data capture and sync configurations
 
-The following components and services support data capture and cloud sync:
+### Direct to MongoDB capture
```
npentrel, line 683: I don't understand this headline - maybe "Capture directly to MongoDB"?

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -705,25 +678,101 @@ To view all the captured data you have access to, go to the [**DATA** tab](https
...
 
-{{< readfile "/static/include/data/capture-supported.md" >}}
+You can configure direct capture of tabular data to a MongoDB instance alongside disk storage on your edge device.
+This can be useful for powering real-time dashboards before data is synced from the edge to the cloud.
+
+Configure using `mongo_capture_config` attributes in your data manager service.
```
npentrel, line 688: grammar?

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -705,25 +678,101 @@ To view all the captured data you have access to, go to the [**DATA** tab](https
...
+- Capturing directly to MongoDB does not retry failed writes to MongoDB. As a consequence, it is NOT guaranteed all data captured will be written to MongoDB.
+  This can happen in cases such as MongoDB being inaccessible to `viam-server` or writes timing out.
+- Capturing directly to MongoDB may reduce the maximum frequency that data capture can capture data due to the added latency of writing to MongoDB.
+  If your use case needs to support very high capture rates, this feature may not be appropriate.
+
+{{< /alert >}}
+
+### Conditional sync
+
+Although data sync checks for new data at the configured interval, you can define additional conditions that must be met to trigger syncing.
```
npentrel, line 754: `Although` doesn't make sense here. There's no contradiction in that - rather it works on top of this: 

```suggestion
By default, `viam-server` checks for new data to sync at the configured interval (`sync_interval_mins `).
You can additionally configure sync to only happen when certain conditions are met.
```

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -705,25 +678,101 @@ To view all the captured data you have access to, go to the [**DATA** tab](https
...
-- **Pausing sync**: You can pause cloud sync at any time by navigating to your machine's **CONFIGURE** tab and disabling **Syncing** for your data management service.
+Configure how long your synced data remains stored in the cloud:
 
-  If you have captured data that you do not want to sync, delete the data on the machine before resuming cloud sync.
-  To delete the data locally, `ssh` into your machine and delete the data in the directory where you capture data.
+- **Retain data up to a certain size or for a specific length of time:** Set `retention_policies` at the resource level.
+  - See the `retention_policy` field in [data capture configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-capture-attributes).
+- **Delete data captured by a machine when you delete the machine:** Control whether your cloud data is deleted when a machine or machine part is removed.
+  - See the `delete_data_on_part_deletion` field in the [data management service configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-management-attributes).
```
npentrel, line 770: This will look a bit better if you remove the additional level or nesting. However you could also consider using the exact same pattern as the one in the Sync optimization content and use no bullets.
```suggestion
- **Retain data up to a certain size or for a specific length of time:** Set `retention_policies` at the resource level.
  See the `retention_policy` field in [data capture configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-capture-attributes).
- **Delete data captured by a machine when you delete the machine:** Control whether your cloud data is deleted when a machine or machine part is removed.
    See the `delete_data_on_part_deletion` field in the [data management service configuration attributes](/data-ai/capture-data/capture-sync/#click-to-view-data-management-attributes).
```

npentrel (changes requested):

docs/data-ai/capture-data/capture-sync.md
sguequierre: good point

docs/data-ai/capture-data/capture-sync.md
sguequierre: ok!

docs/data-ai/capture-data/capture-sync.md
sguequierre, line 688:
```suggestion
Configure using the `mongo_capture_config` attributes in your data manager service.
```

docs/data-ai/capture-data/capture-sync.md
sguequierre: IMO the old page was very setup oriented so I think more of an explanation is worth trying and presenting to eliot

docs/data-ai/capture-data/capture-sync.md
sguequierre: I think those instructions are right below in the tabs. Do we need to repeat them twice?

docs/data-ai/capture-data/capture-sync.md
sguequierre: thank you for the explanation!

docs/data-ai/capture-data/capture-sync.md
sguequierre: still adding

docs/data-ai/capture-data/capture-sync.md
jdamon96: I don't view them as mutually exclusive; imo a quick overview / explanation of the system before getting into specific set up steps helps contextualize the set up work

docs/data-ai/capture-data/capture-sync.md
npentrel: thx

docs/data-ai/capture-data/capture-sync.md
npentrel: Sounds good. It's always a trade off between context and pushing the content people may be trying to get to below the fold. If you both in the future want to check data on this you can always look at clarity data.

only other consideration is whether you want to link to the more detailed how capture & sync works page. Please feel free to resolve once you've considered this

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -691,9 +650,16 @@ You can also control how local data is deleted if your machine's local storage b
...
+
+The following components and services support data capture and cloud sync:
+
+{{< readfile "/static/include/data/capture-supported.md" >}}
+
 ## Stop data capture
 
-If this is a test project, make sure you stop data capture to avoid charges for a large amount of unwanted data.
+If you don't need to capture data, for instance in a test scenario, you can turn off data capture to reduce unnecessary storage.
+Alternatively, see [advanced data capture and sync configurations](/data-ai/capture-data/capture-sync/#advanced-data-capture-and-sync-configurations) for other ways to control data usage, such as conditional sync or retention policies.
```
npentrel, line 662: nice

docs/data-ai/capture-data/capture-sync.md
jdamon96: tiny nit - now that we've added the 'the', it should be _attribute_ (singular) rather than attribute**s**

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -43,9 +43,22 @@ aliases:
...
+
+## How data capture and data sync works
+
+The data management service writes data from your configured Viam resources to local storage on your edge device and syncs data from the edge device to the cloud:
+
+- The data management service stores captured data locally in <file>~/.viam/capture<file> by default.
+- Data is synced to the Viam cloud at a configured sync interval using encrypted gRPC calls and deleted from the disk once synced.
+- You can capture and sync data independently, one can run without the other.
+
+For more information, see [How sync works](data-ai/capture-data/advanced/how-sync-works/).
```
sguequierre, line 59:
```suggestion
For more information, see [How sync works](/data-ai/capture-data/advanced/how-sync-works/).
```

docs/data-ai/capture-data/capture-sync.md
sguequierre: @jdamon96 did you want to keep this section above? I'm in favor of keeping it along with the text you added below

docs/data-ai/capture-data/capture-sync.md
```diff
@@ -691,39 +652,126 @@ You can also control how local data is deleted if your machine's local storage b
...
 
-## Supported resources
+## Stop data capture
 
-The following components and services support data capture and cloud sync:
+If you don't need to capture data, for instance in a test scenario, you can turn off data capture to reduce unnecessary storage.
+Alternatively, see [advanced data capture and sync configurations](/data-ai/capture-data/capture-sync/#advanced-data-capture-and-sync-configurations) for other ways to control data usage, such as conditional sync or retention policies.
 
-{{< readfile "/static/include/data/capture-supported.md" >}}
+To turn off data capture for a specific resource's capture method (for example, a camera component capturing via the `GetImage` capture method) navigate to the **Data capture** section of your resource's configuration card and toggle the configured capture method's switch to **Off**.
```
sguequierre, line 672:
```suggestion
To turn off data capture for a specific resource's capture method (for example, a camera component capturing through the `GetImage` capture method) navigate to the **Data capture** section of your resource's configuration card and toggle the configured capture method's switch to **Off**.
```

npentrel (approved):
LGTM from my end. If there are more major changes, please re-tag

docs/data-ai/capture-data/capture-sync.md
jdamon96: yes - along w/ the text added - agree!

PR #3957, opened 2025-01-31
DOCS-3481: Add billing fragment

npentrel:
@RoxyFarhad and @adithyagurunathan is this how you think this should work? https://deploy-preview-3957--viam-docs.netlify.app/manage/manage/white-labelled-billing/#set-custom-pricing You add a billing fragment and then add the fragment to another fragment that is used for provisioning?


RoxyFarhad:
> You add a billing fragment and then add the fragment to another fragment that is used for provisioning?

It should just be included in the original fragment that you are adding to the machine, no need to double nest i I am understanding what you are saying correctly?

npentrel:
> > You add a billing fragment and then add the fragment to another fragment that is used for provisioning?
> 
> It should just be included in the original fragment that you are adding to the machine, no need to double nest i I am understanding what you are saying correctly?

Wasn't sure whether we wanted to recommend separating that info from other config info but if there's no need I can change that.

npentrel:
Updated @RoxyFarhad !

adithyagurunathan (approved):

PR #3959, opened 2025-01-31
DOCS-3489: Add windows to setup page

npentrel:


docs/operate/get-started/setup.md
```diff
@@ -40,6 +40,9 @@ Viam can run on any computer that runs one of the following operating systems:
 
 - Linux 64-bit operating systems running on AArch64 (ARM64) or x86-64 architectures
 - macOS
+- Windows
```
npentrel, line 43: @abe-winter I think this is all that's needed unless you want anything else covered?

abe-winter: yes lgtm -- I think we will update this guidance as folks play around with windows support, but this feels like the answer today afaik

abe-winter (approved):

PR #3960, opened 2025-02-01
DOCS-3106: Document CLI usability improvements

JessamyT:


npentrel (changes requested):
This doesn't seem to address the full scope of the work done here. Part of the scope was to be consistent for flags. That means we should use one flag consistently in the docs (`--org` possibly for the orgs set) and document the aliases as options.

purplenicole730 (commented):
This generally looks good to me, waiting on the changes Naomi asked for before I approve.
I would also ask @stuqdog for a review, since he introduced profiles and made the machine/part flag changes

docs/dev/tools/cli.md
```diff
@@ -1159,6 +1159,59 @@ viam packages export --org-id=123 --name=MyMLModel --version=latest --type=ml_mo
...
+The `profiles` command allows you to manage different CLI authentication profiles, so you can easily switch between API key authentications (for example authentication to one organization versus another).
+
+```sh {class="command-line" data-prompt="$"}
+viam profiles add --profile-name=<name-of-profile-to-add> --key-id=<API-key-ID> --key=<API-key>
+viam profiles update --profile-name=<name-of-profile-to-update> --key-id=<API-key-ID> --key=<API-key>
+viam profiles list
+viam profiles remove --profile-name=<name-of-profile-to-remove>
+```
+
+Examples:
```
stuqdog, line 1173: These examples are great for showing how to create and store profiles, but don't highlight how to actually use these profiles. Might be useful to have an example of something like `viam --profile=<name-of-profile-to-use> machines list`?

docs/dev/tools/cli.md
```diff
@@ -1191,45 +1244,42 @@ Examples:
 viam machines list
 
 # get machine status
-viam machines status  --machine=123 --organization="Robot's Org" --location=myLoc
+viam machines status  --machine=123 --location=myLoc
```
stuqdog, line 1247: I'm a little unclear about why we don't include `organization` anymore here, but we include it immediately above? Is the goal for one to be all flags and the other to be only mandatory flags? If so, we can probably get rid of `location` here as well.

docs/dev/tools/cli.md
```diff
@@ -1270,8 +1320,8 @@ viam machine part cp --machine=123 --part=123 -r -p machine:my_dir machine:my_fi
 <!-- prettier-ignore -->
 | Argument | Description | Applicable commands | Required? |
 | -------- | ----------- | ------------------- | --------- |
-| `--organization` | Organization name or ID that the machine belongs to | `list`, `status`, `logs`, `part` | **Required** |
-| `--location` | Location name or ID that the machine belongs to or to list machines in | `list`, `status`, `logs`, `part` | **Required** |
+| `--organization` | Organization name or ID that the machine belongs to | `list`, `status`, `logs`, `part` | Optional |
+| `--location` | Location name or ID that the machine belongs to or to list machines in | `list`, `status`, `logs`, `part` | Optional |
```
stuqdog, line 1324: I'm not sure what command this all is so I may be wrong, but `machine` below is still listed as `Required` but we also have `part` as a flag. I think if `part` is useful information then `machine` becomes optional (because `machine` can often be inferred from `part`, provided it's given as a `part-id`).

docs/dev/tools/cli.md
JessamyT: Good call--adding!

docs/dev/tools/cli.md
JessamyT: Yeah that was how I was writing this. Oops yes will remove location.

docs/dev/tools/cli.md
JessamyT: Hmm yeah I guess I can mark `machine` as optional since they could just use `part-id`... I'm also realizing we currently list both `machine` and `machine-id` in this chart, and show that `machine-id` is required for `api-key`.

When I run `viam machine api-key -h` currently, it doesn't show any command options other than help. Should I ticket that?

I will combine the machine and machine-id lines since they're aliases of each other. I'm tempted to simply tell users to always use the (part/machine/org/loc) ID instead of the name, since IDs are unique even across different orgs (right?) whereas names aren't, and it's simpler to say "use ID" rather than say "use ID or if you use name, then you also need to specify organization and location name, or location ID." But, names can be friendlier than IDs--thoughts?

JessamyT: (BTW this table is for the `machines` command)

JessamyT:
> This doesn't seem to address the full scope of the work done here. Part of the scope was to be consistent for flags. That means we should use one flag consistently in the docs (`--org` possibly for the orgs set) and document the aliases as options.

@npentrel In line with the alias improvement goal to be more opinionated and reduce clutter, I'm thinking it'd be best not to document the aliases here either.

Latest commit fixes the `viam machines part` args for consistency, but there are other commands e.g. `viam module upload` takes `org-id` specifically, and `viam data tag ids add` and `viam dataset data add ids` still seem to require e.g. `location-id`. I think it makes sense to sweep them later once they're all consistent.

docs/dev/tools/cli.md
```diff
@@ -1179,15 +1235,15 @@ This includes:
...
 viam machines part logs --machine=<machine> --part=<part> [...named args]
-viam machines part status --organization=<org name> --location=<location name> --machine=<machine id>
-viam machines part run --organization=<org name> --location=<location name> --machine=<machine id> [--stream] --data <meth>
-viam machines part shell --organization=<org name> --location=<location name> --machine=<machine id>
+viam machines part status --org=<org name> --location=<location name> --machine=<machine id>
+viam machines part run --org=<org name> --location=<location name> --machine=<machine id> [--stream] --data <meth>
+viam machines part shell --org=<org name> --location=<location name> --machine=<machine id> --part=<part id>
 viam machines part restart --machine=<machine id> --part=<part id>
-viam machines part cp --organization=<org name> --location=<location name> --machine=<machine id> --part=<part name> <file name> machine:/path/to/file
+viam machines part cp --org=<org name> --location=<location name> --machine=<machine id> --part=<part name> <file name> machine:/path/to/file
```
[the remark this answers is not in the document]
JessamyT: It is verbose but in keeping with the rest of this page, this is a list of all possible args, whereas the examples below show just ones you need to do a certain thing. I don't feel strongly either way but I don't want to change the whole page right now so will leave as-is.

docs/dev/tools/cli.md
JessamyT: The machine-id vs machine lines in this table are still weird since `api-key` still needs `machine-id` but `machine` seems to be the preferred flag for `status` etc. so I think this may be as good as can be for now, and hope people will use the --help flag since it's probably easier to use that than look in the docs when using the CLI anyway.

For the --machine explanation I'm changing it to more strongly recommend ID, and mark it and --part as required for their respective commands.

npentrel:
https://github.com/viamrobotics/docs/pull/3960/commits/45d36b02e67f7d30a9dcbead6d16a732c2040fb1 was what I was looking for for the organization parameter but the same needs to happen for `--location` and `--location-id` and `--machine` and `--machine-id`. We should pick one and use it I think. Are you saying all the remaining occurrances are where SDK work hasn't been done yet? Are there tickets for that work that you are following? Have you created a follow up ticket already and are there SDK tickets we can block that on?

I think there's still a part of the work not documented here, namely module generate work. Unless you have another ticket for that? I don't see one.

npentrel (approved):
LGTM as long as there are tickets to follow up.

JessamyT:
> [45d36b0](https://github.com/viamrobotics/docs/commit/45d36b02e67f7d30a9dcbead6d16a732c2040fb1) was what I was looking for for the organization parameter but the same needs to happen for `--location` and `--location-id` and `--machine` and `--machine-id`. We should pick one and use it I think. Are you saying all the remaining occurrances are where SDK work hasn't been done yet? Are there tickets for that work that you are following? Have you created a follow up ticket already and are there SDK tickets we can block that on?
> 
> I think there's still a part of the work not documented here, namely module generate work. Unless you have another ticket for that? I don't see one.

Yeah so for example `viam dataset data add ids` still uses --location-id specifically, not --location. Same with `viam data tag  ids add`, and with the `api-key` commands. At least, I have CLI version 0.62.0 and when I run `viam machines api-key create --organization=123` I get a `Flag provided but not defined: -organization` error, though https://github.com/viamrobotics/rdk/pull/4669 comment indicates otherwise so commented on the [ticket](https://viam.atlassian.net/browse/RSDK-9287) and slack.

For module generate, there was a decision to encourage users to use the prompts instead of the flags. We do document the flags though, and I think the most recent work was just making the help messages more useful and to allow fall through if a user doesn't provide some flags. Is there something in particular you're referring to that isn't documented? Going to merge to get profiles in, and will follow up as needed.

PR #3980, opened 2025-02-07
DOCS-3497: Update deploy software

npentrel:
@mcvella & @HipsterBrown thanks for the feedback so far - would you mind reviewing the changes to https://deploy-preview-3980--viam-docs.netlify.app/manage/software/deploy-software/ and see if this addresses your feedback? It's not a lot of changes but it may be enough to do the trick. If either of you have 10 min to chat through this and would prefer that, let me know.

### DIFF
--- a/docs/dev/reference/glossary/module.md
+++ b/docs/dev/reference/glossary/module.md
@@ -2,10 +2,10 @@
 title: Module
 id: module
 full_link:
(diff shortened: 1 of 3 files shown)


npentrel:
Thanks @mcvella and @HipsterBrown - I will merge for now and split the control logic into it's own page in https://viam.atlassian.net/browse/DOCS-3584

PR #3994, opened 2025-02-12
DOCS-3480: Change subtype to API in Python and in module docs

JessamyT:
Only a few changes needed for Python SDK but since we now use API more broadly (also in C++ and in the auto-generated JSON in app), might as well clarify these files.

Changes to the example JSONs we have throughout the rest of the docs will be part of https://viam.atlassian.net/browse/DOCS-3582


npentrel (approved):
Awesome

docs/dev/reference/glossary/api-namespace-triplet.md
```diff
@@ -2,18 +2,20 @@
...
 The `namespace` for built-in Viam resources is `rdk`, while the `type` is `component` or `service`.
-`subtype` refers to a specific component or service, like a `camera` or `vision`.
+`api` refers to a specific component or service, like a `camera` or `vision`.
 
-One subtype can have various {{< glossary_tooltip term_id="model" text="models" >}}, custom or built-in, but they all must conform to the subtype's API definition.
-This requirement ensures that when a resource of that model is deployed, you can [interface with it](/dev/reference/sdks/) using the same [client API methods](/dev/reference/apis/) you would when programming resources of the same subtype with a different model.
+One API can have various {{< glossary_tooltip term_id="model" text="models" >}}, custom or built-in, but they all must conform to the same API definition.
+This requirement ensures that when a resource of that model is deployed, you can [interface with it](/dev/reference/sdks/) using the same [client API methods](/dev/reference/apis/) you would when programming resources of the same API with a different model.
+
+Each resource implements one and only one API.
```
purplenicole730, line 18: Is it that one API (`ns:temporary:sensor`) can have various models, or is it that one resource subtype (i.e.: an `arm`) can have various models? If it's the former, good change. If it's the latter, it should stay as `subtype`. This line of thinking will help to decide of these changes were necessary or not.

purplenicole730 (commented):
I only went through the first few, but it's basically what we talked about offline. The word `subtype` should only change to `API` if it is referring to the actual triplet.

docs/tutorials/configure/pet-photographer.md
```diff
@@ -734,7 +734,7 @@ async def main():
     registry before the module adds the resource model.
     """
     Registry.register_resource_creator(
-        Camera.SUBTYPE,
+        Camera.API,
```
JessamyT, line 737: @purplenicole730 this is correct though?

PR #3997, opened 2025-02-13
DOCS-3591: Where do i type a name and click add machine?

JessamyT:
Addresses chatbot feedback


npentrel (approved):
Please link a ticket for this.

JessamyT:
Okay, ticket created and linked. Don't think we actually made a decision earlier whether to close this as a no-op or not though. I'll merge because I think that since it's only on this one unique page (versus general config instructions say), and is the very first page, we can be a smidge more hand-holdy. Also this way we're not being somewhat tautological by saying "start a new machine by adding a new machine and we won't tell you how" :D

PR #3998, opened 2025-02-13
DOCS-3576: Deprecate processes

npentrel:

PR #3999, opened 2025-02-13
DOCS-3585: Update changelog

npentrel:


docs/dev/reference/changelog.md
```diff
@@ -41,13 +41,37 @@ date: "2024-09-18"
 
 <!-- If there is no concrete date for a change that makes sense, use the end of the month it was released in. -->
 
+{{% changelog color="removed" title="Managed Processes" date="2025-02-01" %}}
+
+[Managed Processes](/manage/reference/processes/) are now deprecated and will be removed in a future version of `viam-server`.
```
JessamyT, line 46: Can we say what people should do instead?

JessamyT: And it looks like this changelog addition is also made here https://github.com/viamrobotics/docs/pull/3998

JessamyT (approved):
% comment

docs/dev/reference/changelog.md
npentrel, line 46:
```suggestion
[Managed Processes](/manage/reference/processes/) are now deprecated and will be removed in a future version of `viam-server`.
Instead [use modules for control logic](/manage/software/deploy-software/#create-a-module-with-machine-control-logic).
```

PR #4012, opened 2025-02-14
Update CSS for images and tabs

npentrel:
## Added shadow class to images
Screenshot 2025-02-14 at 22 49 31

## Updated tabs to fill entire content width (similar to app)
Screenshot 2025-02-14 at 22 49 23

### DIFF
--- a/docs/dev/_index.md
+++ b/docs/dev/_index.md
@@ -492,8 +492,6 @@ for i := 0; i < len(detections); i++ {
 }
 ```
 
-Be sure to import `"go.viam.com/rdk/utils"` at the beginning of your file.
-
 {{% /tab %}}
 {{< /tabs >}}
 
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -1192,11 +1192,9 @@ td > ol {
 
 .nav-tabs {
   border-bottom: none; // unset from inherited styles
-  gap: 10px;
+  gap: 1px;
   display: flex;
   flex-direction: row;
-  justify-content: space-between;
-
 }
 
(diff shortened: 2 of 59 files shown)


docs/dev/_index.md
```diff
@@ -492,8 +492,6 @@ for i := 0; i < len(detections); i++ {
 }
 ```
 
-Be sure to import `"go.viam.com/rdk/utils"` at the beginning of your file.
```
JessamyT, line 495: [feel free to ignore] I assume this is something you caught incidentally and included but just flagging since it's a big PR to make sure this is meant to be deleted here

JessamyT (commented):
I think the drop shadows are great.

For tabs, since we do have many cases of only two tabs and some cases of a single tab, as you said yesterday this does look a bit odd. I was thinking something more like what we have at the top of the docs with one horizontal bar with all the names left justified. This also appears in the app like configure control etc.
image

docs/dev/_index.md
npentrel: I think Sierra added this accidentally - but yeah it's meant to get deleted

npentrel:
It did grow on me the more I looked at it, this is an alternative pattern used in the app. I think it's possibly a bit less sleek. With design taking a look at this next week anyway, I think we can leave it as is or pick one of the two options I've put together here.
Screenshot 2025-02-14 at 23 54 29
Screenshot 2025-02-14 at 23 54 24

JessamyT (approved):
Okay seems fine to merge now and have design look after. Lots of linter changes here but looks like a small change so easy to iterate anyway.

PR #4029, opened 2025-02-20
update hardware requirements for micro-server

mattjperez:
The changes set the minimum Flash requirements for the micro-server to 8MB. 

While it's technically possible for a user to build their project with a custom partition table to utilize a 4MB device, they would not be able to utilize certain key features (ex. OTA). 

Also updated some adjacent wording.


JessamyT (approved):
LGTM thanks for the PR!

PR #4031, opened 2025-02-21
Update cli.md

npentrel:


JessamyT (commented):
Okay so, if you run `viam machines part run -h` now, you see `--organization` and not the other aliases. So maybe we should turn these all back to `--organization` as well as in the table. Or, if you want `--org-id`, change the table to match since right now the table says `--org`.

npentrel:
Ok, so that is then actually a third inconsistency but that one can be a ticket for the SDK team, the docs can already change. Unless I am wrong that the default is not `org-id`?

JessamyT:
Not sure I follow. By default, do you mean the flag help spits out, which is `organization` for these commands?

npentrel:
one of the aims of the project you documented was to default to one consistent flag where there are multiple options. I believe that flag is org-id. So this command should also use org-id. Unless there's something I'm missing, the help text seems to not align with the project. Since using org-id works, we can already change to this in the docs and then ask the sdk team to clean up the inconsistency.

JessamyT:
So if I'm not mistaken, the SDK team chose to keep `--organization` as the default flag for these commands for the following reasons (which also apply to machine vs machine-id and location vs location-id):

1. If you pass `--org-id=MyOrg` to one command with no errors, and then you start using a different CLI command that requires `--org-id=1234actualID`, you get an error and become confused.
2. We *could* eliminate this confusion by forcing everyone to always pass the org ID instead of the org name, but there is a desire to keep the flexibility of allowing org name for some commands. Yes, this creates a different type of confusion whereby we sometimes have two similar flags such as `--machine` and `--machine-id` in the same table for different arguments of one command (e.g. `viam machines`), and you can pass ID to either of them, but this was deemed by the SDK team to be the lesser of the two possible confusions. This makes sense and I don't feel strongly either way.

What I think we should do is, in this PR, change the current instances of `--org` (including in the table) to `--organization`, not `--org-id`. This will align with the SDK decision and the help message, and avoid confusion #1.

In tables such as the `machines` command table, I think we should keep it as-is: It indicates that some parameters require a `--machine-id` flag and others require a more flexible `--machine` flag. For the `organizations` command table, it looks like all those parameters require `--org-id`, so we can keep those the way they are too.

npentrel:
Let's change to `--organization` here and I'll bring the rest up for discussion.

docs/dev/tools/cli.md
```diff
@@ -26,7 +26,7 @@ The CLI lets you:
 For example, this CLI command moves a servo to the 75 degree position:
 
 ```sh {class="command-line" data-prompt="$"}
-viam machines part run --machine 82c608a-1be9-46a5 --org "Artoo's Org" \
+viam machines part run --machine 82c608a-1be9-46a5 --org-id 123 \
```
npentrel, line 29:
```suggestion
viam machines part run --machine 82c608a-1be9-46a5 --organization 123 \
```

docs/dev/tools/cli.md
```diff
@@ -636,15 +636,15 @@ This includes:
 
 ```sh {class="command-line" data-prompt="$"}
 viam machines list
-viam machines status --org=<org id> --location=<location id> --machine=<machine id>
-viam machines logs --org=<org id> --location=<location id> --machine=<machine id> [...named args]
+viam machines status --org-id=<org id> --location=<location id> --machine=<machine id>
+viam machines logs --org-id=<org id> --location=<location id> --machine=<machine id> [...named args]
```
npentrel, line 640:
```suggestion
viam machines status --organization=<org id> --location=<location id> --machine=<machine id>
viam machines logs --organization=<org id> --location=<location id> --machine=<machine id> [...named args]
```

docs/dev/tools/cli.md
```diff
@@ -636,15 +636,15 @@ This includes:
...
 viam machines api-key create --machine=<machine id> [...named args]
 viam machines part logs --machine=<machine id> --part=<part id> [...named args]
-viam machines part status --org=<org id> --location=<location id> --machine=<machine id>
-viam machines part run --org=<org id> --location=<location id> --machine=<machine id> [--stream] --data <meth>
-viam machines part shell --org=<org id> --location=<location id> --machine=<machine id> --part=<part id>
+viam machines part status --org-id=<org id> --location=<location id> --machine=<machine id>
+viam machines part run --org-id=<org id> --location=<location id> --machine=<machine id> [--stream] --data <meth>
+viam machines part shell --org-id=<org id> --location=<location id> --machine=<machine id> --part=<part id>
```
npentrel, line 645:
```suggestion
viam machines part status --organization=<org id> --location=<location id> --machine=<machine id>
viam machines part run --organization=<org id> --location=<location id> --machine=<machine id> [--stream] --data <meth>
viam machines part shell --organization=<org id> --location=<location id> --machine=<machine id> --part=<part id>
```

docs/dev/tools/cli.md
```diff
@@ -636,15 +636,15 @@ This includes:
...
 viam machines part restart --machine=<machine id> --part=<part id>
-viam machines part cp --org=<org id> --location=<location id> --machine=<machine id> --part=<part id> <file name> machine:/path/to/file
+viam machines part cp --org-id=<org id> --location=<location id> --machine=<machine id> --part=<part id> <file name> machine:/path/to/file
```
npentrel, line 647:
```suggestion
viam machines part cp --organization=<org id> --location=<location id> --machine=<machine id> --part=<part id> <file name> machine:/path/to/file
```

JessamyT (approved):
LGTM

PR #4042, opened 2025-03-04
DOCS-3623: Review subtype vs API distinction

JessamyT:
I don't think anything else from https://github.com/viamrobotics/docs/pull/3994 needs reverting because the rest is conceptually true either way

### DIFF
--- a/docs/dev/reference/glossary/api-namespace-triplet.md
+++ b/docs/dev/reference/glossary/api-namespace-triplet.md
@@ -2,15 +2,15 @@
 title: API Namespace Triplet
 id: api-namespace-triplet
 full_link:
-short_description: namespace:type:api, for example `rdk:component:sensor`
+short_description: namespace:type:subtype, for example `rdk:component:sensor`
 ---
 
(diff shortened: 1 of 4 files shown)


purplenicole730 (approved):
LGTM, assuming all the other changes were fine

purplenicole730 (changes requested):
image
These are wrong as well, and this is just from me glancing down quickly.

JessamyT:
Oop yeah fixing those now. FWIW that cpp page is not currently viewable in prod anyway. I will add back the top two in your screenshot (as well as the subtype glossary definition) but I think conceptually the third change in the screenshot is true either way? Strings in code aside, it seems like subtype == API conceptually. The old glossary definition said "A subtype is designated by its API namespace triplet." which I will change to "The `subtype` string is the third part of the {{< glossary_tooltip term_id="api-namespace-triplet" text="API namespace triplet" >}}." since I think that's more accurate.

purplenicole730 (approved):
My head hurts. Please save me

PR #4043, opened 2025-03-04
DOCS-3192: Add discovery service page

JessamyT:
https://deploy-preview-4043--viam-docs.netlify.app/operate/reference/services/discovery/


docs/dev/reference/changelog.md
```diff
@@ -41,6 +41,26 @@ date: "2024-09-18"
...
+{{% changelog color="added" title="Button API and switch API" date="2025-01-27" %}}
+
+Two new component APIs, the [button API](/dev/reference/apis/components/button/) and [switch API](/dev/reference/apis/components/switch/), have been added.
+
+{{% /changelog %}}
+
+{{% changelog color="added" title="Discovery service API" date="2025-01-27" %}}
+
+The [discovery service API](/dev/reference/apis/services/discovery/) has been added.
+You can now write modules that discover resources on a robot.
```
JessamyT, line 53:
```suggestion
You can now write modules that discover resources on a machine.
```

docs/operate/reference/components/camera/webcam.md
```diff
@@ -31,7 +31,8 @@ Enter a name or use the suggested name for your camera and click **Create**.
 {{< imgproc src="/components/camera/configure-webcam.png" alt="Configuration of a webcam camera in the Viam app config builder." resize="1200x" style="width=600x" class="shadow"  >}}
 
 Edit and fill in the attributes as applicable.
-If you click on **Show more**, then the **video_path** field while your machine is live, a dropdown autopopulates with identified camera paths.
+Leave the **video_path** blank and the camera will use the default video path for your machine.
+If this doesn't work when you test your camera later, you can try a different video path by following the prompt in the Viam app to add a discovery service and identify path options.
```
JohnN193, line 35: would we want to link to the registry page for the discovery service? `following the prompt` is prob fine just wondering

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,66 @@
...
+## Example usage
+
+Imagine you are creating a vision service module that depends on a camera.
+You might want to include both a vision service and a discovery service in your module.
+You could write a discovery service that reports all camera paths recognized by your computer or SBC.
+Users of your module could then:
+
+1. Configure the discovery service in their machine's configuration.
+1. Use it to discover any cameras that are connected to the machine.
+1. Use one of the discovered cameras in their vision service configuration.
```
JohnN193, line 30: [q] so visually it looks like this works, but I notice in some examples we step through numbers ourselves(ex 1. 2. 3.) is it preferred to leave it like this?

JohnN193 (approved):
lgtm, just some smallish questions

docs/operate/reference/services/discovery/_index.md
JessamyT: These render correctly and I like doing it this way since when you add or delete a step you don’t have to change them all. Thanks for checking!

docs/operate/reference/components/camera/webcam.md
JessamyT:
```suggestion
If this doesn't work when you test your camera later, you can try a different video path by following the prompt in the Viam app to add a [discovery service] (https://app.viam.com/module/rand/find-webcams) and identify path options.
```
Ah yeah fair point. There’s a ticket Rand just made for app to actually change the prompt to auto add the service but doesn’t hurt to add this link for now.

docs/operate/reference/components/camera/webcam.md
```diff
@@ -31,7 +31,8 @@ Enter a name or use the suggested name for your camera and click **Create**.
 {{< imgproc src="/components/camera/configure-webcam.png" alt="Configuration of a webcam camera in the Viam app config builder." resize="1200x" style="width=600x" class="shadow"  >}}
 
 Edit and fill in the attributes as applicable.
-If you click on **Show more**, then the **video_path** field while your machine is live, a dropdown autopopulates with identified camera paths.
+Leave the **video_path** blank and the camera will use the default video path for your machine.
+If this doesn't work when you test your camera later, you can try a different video path by following the prompt in the Viam app to add a [discovery service](https://app.viam.com/module/rand/find-webcams) and identify path options.
```
npentrel, line 35: might be good to be more specific to avoid confusion tha tthis is the only discovery service
```suggestion
If this doesn't work when you test your camera later, you can try a different video path by following the prompt in the Viam app to add a [webcam discovery service](https://app.viam.com/module/rand/find-webcams) and identify path options.
```

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,66 @@
+---
+title: "Discovery Service"
+linkTitle: "Discovery"
+description: "The discovery service is used to discover resources on a machine."
```
npentrel, line 4: passive voice

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,66 @@
...
+#images: ["/services/icons/discovery.svg"]
+tags: ["discovery", "services"]
+no_service: true
+date: "2025-03-03"
+# updated: ""  # When the content was last entirely checked
+# SMEs: John N.
+modulescript: true
+---
+
+The discovery service is used to discover resources on a machine.
```
npentrel, line 18: passive voice

npentrel: Does the discovery service discover resources as in components OR does it discover attached hardware (and the paths under which the hardware is available) or both? Clarifying this further is I think important.

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,66 @@
...
+If you are [creating a modular resource](/operate/get-started/other-hardware/) that depends on other resources that are discoverable in a systematic way, you can create a discovery service as part of your module to discover those resources.
+
+## Example usage
+
+Imagine you are creating a vision service module that depends on a camera.
+You might want to include both a vision service and a discovery service in your module.
+You could write a discovery service that reports all camera paths recognized by your computer or SBC.
+Users of your module could then:
```
npentrel, line 26: Avoid conditional in favour of explaining the value this would add up front. 
Vale should be underlining conditionals and passive voice for you. Please check it's working.

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,66 @@
...
+date: "2025-03-03"
+# updated: ""  # When the content was last entirely checked
+# SMEs: John N.
+modulescript: true
+---
+
+The discovery service is used to discover resources on a machine.
+If you are [creating a modular resource](/operate/get-started/other-hardware/) that depends on other resources that are discoverable in a systematic way, you can create a discovery service as part of your module to discover those resources.
+
+## Example usage
```
npentrel, line 21: This is all a bit abstract to me right now. What does it discover/return and how do I use it? Do I need to view that in the logs or does the UI just add a drop down? Since this is already covering example usage, can you walk through a specific vision service that does this? Then you can explain what the discovery service returns and how you can use it.

npentrel (changes requested):

docs/operate/reference/services/discovery/_index.md
JessamyT: Okay I added some more detail here. On Monday I'll check with John whether there's a better sort of example--right now there's rtsp and webcam, but not sure we want to go that specific?

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,67 @@
...
+tags: ["discovery", "services"]
+no_service: true
+date: "2025-03-03"
+# updated: ""  # When the content was last entirely checked
+# SMEs: John N.
+modulescript: true
+---
+
+A discovery service allows you to return a list of physical hardware available on a machine, and suggest configurations for those components to integrate the hardware into the machine.
+If you are [creating a modular resource](/operate/get-started/other-hardware/) that depends on other resources that are discoverable in a systematic way, you can create a discovery service as part of your module to discover those resources.
```
npentrel, line 19: I think also here we want to be more specific than resources. If it's hardware then I'd say hardware resources and if it's viam resources then I think add the glossary entry so there's no room for misinterpretation. Still not sure whether it's meant just for hardware or for hardware or configured components/services

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Example usage
+
+Imagine you are creating a vision service module that depends on a camera, and you want to make it easier for users to configure the camera.
+You can include both a vision service and a discovery service in your module.
+You'd write a discovery service that reports all camera paths recognized by your computer or SBC.
+Users of your module could then:
```
npentrel, line 26: Remove conditional & shorten
```suggestion
Imagine you are creating a vision service module that depends on a camera.
To make it easier for users to configure the camera, you include a discovery service in your module.
You implement the discovery service to report all the camera paths your computer or SBC finds.
Users of your module can then:
```

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Configure both the vision service and the discovery service in their machine's configuration.
+1. Click the **Test** panel in the discovery service configuration to see all cameras recognized by the machine, presented as a list of configuration snippets.
+1. Copy the copy-pasteable configuration snippet for the camera they want to use, with the camera path already filled in.
```
npentrel, line 30:
```suggestion
1. Create a camera with the copy-pasteable configuration snippet for the camera they want to use, with the camera path already filled in.
```

docs/operate/reference/services/discovery/_index.md
```diff
@@ -0,0 +1,67 @@
...
+1. Use one of the discovered cameras in their vision service configuration.
```
npentrel, line 31:
```suggestion
1. Use the configured camera in their vision service configuration.
```

npentrel (approved):
I think this is better. Explaining it returns configuration snippets is I think what was missing for me. If there is a better example then possibly still worth changing but otherwise I think it's ok now

docs/operate/reference/services/discovery/_index.md
JessamyT: I mean, it is just an API. So someone could use it differently. I believe the idea for it was to surface hardware, but you could implement it differently--just have to return a list of resource configs.

JessamyT, line 19:
```suggestion
If you are [creating a modular resource](/operate/get-started/other-hardware/) that depends on other {{< glossary_tooltip term_id="resource" text="resources" >}} that are discoverable in a systematic way, you can create a discovery service as part of your module to discover those resources.
```

PR #4047, opened 2025-03-05
DOCS-3592 and DOCS-2331: Add note for renaming a module and instructions to transfer ownership

JessamyT:


docs/operate/get-started/other-hardware/manage-modules.md
```diff
@@ -416,6 +416,42 @@ To delete just one version of a module:
...
+
+1. Run `viam module update` to push the changes to the registry.
+
+1. Publish a new version of the module to the registry by following either set of update steps on this page.
+
+## Rename a module
+
+If you need to change the name of a module, please reach out to the Viam team at [support@viam.com](mailto:support@viam.com).
+
 ## Next steps
```
npentrel, line 455: Out of diff but please remove the next steps. Since we have the next button at the bottom, this feels out of place. It also seems odd after info about transfering a module. This link could be with updating an existing module.

JessamyT: Okay! We also have this link up top so yeah removing

PR #4054, opened 2025-03-06
Remove giant images

npentrel:


JessamyT (approved):
Looks like these images only appear here so can delete them.

[opt] Consider replacing with smaller images (or resizing), since we do get this question in the chatbot sometimes.

PR #4071, opened 2025-03-12
DOCS-3637: Add VIAM_MODULE_STARTUP_TIMEOUT env variable

JessamyT:
@benjirewis to double check--this is there for local modules too right?


npentrel (approved):

benjirewis:
> to double check--this is there for local modules too right?

Yep, for both registry and local modules.

PR #4074, opened 2025-03-12
DOCS-3690: Make clear that one model implements one API

JessamyT:


npentrel (changes requested):
The changes here are fine but I think the issue the person was having may be a different one: I assume what they wanted to do here is something like add the sensor component to the camera component in the `depends_on` field. Then they could use the API methods of the sensor to also interact with the camera. I think we don't really cover how the `depends_on` config is useful for modules and that it allows one modular resource to also access other resources of the machine. I have seen multiple chats that seemed to be around this - so I think we should cover it.

JessamyT:
I put a note on the spreadsheet but my thought is that this Q was super vague, but did bring up this API issue which I wanted to address. Meanwhile there are other qs on the sheet that are more specifically to do with dependencies so I'm going to address those separately.

PR #4081, opened 2025-03-12
DOCS-3578: Add custom API imports note

JessamyT:
I know a note in next steps is atypical but it relates to next steps, and shouldn't be on the regular implement a model page

### DIFF
--- a/docs/operate/reference/advanced-modules/create-subtype.md
+++ b/docs/operate/reference/advanced-modules/create-subtype.md
@@ -106,3 +106,15 @@ Now that your resource API is defined, create a new model that implements your n
 
 {{% /manualcard %}}
 {{< /cards >}}
+
+{{% alert title="Note" color="note" %}}
+
+After you define a new API and create a model that implements it, keep the following in mind when writing code against your new API:
+
+- You can't use [SDKs](/dev/reference/sdks/) to call your new API unless you build out the SDKs to support it.
+  It will be easiest to write code against your new API in the language you used to define it.
+- Since your API doesn't have built-in SDK support, you'll need a local copy of the module code on whatever machine is running client code against it.
+- Be sure to import the API definition from the module directory, for example `from /path/to/module/src.gizmo import Gizmo`.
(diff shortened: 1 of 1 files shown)


docs/operate/reference/advanced-modules/create-subtype.md
```diff
@@ -106,3 +106,15 @@ Now that your resource API is defined, create a new model that implements your n
 
 {{% /manualcard %}}
 {{< /cards >}}
+
+{{% alert title="Note" color="note" %}}
+
+After you define a new API and create a model that implements it, keep the following in mind when writing code against your new API:
+
+- You can't use [SDKs](/dev/reference/sdks/) to call your new API unless you build out the SDKs to support it.
```
njooma, line 114:
```suggestion
- You can't use [SDKs](/dev/reference/sdks/) to call your new API unless you build out the client to support it.
```

docs/operate/reference/advanced-modules/create-subtype.md
```diff
@@ -106,3 +106,15 @@ Now that your resource API is defined, create a new model that implements your n
...
+  It is easiest to write code against your new API in the language you used to define it.
+- Since your API doesn't have built-in SDK support, you'll need a local copy of the module code on whatever machine is running client code against it.
+- Be sure to import the API definition from the module directory, for example `from /path/to/module/src.gizmo import Gizmo`.
```
njooma, line 117: the import reads a bit like pseudo-code. Should be more like `from path.to.module import Gizmo` if we're doing python

njooma (changes requested):

docs/operate/reference/advanced-modules/create-subtype.md
JessamyT:
```suggestion
- Be sure to import the API definition from the module directory, for example in Python it would be of the form `from path.to.module.src.gizmo import Gizmo`.
```

njooma (approved):

npentrel (approved):

PR #4085, opened 2025-03-13
DOCS-3251: Add motion DoCommand Go example

JessamyT:
Considered putting into an expander, but it's towards the bottom of the page anyway and people are likely to miss it if they see the short example and this one is hidden


biotinker (changes requested):
We do not want to be documenting this; the undocumented nature of DoCommand is a feature, not an oversight.

PR #4087, opened 2025-03-14
DOCS-3687: Elevate supported hardware and component APIs

JessamyT:
- Move component API links to separate left nav item
- Add filter tabs to search models by API
- Add screenshot to configure procedural
- Remove separate virtual hardware components heading and move it into a sentence at the end of the page intro

https://deploy-preview-4087--viam-docs.netlify.app/operate/get-started/supported-hardware/


npentrel (commented):
I like this a lot. I am wondering whether it would be a good idea to swap "About the Viam Registry of supported hardware and software" and "Configure hardware on your machine" in the order of the page. That would put the action first and the "additional info" later.

npentrel (approved):

IanWhalen:
Thanks for poking me. Looks like the big thing here is that it's: intro, procedure, details, component list?  If so that sounds great to me.

JessamyT:
> Thanks for poking me. Looks like the big thing here is that it's: intro, procedure, details, component list? If so that sounds great to me.

@IanWhalen So this PR doesn't really change the order of anything--the order you just listed is already live in the docs. Rather, this PR does the following (added to PR description just now too):
- Move component API links to separate left nav item so that they don't explode the left nav when you click configure hardware, and so that they're more discoverable
- Add filter tabs to search models by API
- Add screenshot to configure procedural
- Remove separate virtual hardware components heading and move it into a sentence at the end of the page intro

IanWhalen:
Ah thanks I think I missed number 1 because I was looking on my phone.
Looks good to me!

Ian M. Whalen / 347-563-8342
http://www.ianwhalen.com
http://www.linkedin.com/in/ianwhalen

PR #4088, opened 2025-03-14
DOCS-3554: Add TypeScript

npentrel:
Please only review for mistakes (wrong information or accidentally deleting anything that shouldn't be deleted or things like that). I'm not fixing missing parameter descriptions or similar here. We'll need to do that upstream when we're adding code samples.

Note: This doesn't output code samples right now, I will need to add that logic once we have code samples in the sdk docs.

### DIFF
--- a/.github/workflows/parse_typescript.py
+++ b/.github/workflows/parse_typescript.py
@@ -1,0 +1,140 @@
+import pprint
+
+from parser_utils import make_soup
+from markdownify import markdownify as md
+
+## Language-specific resource name overrides:
+typescript_resource_overrides = {
+    "generic_component": "GenericComponent",
+    "generic_service": "GenericService",
+    "movement_sensor": "MovementSensor",
+    "power_sensor": "PowerSensor",
+    "vision": "Vision",
+    "robot": "Robot",
+    "data": "Data",
+    "dataset": "Data",
+    "data_sync": "Data",
+    "data_manager": "DataManager",
+    "mltraining": "MlTraining"
+}
+
+## Ignore these specific APIs if they error, are deprecated, etc:
+typescript_ignore_apis = [
+    'getRoverRentalRobots' # internal use
+]
+
+## Use these URLs for data types that are built-in to the language:
+typescript_datatype_links = {}
+
+class TypeScriptParser:
+    def __init__(self, proto_map_file, typescript_staging_url = None):
+        self.proto_map_file = proto_map_file
+        self.sdk_url = "https://ts.viam.dev"
+
+        if typescript_staging_url:
+            self.scrape_url = typescript_staging_url
+            self.staging = True
+        else:
+            self.scrape_url = "https://ts.viam.dev"
+            self.staging = False
+        self.typescript_methods = {}
+
+
+    def parse(self, type, viam_resources, args):
+        self.typescript_methods[type] = {}
+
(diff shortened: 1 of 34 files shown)


.github/workflows/parse_typescript.py
```diff
@@ -0,0 +1,142 @@
...
+        # Skip resources not supported in TypeScript
+        unsupported_resources = [
+            "base_remote_control", "input_controller", "pose_tracker",
+            "mlmodel"
+        ]
+
+        for resource in viam_resources:
+
+            ## Determine URL form for TypeScript depending on type (like 'component').
+            ## TEMP: Manually exclude Base Remote Control Service (Go only):
```
JessamyT, line 54: [nit] doesn’t need to be here since go has its own file now right?

.github/workflows/parse_typescript.py
```diff
@@ -0,0 +1,142 @@
...
+
+            url = f"{self.scrape_url}/classes/{resource.capitalize()}Client.html"
+            if resource in typescript_resource_overrides:
+                url = f"{self.scrape_url}/classes/{typescript_resource_overrides[resource]}Client.html"
+
+            if args.verbose:
+                print(f'DEBUG: Parsing TypeScript URL: {url}')
+
+            ## Scrape each parent method tag and all contained child tags for TypeScript by resource.
+            ## TEMP: Manually exclude Base Remote Control Service (Go only).
```
JessamyT, line 73: And here

.github/workflows/sdk_protos_map.csv
```diff
@@ -1,480 +1,478 @@
...
-gantry,MoveToPosition,,move_to_position,MoveToPosition,moveToPosition
-gantry,GetLengths,,get_lengths,Lengths,lengths
-gantry,Home,,home,Home,home
+gantry,GetPosition,,get_position,Position,position,getPosition
+gantry,MoveToPosition,,move_to_position,MoveToPosition,moveToPosition,moveToPosition
+gantry,GetLengths,,get_lengths,Lengths,lengths,getLengths
+gantry,Home,,home,Home,home,home
 ## TED: Gantry in Go SDK doesn't appear to implement (inherit) this:
-gantry,GetGeometries,,,,
+gantry,GetGeometries,,,,,getGeometries
```
JessamyT, line 95: looks like Python get_geometries was removed from the duplicate line below but should be here

static/include/components/apis/generated/generic_component.md
```diff
@@ -111,61 +125,6 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 {{% /tab %}}
 {{< /tabs >}}
 
-### DoCommand
```
JessamyT, line 114: Is this meant to be deleted?

JessamyT (commented):
Still need to review some of the generated files later today

static/include/components/apis/generated/generic_component.md
npentrel: If you expand this file above, you can see it's here twice actually.

.github/workflows/sdk_protos_map.csv
npentrel: yep, fixing in https://github.com/viamrobotics/docs/pull/4090

PR #4090, opened 2025-03-17
DOCS 3554: Check for missing methods

npentrel:
Adding GitHub Action to alert us to missing method docs based on what's in the SDK docs


.github/workflows/check-methods.yml
JessamyT, line 1: [nit] Line 46 is still Python-specific; ticket name should be more general I think

.github/workflows/update_sdk_methods.py
```diff
@@ -50,6 +50,7 @@
                      the complete data object from parse() to /tmp/update_sdk_methods_debug.txt. \
                      Also prints high-level status updates to STDOUT. \
                      Deletes previous debug file when run again.")
+parser.add_argument('-c', '--coverage', action='store_true', help="Fails the build if any methods are unused.")
```
JessamyT, line 53: What does "unused" mean? Seems like this should say "not documented"?

JessamyT (approved):
Generated files look good so I assume code works

.github/workflows/update_sdk_methods.py
npentrel: methods in the sdk docs that didn't get used / put into generated files

JessamyT: I just think it's not the clearest term for this

PR #4102, opened 2025-03-19
DOCS-3683: Add module timeout to troubleshooting

JessamyT:


npentrel (approved):

docs/dev/tools/common-errors.md
```diff
@@ -172,6 +172,37 @@ In **JSON** mode, it will resemble the following:
...
+      ...
+      "env": {
+        "VIAM_MODULE_STARTUP_TIMEOUT": "6m30s",
+        "VIAM_RESOURCE_CONFIGURATION_TIMEOUT": "3m0s"
+      }
+    }
+  ]
+}
+```
+
```
JessamyT, line 203:
```suggestion

- If you are the module author, consider packaging the module with required dependencies so they don't need to be downloaded on startup.
   You can do this using the [Viam cloud build action](/operate/get-started/other-hardware/manage-modules/#update-automatically).

```

JessamyT: @mcvella in the related discord thread https://discord.com/channels/1083489952408539288/1154806724381917256 there's also the docker example but I'm not too familiar with different packaging methods and my suggestion here feels vague. Is there any specific wording here you'd suggest?

mcvella: I think you could maybe point to using pyinstaller as an example as per:
https://docs.viam.com/operate/get-started/other-hardware/#upload-your-module
Cloud build is just a way to automate building using something like pyinstaller but of course this also depends on the language you are writing your module in.  But often youd statically link your dependancies with Go, whereas you'd not with Python if you are using the venv approach, which can lead to long startup times the first time.

JessamyT:
```suggestion

- If you are the module author, consider packaging the module with required dependencies so they don't need to be downloaded on startup.
   For Python modules, you can package your module with dependencies by using the PyInstaller steps when [uploading your module](/operate/get-started/other-hardware/#upload-your-module).

```

docs/dev/tools/common-errors.md
```diff
@@ -172,6 +172,40 @@ In **JSON** mode, it will resemble the following:
...
+- If you are the module author, consider packaging the module with required dependencies so they don't need to be downloaded on startup.
+   For Python modules, you can package your module with dependencies by using the PyInstaller steps when [uploading your module](/operate/get-started/other-hardware/#upload-your-module).
```
JessamyT, line 205:
```suggestion
  For Python modules, you can package your module with dependencies by using the PyInstaller steps when [uploading your module](/operate/get-started/other-hardware/#upload-your-module).
```

PR #4103, opened 2025-03-19
DOCS-3714: Update tab hover effects

npentrel:


JessamyT (approved):
I think this is an improvement but would still be good to allow users to click in a wider area so navigating the docs doesn't feel fussy/require such careful mouse precision

npentrel:
> I think this is an improvement but would still be good to allow users to click in a wider area so navigating the docs doesn't feel fussy/require such careful mouse precision

I don't have strong feelings about this. The designer wants it like this, so I will do that for now. If there are a lot of dead clicks I will revisit but otherwise I won't bring this up again with design

PR #4116, opened 2025-03-21
DOCS-3746: Update robot dog tutorial to use new generator

npentrel:


docs/tutorials/custom/custom-base-dog.md
```diff
@@ -221,37 +221,51 @@ You can also try turning the robot off and on again, and then retrying the proce
...
+     - In the [Viam app](https://app.viam.com), navigate to your organization settings through the menu in upper right corner of the page.
+       Find the **Public namespace** (or create one if you haven't already) and copy that string.
+       In the example snippets below, the namespace is `viamlabs`.
+   - Resource to add to the module (API): `Base Component`.
+   - Model name: `robotdog`
+   - Enable cloud build: `No`
+   - Register module: `Yes`
 
-You can use a different model name, module namespace, and module name, but you need to use the existing API triplet `rdk:component:base` in order for your custom base to work properly as a base with `viam-server` and the [Viam app](https://app.viam.com/).
+   It's a good idea to use the same names, aside from the namespace. That way your code will match the example code.
```
JessamyT, line 261:
```suggestion
   It's a good idea to use the same names, aside from the namespace.
   That way your code will match the example code.
```
Nit

docs/tutorials/custom/custom-base-dog.md
```diff
@@ -221,37 +221,51 @@ You can also try turning the robot off and on again, and then retrying the proce
...
+   - Language: `Python`
+   - Visibility: `Private`
+   - Namespace/Organization ID:
+     - In the [Viam app](https://app.viam.com), navigate to your organization settings through the menu in upper right corner of the page.
+       Find the **Public namespace** (or create one if you haven't already) and copy that string.
+       In the example snippets below, the namespace is `viamlabs`.
+   - Resource to add to the module (API): `Base Component`.
+   - Model name: `robotdog`
+   - Enable cloud build: `No`
+   - Register module: `Yes`
```
JessamyT, line 259: Not sure about this since they use it as a local module. Guess it doesn’t matter.

JessamyT (approved):
So, if the point of this is to just get the AI to not point to the old generator LGTM but, I’m assuming the generated code doesn’t perfectly match. Did you test it and check whether the code is the same? Also this still uses processes.

npentrel:
> So, if the point of this is to just get the AI to not point to the old generator LGTM but, I’m assuming the generated code doesn’t perfectly match. Did you test it and check whether the code is the same? Also this still uses processes.

Yeah I think removing Yeoman is priority. I can't really test this easily since I don't have the base. We probably should update processes and then also retest but I think that can be separate. I'll make a ticket.

PR #4118, opened 2025-03-21
DOCS-3749: Store data from viam in MongoDB

npentrel:


npentrel:
@jdamon96 do you mind reviewing these changes? I'm wondering: is there a way to only capture to the configured MongoDB and disable data sync outside of that? 

There was a question about this from a user.

jdamon96:
Changes LGTM. Good clarification that the feature works regardless of whether the Mongo collection is on the local network or not.

Re:  is there a way to only capture to the configured MongoDB and disable data sync outside of that?
- because this was built as a feature of data capture and not data sync, I believe the answer to this question is yes: the user can only capture to the configured MongoDB collection by enabling this feature while disabling data sync.
- it may be worth noting that at scoping stage, the fact this feature is coupled with data capture rather than data sync was viewed as a shortcoming and this implementation (adding this functionality to the data capture path vs data sync path) was chosen for its speedier implementation.

jdamon96:
^ ran the above answer by Nick Sanford, he confirmed it's correct and added the clarification that in the case where the user has a direct to mongo URI provided and data sync turned off, the data will still be captured to disk (in addition to being captured to mongo)

npentrel:
> ^ ran the above answer by Nick Sanford, he confirmed it's correct and added the clarification that in the case where the user has a direct to mongo URI provided and data sync turned off, the data will still be captured to disk (in addition to being captured to mongo)

quite cool, I'll make the changes later

npentrel:
Alternatively @nicksanford would you mind reviewing this for accuracy? Thanks!

nicksanford (approved):
correct

npentrel:
Thanks!

PR #4119, opened 2025-03-21
DOCS-3744: Add version information for Jetpack

npentrel:


JessamyT (approved):
Does this mean people cannot use both GPIO and these vision services at the same time? I guess it is what it is

npentrel:
> Does this mean people cannot use both GPIO and these vision services at the same time? I guess it is what it is

We don't know for sure but testing each configuration goes well beyond what we can do. The Jetson boards are quite time-intensive to get running so I'll rely on Matt's info

PR #4130, opened 2025-03-24
DOCS-3417: Add sample code to Create a Web App

JessamyT:


docs/operate/control/web-app.md
```diff
@@ -36,6 +36,222 @@ Refer to the [Viam TypeScript SDK](https://ts.viam.dev/) documentation for avail
 
 ### Example usage
 
+{{< expand "Example camera and sensor code" >}}
+
+The following files are an example of a simple TypeScript web app that connects to a machine and displays the latestimage from the machine's camera, and the latest sensor readings.
```
npentrel, line 41:
```suggestion
The following files are an example of a TypeScript web app that connects to a machine and displays the latest image from the machine's camera, and the latest sensor readings.
```

docs/operate/control/web-app.md
```diff
@@ -36,6 +36,222 @@ Refer to the [Viam TypeScript SDK](https://ts.viam.dev/) documentation for avail
 
 ### Example usage
 
+{{< expand "Example camera and sensor code" >}}
```
npentrel, line 39: I don't think this should be in an expander. It's too easy to miss

docs/operate/control/web-app.md
```diff
@@ -36,6 +36,222 @@ Refer to the [Viam TypeScript SDK](https://ts.viam.dev/) documentation for avail
...
+
+The following files are an example of a simple TypeScript web app that connects to a machine and displays the latestimage from the machine's camera, and the latest sensor readings.
+
+{{<imgproc src="/operate/ts-dashboard.png" resize="x1100" declaredimensions=true alt="A web browser displaying a dashboard with a camera feed and sensor readings." style="max-width:500px" class="imgzoom" >}}
+
+<file>main.ts</file>:
```
npentrel, line 45: A common pattern with web dev is to put these three files in tabs. We could do that here to preserve space

docs/operate/control/web-app.md
npentrel: I like how the Flutter app page works for that better

docs/operate/control/web-app.md
```diff
@@ -36,6 +36,222 @@ Refer to the [Viam TypeScript SDK](https://ts.viam.dev/) documentation for avail
...
+}
+```
+
+{{< /expand >}}
+
+<br>
+
 For an example using Vite to connect to a machine, see [Viam's vanilla TypeScript quickstart example on GitHub](https://github.com/viamrobotics/viam-typescript-sdk/tree/main/examples/vanilla).
 
 The following tutorial uses the Viam TypeScript SDK to query data that has been uploaded to the Viam cloud from a sensor, and display it in a web dashboard.
```
npentrel, line 257: I think this looks a bit odd here.  It takes up more space than the expander

npentrel (changes requested):

docs/operate/control/web-app.md
npentrel: yep

npentrel (approved):
LGTM. I do wonder if instead of displaying a single image it should show a live stream. 

We do now have 2 of these pages feature examples that have the drive a rover in a square in it. And this one doesn't. We naturally don't have to stick with the driving a rover examples. I think it might in fact make sense to move away from it and using an example that aligns better with the use cases we've seen more of recently? What do you think?

JessamyT:
> LGTM. I do wonder if instead of displaying a single image it should show a live stream.
> 
> We do now have 2 of these pages feature examples that have the drive a rover in a square in it. And this one doesn't. We naturally don't have to stick with the driving a rover examples. I think it might in fact make sense to move away from it and using an example that aligns better with the use cases we've seen more of recently? What do you think?

Yeah so this was an intentional move away from that and towards the use cases we're seeing. We've gotten AI Qs both about getting an image and getting a video stream, as well as getting sensor data. The air quality tute already shows querying data so I figured just directly getting data would be good for the simple example here.

For the headless app, I think people are slightly less likely to be looking for a camera stream, and we get fewer Qs about displaying a camera stream or querying sensor data with Go or C++ (or maybe even Python), so not important to change that example. For the mobile app, I will make [a ticket](https://viam.atlassian.net/browse/DOCS-3778) to change it to something more like this but not going to prioritize that.

PR #4151, opened 2025-03-26
DOCS-3737: Move TOC

npentrel:


layouts/docs/search.html
JessamyT, line 1: Is this related to this PR?

npentrel: not strictly

PR #4152, opened 2025-03-26
DOCS-3417: Make web app sample show a video stream instead of image

JessamyT:


npentrel (approved):
Nice! Thanks for updating!

micheal-parks (commented):
nice! 🚀

in the future it might be cool to use the [new svelte sdk](https://github.com/viamrobotics/viam-svelte-sdk?tab=readme-ov-file#createstreamclient) or react sdk (whenever it gets built) since it makes this much easier to work with

micheal-parks (approved):

JessamyT:
Ah okay I'll make [a ticket](https://viam.atlassian.net/browse/DOCS-3800) to do that! Would you recommend one over the other for any reason @micheal-parks ? (assuming I wait till react gets built)

PR #4156, opened 2025-03-27
DOCS-3699, DOCS-2540: Add module lifecycle and dependencies guidance

JessamyT:
Discovery service will be further fleshed out separately

Staging:
https://deploy-preview-4156--viam-docs.netlify.app/operate/get-started/other-hardware/#how-and-where-do-modules-run and https://deploy-preview-4156--viam-docs.netlify.app/operate/get-started/other-hardware/dependencies/


docs/operate/get-started/other-hardware/dependencies.md
```diff
@@ -191,7 +191,7 @@ If you need to maintain the state of your resource, see [(Optional) Create and e
 {{< tabs >}}
 {{% tab name="Python" %}}
 
-If you prefer to use explicit dependencies (for example, for an optional dependency), the steps are the same as for implicit dependencies above, except that you do not need to return the dependency from the `validate_config` method and can instead return an empty list:
+If you prefer to use explicit dependencies (for example, for an optional dependency), the steps are the same as for implicit dependencies, except that you do not need to return the dependency from the `validate_config` method and can instead return an empty list:
```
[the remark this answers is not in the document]
michaellee1019: We have two mechanisms for declaring dependencies which are essentially the same result. Implicit is more common than explicit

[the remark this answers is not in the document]
JessamyT: What I mean by "optional" here is that if the user doesn't configure a "camera_name" at all, either in depends_on or in attributes, that the modular resource should still work. But I'm realizing the example I have here doesn't indicate that--I should make this validate_config explicit dependency example show "camera_name" as an optional attribute, and keep it as a required attribute for the implicit dependency example.

Is it true that people should use implicit dependencies for required things and explicit dependencies for optional things?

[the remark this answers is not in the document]
JessamyT: I guess, is it currently possible for people to have an optional implicit dependency, or not until the upcoming work happens?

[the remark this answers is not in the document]
michaellee1019: No way to have an optional explicit dependency. Yeah you could make an implicit dependency optional. Its a bit annoying right now:
1. See in validate that its not defined and don't return it as a dependency.
2. In your implementation you check to make sure the dependency is not null and gracefully skip using it.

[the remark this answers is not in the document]
JessamyT: To illuminate other reviewers: Based on offline sync, we were using "optional" in different ways--If a user does put something in depends_on, the resource won't currently start if the dependency doesn't start. What I was meaning with "optional" is it's optional for the user to even configure it at all. But sounds like you can do that with either implicit or explicit deps--it's all about how you implement validate.

Based on offline convo I will remove explicit dependencies since they just confuse users and aren't actually necessary except in cases of backwards compatibility.

JessamyT:
@michaellee1019 I changed it to optional vs required but LMK if you think that is still confusing or overkill and it'd be better to just have one heading with "use dependencies" in general, with the required deps examples, and omit the optional examples.

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,68 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+The lifecycle of a module and the resources it provides is as follows:
+
+1. `viam-server` starts, and if it is connected to the internet, it checks for configuration updates.
+1. `viam-server` starts any configured modules.
+1. When a module initializes, it registers its model or models and associated [APIs](/dev/reference/apis/) with `viam-server`, making the models available for use.
+1. For each modular resource configured on the machine, `viam-server` uses the resource's `validate` function and the `depends_on` field in the resource configuration to determine the dependencies of the resource.
+1. If a dependency is not already running, `viam-server` starts it before starting the resource.
+   If a dependency is not found or fails to start, `viam-server` will not start the resource that depends on it.
+1. `viam-server` builds the resource based on its resource configuration.
+1. If configuration fails due to a validation failure or an exception thrown by the reconfigure function, `viam-server` attempts to reconfigure the resource.
```
JessamyT, line 67:
```suggestion
1. `viam-server` calls the resource's constructor to build the resource based on its configuration.
   Typically, the constructor calls the reconfigure function.
1. If configuration fails due to a validation failure or an exception thrown by the reconfigure or constructor functions, `viam-server` attempts to reconfigure the resource.
```
I realized we got the following question which I want to make sure this lifecycle explanation answers: 
"if constructor returns an error, will the resource manager try again repeatedly to register the camera by re-calling constructor?"
What's a good way to answer this @michaellee1019 ?

cheukt: the resource manager will try repeatedly if the constructor errors, yes

docs/operate/get-started/other-hardware/dependencies.md
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+
+Be sure to handle the case where the dependency is not configured in your API implementation as well.
+
+{{% /tab %}}
+{{< /tabs >}}
+
+{{% hiddencontent %}}
+There is not currently an SDK method to directly access configuration attributes of dependencies in Python or Go, but in Python it is possible to use `get_robot_part` to return information including the whole configuration of a machine part, and then access the configuration attributes of the dependency from there.
```
cheukt, line 226: what does this mean? I think we would want to discourage users from looking at configs for other resources

cheukt (commented):
did a quick pass, I'll be out next week so no need for my approval - if we want an opinion from netcode we can add @benjirewis

docs/operate/get-started/other-hardware/dependencies.md
JessamyT: We got a chatbot question about this specifically (from Sean Yu) and the bot answer was incorrect so was hoping to fix it. This hidden content short code means it’s not visible on the page, only to the chatbot if someone specifically asks. But if we don’t want people to do this at all should I just change this to say “you can’t, period”?

cheukt: ok, I guess in that case this is fine - we should definitely discourage it tho

benjirewis (commented):
This is awesome! You guys do a great job of adding stuff around this somewhat opaque logic. Some comments and I can help further while Cheuk is on PTO.

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,68 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+1. The modular resource is ready to use.
+1. If at any point the user changes the configuration of the machine, `viam-server` reconfigures the affected resources within 15 seconds.
+1. If a resource crashes, `viam-server` attempts to rebuild it.
+1. When `viam-server` shuts down, it attempts to stop all running modules.
+   If a module fails to shut down within 90 seconds, `viam-server` will forcefully kill it after 15 seconds.
```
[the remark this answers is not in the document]
cheukt: The SIGTERMs are on a different timer, they are only triggered when we actually get to the clean module shutdown phase.
SIGKILLs are triggered after 90s from when the restart/shutdown is started, and the SIGTERMS can happen at any time between the shutdown is started and 90s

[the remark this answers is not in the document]
benjirewis: Ah I forgot about those `SIGKILL`s thanks @cheukt. So, yeah, slightly complicated "killing" logic (because we can SIGKILL a module due to it taking 90s since `viam-server` started shutdown _or_ it taking 30s since `viam-server` tried to start shutting down a single module, right?)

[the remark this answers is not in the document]
cheukt: ya, you're right about that - I forgot about the SIGKILL after 30s from the viam-server trying to shutdown a module.

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,68 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+Modules run on your machine, alongside `viam-server` as separate processes, communicating with `viam-server` over UNIX sockets.
+
+[`viam-server` manages](/operate/reference/viam-server/) the dependencies, start-up, reconfiguration, [data management](/data-ai/capture-data/capture-sync/), and shutdown behavior of your modular resource.
+
+The lifecycle of a module and the resources it provides is as follows:
+
+1. `viam-server` starts, and if it is connected to the internet, it checks for configuration updates.
+1. `viam-server` starts any configured modules.
+1. When a module initializes, it registers its model or models and associated [APIs](/dev/reference/apis/) with `viam-server`, making the models available for use.
+1. For each modular resource configured on the machine, `viam-server` uses the resource's `validate` function and the `depends_on` field in the resource configuration to determine the dependencies of the resource.
```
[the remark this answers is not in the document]
JessamyT: Yeah...I figured since the work hasn't started on implementing it it makes sense to try and get this merged with what is true now. Any idea on timeline for this change though?

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,68 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+1. If a dependency is not already running, `viam-server` starts it before starting the resource.
+   If a dependency is not found or fails to start, `viam-server` will not start the resource that depends on it.
+1. `viam-server` builds the resource based on its resource configuration.
+1. If configuration fails due to a validation failure or an exception thrown by the reconfigure function, `viam-server` attempts to reconfigure the resource.
+1. The modular resource is ready to use.
+1. If at any point the user changes the configuration of the machine, `viam-server` reconfigures the affected resources within 15 seconds.
+1. If a resource crashes, `viam-server` attempts to rebuild it.
```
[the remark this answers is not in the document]
JessamyT: Okay I'll delete this!
```suggestion
```

docs/operate/get-started/other-hardware/dependencies.md
```diff
@@ -0,0 +1,251 @@
...
+
+1. In your modular resource's `Validate` method, check the configuration attributes, then add the dependency name to the list of dependencies:
+
+   ```go {class="line-numbers linkable-line-numbers"}
+   func (cfg *Config) Validate(path string) ([]string, error) {
+     var deps []string
+     if cfg.CameraName == "" {
+       return nil, resource.NewConfigValidationFieldRequiredError(path, "camera_name")
+     }
+     if reflect.TypeOf(cfg.CameraName).Kind() != reflect.String {
```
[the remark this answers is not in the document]
JessamyT: Does it get checked by viam-server automatically because <code>CameraName string `json:"camera_name"`</code> specifies type?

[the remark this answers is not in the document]
benjirewis: Basically yep! It's not technically `viam-server` that does the conversion from proto to a golang string, but the Golang module library code does it and will error if the thing specified in JSON is not actually a string/`cfg.CameraName` will just `""`.

docs/operate/get-started/other-hardware/_index.md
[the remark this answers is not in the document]
JessamyT: So, this?
```suggestion
   When `viam-server` shuts down, it first shuts down all modules and kills them if they fail to shut down within 90 seconds.
   If `viam-server` attempts to shut down an individual module (for example due to a user disabling a module) and the module does not shut down within 30 seconds, `viam-server` kills the module.
```

docs/operate/get-started/other-hardware/dependencies.md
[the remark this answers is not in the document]
JessamyT: Ah okay ty!! Wait it'll error or it'll assign it an empty string or both?

michaellee1019 (approved):
Read through it and LGTM! A great improvement. I would double check on that comment thread with net code about the exact shutdown behavior. It might warrant a quick explanation about the overall timing, and then a more detailed section with the caveats about the shutdown timing based on the modules behavior. For example:

```When `viam-server` shuts down, it attempts to stop all running modules.
   This can take up to x seconds depending on the response time required by a module to shut down.```

And then a section that describes the shutdown and killing behavior.

File: docs/operate/get-started/other-hardware/_index.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (JessamyT) [reply docs/operate/get-started/other-hardware/_index.md:72]: Benji okay'ed this^ via Slack

**Comment by JessamyT** [comment]
Thanks!

Benji okayed the suggestion on that thread via slack so went ahead and committed now

**Review by benjirewis — APPROVED** [review]
All LGTM!

File: docs/operate/get-started/other-hardware/_index.md [hunk 8]
```diff
@@ -51,48 +51,69 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+1. `viam-server` starts, and if it is connected to the internet, it checks for configuration updates.
+1. `viam-server` starts any configured modules.
+1. When a module initializes, it registers its model or models and associated [APIs](/dev/reference/apis/) with `viam-server`, making the models available for use.
+1. For each modular resource configured on the machine, `viam-server` uses the resource's `validate` function and the `depends_on` field in the resource configuration to determine the dependencies of the resource.
+1. If a dependency is not already running, `viam-server` starts it before starting the resource.
+   If a dependency is not found or fails to start, `viam-server` will not start the resource that depends on it.
+1. `viam-server` calls the resource's constructor to build the resource based on its configuration.
+   Typically, the constructor calls the reconfigure function.
+1. If configuration fails due to a validation failure or an exception thrown by the reconfigure or constructor functions, `viam-server` attempts to reconfigure the resource.
+1. The modular resource is ready to use.
```
npentrel, line 69: This should also cover the failure scenario.  How often does it retry? What happens if it fails? Otherwise it's a bit odd to go from here to the next step. 
What makes it ready to use? If reconfigure succeeds? Can we be more specific?

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,69 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+1. If at any point the user changes the configuration of the machine, `viam-server` reconfigures the affected resources within 15 seconds.
+1. When `viam-server` shuts down, it attempts to stop all running modules.
```
npentrel, line 71: attempts? I mean it does stop them regardless. It attempts to shut them down gracefully and otherwise just kills them

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,69 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+   If `viam-server` attempts to shut down an individual module (for example due to a user disabling a module) and the module does not shut down within 30 seconds, `viam-server` kills the module.
+1. When `viam-server` shuts down, it first shuts down all modules and kills them if they fail to shut down within 90 seconds.
```
npentrel, line 73: I'm confused how the 30 and 90 seconds interact. If I have 3 modules that each take 30 seconds then any modules after that would just be killed?

docs/operate/get-started/other-hardware/_index.md
JessamyT: Maybe this will make more sense if I remove this sentence since the next sentence says this.

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -51,48 +51,69 @@ For C++ module examples, see the [C++ examples directory on GitHub](https://gith
...
+1. If a dependency is not already running, `viam-server` starts it before starting the resource.
+   If a dependency is not found or fails to start, `viam-server` will not start the resource that depends on it.
+1. `viam-server` calls the resource's constructor to build the resource based on its configuration.
+   Typically, the constructor calls the reconfigure function.
+1. If configuration fails due to a validation failure or an exception thrown by the reconfigure or constructor functions, `viam-server` attempts to reconfigure the resource.
+1. The modular resource is ready to use.
+1. If at any point the user changes the configuration of the machine, `viam-server` reconfigures the affected resources within 15 seconds.
+1. When `viam-server` shuts down, it attempts to stop all running modules.
+   If `viam-server` attempts to shut down an individual module (for example due to a user disabling a module) and the module does not shut down within 30 seconds, `viam-server` kills the module.
```
JessamyT, line 72:
```suggestion
1. If `viam-server` attempts to shut down an individual module (for example due to a user disabling a module) and the module does not shut down within 30 seconds, `viam-server` kills the module.
```

docs/operate/get-started/other-hardware/_index.md
JessamyT: Deleted a sentence that was essentially duplicated. Does it make more sense now? I can also tweak this.

My understanding is that `viam-server` tries to close all modules simultaneously as soon as `viam-server` is told to shut down. So the number of modules shouldn't matter. I know dependencies influence order of startup but do they influence order of shutdown @benjirewis ?
```suggestion
1. When `viam-server` shuts down, it first attempts to shut down all modules, and kills them if they fail to shut down within 90 seconds.
```

docs/operate/get-started/other-hardware/_index.md
JessamyT: @benjirewis If reconfiguration happens within 15 seconds, does that also mean that initial configuration is retried every 15 seconds until it times out?
```suggestion
1. If configuration fails due to a validation failure or an exception thrown by the reconfigure or constructor functions, `viam-server` attempts to reconfigure the resource every 15 seconds.
   If the module exceeds the [configured timeout limits](/operate/reference/module-configuration/#environment-variables) (default 5 minutes to start up and 1 minute to reconfigure), `viam-server` logs an [error](/dev/tools/common-errors/#timed-out-waiting-for-module).
1. Once the modular resource has started up and configured, it is available for use.
```

npentrel: mentioning the env variables is a good idea

docs/operate/get-started/other-hardware/_index.md
JessamyT: I believe 30s is for when viam-server is only shutting down an individual module, not shutting down all of viam-server. But when viam-server starts shutting itself down, there's a 90s opportunity for modules to shut down instead of only 30s. https://github.com/viamrobotics/docs/pull/4156/files/bb8a64404d0266bc8f8f33595a08fccb7ba769a2#r2019210281

> So, yeah, slightly complicated "killing" logic (because we can SIGKILL a module due to it taking 90s since viam-server started shutdown or it taking 30s since viam-server tried to start shutting down a single module, right?)

docs/operate/get-started/other-hardware/_index.md
benjirewis: > @benjirewis If reconfiguration happens within 15 seconds, does that also mean that initial configuration is retried every 15 seconds until it times out?

So we have a ticker that checks for "unconfigured" resources (never constructed) every _5_ seconds. We'll try to construct a not-yet-successfully-constructed modular resource every _5_ seconds as long as `viam-server` and the module are alive and the resource is not successfully constructing. Same thing goes for reconfigure and a not-yet-successfully-reconfigured modular resource.

docs/operate/get-started/other-hardware/_index.md
benjirewis: The current phrasing you have is correct.

> My understanding is that viam-server tries to close all modules simultaneously as soon as viam-server is told to shut down. So the number of modules shouldn't matter. I know dependencies influence order of startup but do they influence order of shutdown @benjirewis ?

`viam-server` does _not_ close all modules simultaneously. It does it one by one (sequentially) in a somewhat random order. Dependencies do not influence order of shutdown of resources or modules in any way.

> I'm confused how the 30 and 90 seconds interact. If I have 3 modules that each take 30 seconds then any modules after that would just be killed?

Correct. If `viam-server` attempted to shut down three modules and all three took >= 30s each to shutdown, all three would be shutdown by a SIGKILL. All modules after those three would _also_ be shutdown by a SIGKILL and then `viam-server` would exit.

docs/operate/get-started/other-hardware/dependencies.md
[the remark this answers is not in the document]
benjirewis: Sorry for ambiguity! There will be an error during validation for the resource if `camera_name` is not a string in the JSON.

docs/operate/get-started/other-hardware/_index.md
JessamyT: Am I erroneously conflating construction and reconfiguration? I've thought of them as conjoined since the constructor calls reconfigure right? Should the wording of the suggestion above change @benjirewis ?

docs/operate/get-started/other-hardware/_index.md
benjirewis: That looks great to me!

docs/operate/get-started/other-hardware/_index.md
benjirewis: Yeah if you want to be super pedantic it would be:
```
If construction fails due to a validation failure or an exception thrown by the modular resource's constructor, `viam-server` attempts to construct the resource every 5 seconds.

Similarly, if reconfiguration fails due to a validation failure or an exception thrown by the modular resource's reconfigure method, `viam-server` attempts to reconfigure the resource every 5 seconds.
```

npentrel (approved):

PR #4176, opened 2025-04-01
fix: wrap app.viam.com link in anchor tag

HipsterBrown:
The link to the Viam app in the docs navbar doesn't appear to work anymore. I'm not sure how it was working in the past since it was only a button with no anchor tag. To keep this from being interrupted by JS issues, I've updated the button to wrap the `app.viam.com` and icon in an anchor tag instead of a span to make it a working link.


npentrel (approved):
Thanks!

PR #4177, opened 2025-04-01
DOCS-3736: Update parts.md

npentrel:


docs/operate/reference/architecture/parts.md
```diff
@@ -97,13 +97,13 @@ To establish a connection between a part of one machine and a part of a second m
    {{<imgproc src="/build/configure/parts/remote-create.png" resize="x1100" declaredimensions=true alt="The create menu with options including remote part shown." style="width:500px" class="shadow" >}}
 
 7. Click **Remote part**.
-8. Find the newly-created remote part in the left hand menu.
-   Click it to navigate to its configuration card.
-9. Delete the auto-populated JSON from the new remote's config area and replace it by pasting the remote config you copied in step 4 into the empty field.
+8. Select from the list of machines.
```
JessamyT, line 100: Worth clarifying which machine they should select, i.e.
```suggestion
8. Select the remote part from the list of parts.
```
and is it a list of parts or machines?

docs/operate/reference/architecture/parts.md
```diff
@@ -97,13 +97,13 @@ To establish a connection between a part of one machine and a part of a second m
...
+   Alternatively, click **Add empty remote** and then scroll to the newly-created remote part configuration card.
+   Click on **{}** (Switch to advanced) and replace the JSON object with the remote config you copied in step 4.
```
JessamyT, line 102: Why merge into one step? Aren't we moving towards more smaller steps?

npentrel: Because if we selected from the list of parts we don't need any of the other steps. I'm including the rest only for people who add a remote part that isn't in the same location

JessamyT (approved):

PR #4178, opened 2025-04-01
DOCS-3816: Add workflow permissions to update module instructions

JessamyT:


npentrel (commented):
Why are you adding this? I've used the gh workflows and have not had to do this for my repositories

JessamyT:
> Why are you adding this? I've used the gh workflows and have not had to do this for my repositories

Oh really? I’ve always had to do it, and with the micro stuff that is in the works, Esha had to do it with those as well (that’s separate but seemingly indicative of what I ran into)

JessamyT:
Perhaps this is because I was in my own private repo vs in viam-labs or somesuch? I could move this into a caveat/troubleshooting tip since repos may have different defaults

JessamyT:
> (And I think a step is fine, probably most people in an org will then need to take this option that makes sense. And we do want to reduce expanders I think)

For this and the wifi PR, you've mentioned reducing expanders but this is an alert. Do you mean that we want to reduce both expanders and alerts?

npentrel:
> > (And I think a step is fine, probably most people in an org will then need to take this option that makes sense. And we do want to reduce expanders I think)
> 
> For this and the wifi PR, you've mentioned reducing expanders but this is an alert. Do you mean that we want to reduce both expanders and alerts?

Yes. Everything that adds a box should really only be there if necessary.

I think we still need to clarify under which condition this would require write permissions (as I said, it works without for https://github.com/npentrel/control-logic-example-module so I am not even sure if that is true)

JessamyT:
I tried replicating on one of my private modules and can't...I'll close and will reinvestigate if it comes up again.

PR #4179, opened 2025-04-01
DOCS-3207: Remove xArm arm model pages

JessamyT:


npentrel (changes requested):
You need to update the hash key for netlify if you're removing pages otherwise the site will not build once you merge this.

npentrel (approved):

PR #4185, opened 2025-04-02
DOCS-3760: Fix list width

JessamyT:
Still slightly wider than it should be on mobile width but seems significantly better


npentrel (changes requested):
still grows out of bounds

JessamyT:
Closing since we're going with expanders for complex content for now

PR #4186, opened 2025-04-03
DOCS-3696: Add Micro-RDK multiple wifi config info

JessamyT:


mattjperez (approved):
LGTM!

static/include/wifi-micro.md
npentrel, line 1: I think it would be better to follow the same pattern as for methods where we add a column with whether the option is supported in the micro rdk. That also helps us limit expanders a bit more. Thanks for changing the booleans

JessamyT: We were thinking the note would keep the table less cluttered/more legible, and no expanders were used here! I've moved it into the table though--lmk what you think now.

npentrel (approved):

PR #4187, opened 2025-04-03
DOCS-2930: Add first_run flow for Docker modules

JessamyT:
https://deploy-preview-4187--viam-docs.netlify.app/operate/reference/advanced-modules/#package-and-deploy-using-docker

### DIFF
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -56,6 +56,7 @@ Use cases for this include:
 - You have specific security requirements that are difficult to meet with the default module deployment.
 
 If you choose to deploy your module using Docker, we recommend creating a "first run" script or binary to run any necessary setup steps.
+Note this is _not_ recommended for modules that do not use Docker, it adds unnecessary complexity.
 
 {{% expand "Click for first run script instructions" %}}
 
@@ -67,11 +68,6 @@ If you choose to deploy your module using Docker, we recommend creating a "first
 
      ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
      #!/usr/bin/env bash
-
-     if [[ -n "$VIAM_TEST_FAIL_RUN_FIRST" ]]; then
-         echo "Sorry, I've failed you."
-         exit 1
-     fi
 
      docker pull mongo:6
 
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -51,11 +51,11 @@ However, if you are unable to use modular resources because you need to host `vi
 In rare cases, you may need to package and deploy a module using Docker.
 Use cases for this include:
 
-- Your module has complex system dependencies that cannot be installed on a machine.
+- Your module has complex system dependencies that cannot be easily installed on a machine.
 - You have a large bundle and want to use layer caching to reduce the size of the download.
 - You have specific security requirements that are difficult to meet with the default module deployment.
 
-If you choose to deploy your module using Docker, we recommend creating a first run script or binary to run any necessary setup steps.
+If you choose to deploy your module using Docker, we recommend creating a "first run" script or binary to run any necessary setup steps.
 
 {{% expand "Click for first run script instructions" %}}
 
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -77,12 +77,7 @@ If you choose to deploy your module using Docker, we recommend creating a "first
 
      cat << EOF
      -------------------------------------
-     Congratulations!
      The setup script ran successfully!
-     This message is obnoxiously large for
-     testing purposes.
-     Sincerely,
-     First Run Script
      -------------------------------------
      EOF
 
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -62,7 +62,13 @@ Note this is _not_ recommended for modules that do not use Docker, it adds unnec
 
 1. Create a tarball that contains:
 
-   - The module's entrypoint script or binary
+   - The module's entrypoint script or binary, for example:
+
+     ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
+     #!/bin/bash
+     exec docker run <YOUR_CONTAINER_IMAGE> <YOUR_CONTAINER_OPTIONS>
+     ```
+
    - A <file>meta.json</file>
    - A first run script or binary that will be executed during the setup phase, for example:
 
@@ -93,14 +99,16 @@ Note this is _not_ recommended for modules that do not use Docker, it adds unnec
 
 1. Configure your module on your machine in the same way as you would for a regular module.
    The first run script will execute once when `viam-server` receives a new configuration.
-   It will only execute once per module or per version of the module.
+   It will only execute once per module or per version of the module.<br><br>
 
-1. (Optional) After a first run script runs successfully, Viam adds a marker file with a `.first_run_succeeded` suffix in the module’s data directory on disk.
-   It has the location and form: `/root/.viam/packages/data/module/<MODULE_ID>-<VERSION>-<ARCH>.first_run_succeeded`.
-   If you want to force a first run script to run again without changing the configured module or module version, you can do so by deleting this file.
+1. (Optional) After a first run script runs successfully, Viam adds a marker file in the module's data directory.
+   The marker file path format is of the form `unpackedModDir + FirstRunSuccessSuffix`.
+   For example, `.viam/packages/data/module/abcd1234-abcd-abcd-abcd-abcd12345678-viam-rtsp-0_1_0-linux-amd64/bin.first_run_succeeded`.
+
+   If you want to force a first run script to run again without changing the configured module or module version, you can do so by deleting this file.<br><br>
 
 1. (Optional) By default, a first run script will timeout after 1 hour.
-   This can be adjusted by adding a `first_run_timeout` to the module’s configuration.
+   This can be adjusted by adding a `first_run_timeout` to the module's configuration.
    For example, `"first_run_timeout": "5m"` will lower the script timeout to 5 minutes.
 
 {{% /expand %}}
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -52,7 +52,7 @@ In rare cases, you may need to package and deploy a module using Docker.
 Use cases for this include:
 
 - Your module has complex system dependencies that cannot be easily installed on a machine.
-- You have a large bundle and want to use layer caching to reduce the size of the download.
+- You use a large container image and some layers are already used by your machine which means layer caching can reduce the size of the download.
 - You have specific security requirements that are difficult to meet with the default module deployment.
 
 If you choose to deploy your module using Docker, we recommend creating a "first run" script or binary to run any necessary setup steps.
--- a/docs/operate/reference/advanced-modules/_index.md
+++ b/docs/operate/reference/advanced-modules/_index.md
@@ -11,7 +11,7 @@ tags:
     "components",
     "services",
   ]
-description: "Some usage may require you to define new APIs or deploy custom components using a server on a remote part"
+description: "Some usage may require you to define new APIs or deploy custom components in non-standard ways."
 aliases:
   - /program/extend/
   - /modular-resources/advanced/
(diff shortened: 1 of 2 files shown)


docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
 
 However, if you are unable to use modular resources because you need to host `viam-server` on a non-Linux system or have an issue with compilation, you may need to [implement a custom component and register it on a server configured as a remote](/operate/reference/advanced-modules/custom-components-remotes/) on your machine.
 
+## Package and deploy using Docker
+
+In rare cases, you may need to package and deploy a module using Docker.
+Use cases for this include:
+
+- Your module has complex system dependencies that cannot be installed on a machine.
```
penguinland, line 54: *that cannot _easily_ be installed on a machine.

If they can be installed in docker, they can be installed on the machine. It just might be a giant pain.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+- You have a large bundle and want to use layer caching to reduce the size of the download.
```
penguinland, line 55: The world "bundle" seems odd, but I can't think of a better word. If no one else can think of a better word either, leave as-is.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+- You have specific security requirements that are difficult to meet with the default module deployment.
+
+If you choose to deploy your module using Docker, we recommend creating a first run script or binary to run any necessary setup steps.
```
penguinland, line 58: I'd like to better indicate that "first run" is a single thing, so no one mistakes it for, say, the first of many different run scripts or something. Would putting quotes around "first run" be appropriate? What about an underscore in first_run? I dunno, maybe leave as-is if that's standard in the docs.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+1. Create a tarball that contains:
+
+   - The module's entrypoint script or binary
+   - A <file>meta.json</file>
+   - A first run script or binary that will be executed during the setup phase, for example:
+
+     ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
+     #!/usr/bin/env bash
+
+     if [[ -n "$VIAM_TEST_FAIL_RUN_FIRST" ]]; then
```
penguinland, line 71: I'm not familiar with this env var. Is it supposed to be in all first_run scripts? What does it do? Maybe I should add it to mine... :flushed:

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+     "first_run": "first_run.sh"
+   }
+   ```
+
+1. Configure your module on your machine in the same way as you would for a regular module.
+   The first run script will execute once when `viam-server` receives a new configuration.
+   It will only execute once per module or per version of the module.
+
+1. (Optional) After a first run script runs successfully, Viam adds a marker file with a `.first_run_succeeded` suffix in the module’s data directory on disk.
+   It has the location and form: `/root/.viam/packages/data/module/<MODULE_ID>-<VERSION>-<ARCH>.first_run_succeeded`.
```
penguinland, line 108: I suspect there should be a slash before `.first_run_succeeded`, but haven't checked on that personally. Maybe it's correct as-is.

docs/operate/reference/advanced-modules/_index.md
JessamyT:
```suggestion
- Your module has complex system dependencies that cannot be easily installed on a machine.
```

docs/operate/reference/advanced-modules/_index.md
JessamyT:
```suggestion
If you choose to deploy your module using Docker, we recommend creating a "first run" script or binary to run any necessary setup steps.
```

docs/operate/reference/advanced-modules/_index.md
JessamyT: Great question--just pulled this in from the setup phase guide on confluence which in turn links to https://github.com/viam-labs/wifi-sensor/blob/7823b6ad3edcbbbf20b06c34b3181453f5f3f078/first_run.sh

docs/operate/reference/advanced-modules/_index.md
JessamyT: Hmm also from Maxim's guide--not sure if there's a reasonably practical way to check for sure

JessamyT: Eep! I don't have any Docker modules to test this on... @cheukt since it looks like you've been maintaining the confluence instructions, can you confirm?

docs/operate/reference/advanced-modules/_index.md
npentrel: I suppose that depends on what's meant by bundle? Does the layer caching help because an image with the same base container and possibly some layers can be used between different installed modules? That's the only way I can think this would be helpful.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+
+   - The module's entrypoint script or binary
+   - A <file>meta.json</file>
+   - A first run script or binary that will be executed during the setup phase, for example:
+
+     ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
+     #!/usr/bin/env bash
+
+     if [[ -n "$VIAM_TEST_FAIL_RUN_FIRST" ]]; then
+         echo "Sorry, I've failed you."
```
npentrel, line 72: Does this error indicate an issue with the module? Maybe better to change to something that indicates where the error is

docs/operate/reference/advanced-modules/_index.md
npentrel: I also am unclear on what this is for here. Can you add an explanation in? Does this get set to true if a previous run failed? Or is this jsut for testing?

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+     docker pull mongo:6
+
+     cat << EOF
+     -------------------------------------
+     Congratulations!
+     The setup script ran successfully!
+     This message is obnoxiously large for
+     testing purposes.
+     Sincerely,
+     First Run Script
```
npentrel, line 85: Is it necessary to show a large message for the docs?

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,74 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+
+     ```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
+     #!/usr/bin/env bash
+
+     if [[ -n "$VIAM_TEST_FAIL_RUN_FIRST" ]]; then
+         echo "Sorry, I've failed you."
+         exit 1
+     fi
+
+     docker pull mongo:6
```
npentrel, line 76: So this pulls the image but then how does the container get run? I think we probably want to link to an example of that. Also 

And should this run with https://github.com/viam-soleng/viam-docker-manager or is this entirely separate?

npentrel (changes requested):
I'm a bit confused at a few points. 
I also think it would be really helpful to link to a full example. I think soleng used docker in a few projects, is there none we can link to?

docs/operate/reference/advanced-modules/_index.md
penguinland: In the RDK, this env var is mentioned in 2 places: a test and an example first_run.sh. I have cloned about 60 Viam-related repos locally, and it's not mentioned in any of the others. I've searched Github for it, and can only find copies of those 2 locations. I wonder if there was some plan to add this in, but it never happened and these are just the vestiges of a feature that never was.

I think we should remove this paragraph from the example.

docs/operate/reference/advanced-modules/_index.md
penguinland: I think this example is right: in first_run.sh, you pull the image, and then in your module's main executable, you run the image, and that second part should be obvious for anyone developing with docker. We could add in such an example, but it's trivial:
```bash
#!/bin/bash
set -euxo pipefail
# This example module runs inside the docker image you should have downloaded in first_run.sh
SOCKET_DIR=`dirname $1`
VIAM_DIR=`realpath ~/.viam`
exec docker run \
     --rm \
     --runtime=nvidia --gpus=all \
     -v /etc/passwd:/etc/passwd \
     -v /etc/group:/etc/group \
     -v $SOCKET_DIR:$SOCKET_DIR \
     -v $VIAM_DIR:$VIAM_DIR \
     ghcr.io/viam-modules/viam-mlmodelservice-triton/jetpack6:0.9.0 \
     LD_PRELOAD=libjemalloc.so.2 VIAM_MODULE_DATA="$VIAM_MODULE_DATA" /opt/viam/bin/viam_mlmodelservice_triton "$@" 2>&1
```

penguinland: (that example is from the [Triton module](https://github.com/viam-modules/viam-mlmodelservice-triton/tree/main), and although there are lots of options specified for the `docker` line, anyone developing with docker should know which ones they need to set.)

docs/operate/reference/advanced-modules/_index.md
penguinland: The people who love docker will talk your ear off about how useful layer caching is: the idea is that if you have a bunch of small layers that add specific things to your docker image, you can reuse these layers in different projects to get different images without needing to create each thing multiple times. This is an argument for docker use in general, not something Viam-specific.

I think the bundle is supposed to refer to the collection of layers. I don't know enough about docker to know if that's the standard term; it might be. Even if it's not, I think the meaning is clear by the end of the sentence, I just stumbled a little in the middle.

docs/operate/reference/advanced-modules/_index.md
JessamyT: I will reduce it. This was Max's original message and thought it might be okay to bring some amusement to the contents of an expander, further hidden within a hidden page, but no it doesn't need to be here.
```suggestion
     The setup script ran successfully!
```

docs/operate/reference/advanced-modules/_index.md
JessamyT: I am not sure where it is. Since this is an edge case documented within a hidden page, I figured it made sense to act quickly and use the existing example provided by Maxim. Sounds like I'll remove the whole paragraph anyway.

docs/operate/reference/advanced-modules/_index.md
JessamyT: Thanks for looking around. Will remove.

docs/operate/reference/advanced-modules/_index.md
JessamyT: @abe-winter I borrowed your words that were perhaps not intended for official documentation. Is there a word other than "bundle" you'd recommend here?

JessamyT:
RE examples to point to, it looks like https://github.com/abe-winter/python-container-module doesn't actually use a first_run script as such? And https://github.com/viam-labs/wifi-sensor/tree/RSDK-8886-setup-phase is what Maxim pointed to as an example but the main branch of wifi-sensor doesn't use it. Based on [the thread](https://viaminc.slack.com/archives/C081S6K8BS4/p1743549371738299?thread_ts=1743546486.180029&cid=C081S6K8BS4) it sounds like the sol-eng uses would've been a bit atypical. @abe-winter do you recommend a particular example? Or @penguinland is https://github.com/viam-modules/viam-mlmodelservice-triton/tree/main like you mentioned a good one, and if so, should the examples in this PR be changed to match it?

penguinland:
I've been mentioning the Triton one because that's the only docker module I work with. TBH, it's kind of a pain: the way Drew set it up, in the repo there are merely _templates_ of the module script and first_run script, and you need to generate the actual ones from those templates. My hope is there's a simpler example out there, but I'm not aware of one.

docs/operate/reference/advanced-modules/_index.md
abe-winter: maybe 'artifact'?

npentrel: Right so it's useful if you have a large image and you use some of the same layers already on the machine so that they'll already be present. So that is the use case. Otherwise there wouldn't be anything reducing the size of the download. 

```suggestion
- You use a large container image and some layers are already used by your machine which means layer caching can reduce the size of the download.
```

docs/operate/reference/advanced-modules/_index.md
cheukt: it is `unpackedModDir + FirstRunSuccessSuffix`, an example would be `.viam/packages/data/module/e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb-viamrtsp-0_1_0-linux-amd64/bin.first_run_succeeded`
https://github.com/viamrobotics/rdk/blob/1af4d812c36e43c2f08b847290e1642489eb1df8/config/module.go#L279

docs/operate/reference/advanced-modules/_index.md
JessamyT: Okay, added above. Tested using the hello-world image, though it of course doesn't behave exactly like a real module that would continue to run. Did confirm that it ran the first_run and made a marker file.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,73 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
 
 However, if you are unable to use modular resources because you need to host `viam-server` on a non-Linux system or have an issue with compilation, you may need to [implement a custom component and register it on a server configured as a remote](/operate/reference/advanced-modules/custom-components-remotes/) on your machine.
 
+## Package and deploy using Docker
```
npentrel, line 49: For both Custom APIs and components as remote parts, we have subpages. I think we should do the same here.

docs/operate/reference/advanced-modules/_index.md
```diff
@@ -46,6 +46,73 @@ Running {{< glossary_tooltip term_id="modular-resource" text="modular resources"
...
+
+In rare cases, you may need to package and deploy a module using Docker.
```
npentrel, line 51: Be aware that we are not providing complete info here. We're showing how to pull and run a container. We will likely need to look into how to package a module with Docker. So far this really only covers the deploy aspect.

Please file a follow up ticket to cover this. A full example for reference is really what I think we need here.

npentrel (changes requested):
Let's move this into a subpage. It can be referencd from the current page but the rest should be a subpage.

Thank you for testing it. Please do also file a follow up ticket for yourself to test packaging an actual module with docker. Currently the docs really only cover the deploying part.

docs/operate/reference/advanced-modules/_index.md
JessamyT: https://viam.atlassian.net/browse/DOCS-3863

JessamyT:
> Please do also file a follow up ticket for yourself to test packaging an actual module with docker. Currently the docs really only cover the deploying part.

To be clear, do you mean an additional ticket separate from and before adding the full example how-to, or as part of that? I'd like to do as part of https://viam.atlassian.net/browse/DOCS-3863 whenever that happens.

npentrel (approved):

PR #4194, opened 2025-04-07
DOCS-3806: Update front page

npentrel:


JessamyT (approved):
I'd still say it'd be good to shorten the descriptions on the three cards at the bottom so it's not the same copy twice, and maybe make those three hover boxes clickable instead of "Learn more" since "Learn more" isn't action-oriented. Not blocking though if you want to merge and iterate later.

npentrel:
> I'd still say it'd be good to shorten the descriptions on the three cards at the bottom so it's not the same copy twice, and maybe make those three hover boxes clickable instead of "Learn more" since "Learn more" isn't action-oriented. Not blocking though if you want to merge and iterate later.

Yeah let's iterate later. This was what design wanted so I don't want to stray too far from it until we redo the cotnent for this page

PR #4195, opened 2025-04-07
DOCS-3825: Add component TypeScript examples

JessamyT:


static/include/components/apis/generated/gantry.md
```diff
@@ -59,7 +59,16 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/c
 
 **Returns:**
 
-- (Promise<number[]>): The current position of each axis.
```
npentrel, line 62: @JessamyT Did you remove this on purpose? If so why?

npentrel (commented):
Small comment - otherwise LGTM

static/include/components/apis/generated/gantry.md
JessamyT: Mm so I was following the pattern of the camera API samples, and for TS and Flutter it looks like we generally don't have descriptions on returned objects, e.g. https://docs.viam.com/dev/reference/apis/components/camera/#getimage. Would you like me to make tickets to add return descriptions throughout the TS and Flutter SDK docs? For the purposes of this PR here I can revert this change and the one other one since there's no need to remove this description, though it is just reiterating the method description.

static/include/components/apis/generated/gantry.md
```diff
@@ -59,7 +59,16 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/c
...
+- (Promise<number[]>)
```
JessamyT, line 62:
```suggestion
- (Promise<number[]>): The current position of each axis.
```

static/include/components/apis/generated/gantry.md
```diff
@@ -254,7 +276,16 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/c
 
 **Returns:**
 
-- (Promise<number[]>): The lengths of the axes of the gantry in millimeters.
+- (Promise<number[]>)
```
JessamyT, line 279:
```suggestion
- (Promise<number[]>): The lengths of the axes of the gantry in millimeters.
```

static/include/components/apis/generated/gantry.md
npentrel: I think they're not there because the SDK folks didn't have time to add them. That doesn't mean we should remove the places where they are for consistency. Ideally we want people to start adding more of these because they are nice. However, I do not want you to go through the entire TS and Flutter docs to add them now, no. Not unless AI allows you to do that within less than an hour.

JessamyT: Okay. Reverted here and will revert upstream for these two but won't worry about adding to others.

PR #4207, opened 2025-04-09
DOCS-3851: Update troubleshoot.md

npentrel:


docs/manage/troubleshoot/troubleshoot.md
```diff
@@ -25,9 +25,14 @@ Go to your machine's **CONFIGURE** page and check whether any configured compone
 If so click on the symbol or expand the **ERROR LOGS** panel.
 The expanded panel shows you errors produced by that resource.
 
+#### Machine shows as offline
+
+If your machine shows as offline in the Viam app, restart `viam-server` by running the command to start `viam-server` and adding the `-debug` option.
```
JessamyT, line 30: Add the command or a link to the command https://docs.viam.com/operate/reference/viam-server/manage-viam-server/#run-viam-server
Also `--debug` or `-d` but `-debug` doesn't work

Suggestion doesn't work with code fences but can edit and copy paste

```suggestion
If your machine shows as offline in the Viam app, restart `viam-server` by running the command to start `viam-server` and adding the `--debug` option:

{{< tabs >}}
{{% tab name="Linux" %}}

```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
sudo systemctl restart viam-server --debug
```

{{% /tab %}}
{{% tab name="macOS" %}}

```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
brew services restart viam-server --debug
```

{{% /tab %}}
{{< /tabs >}}
```

**Review by JessamyT — CHANGES_REQUESTED** [review]

File: docs/manage/troubleshoot/troubleshoot.md (hunk 1 above)
Comment (npentrel) [reply docs/manage/troubleshoot/troubleshoot.md:30]: Did you test this?  Both one or two dashes work for `viam-server` but the app setup instructions and the help messages for `viam-server` only use one. `-d` does not work. You can run `viam-server -help` do confirm this. 

You cannot pass arguments to `viam-server` through `systemctl` or `brew` like that. That's attempting to set debug for  `systemctl`/`brew`. `systemctl` doesn't support this, `brew` does but then you get debugging info for `brew` not `viam-server`. Also the default way to start `viam-server` on a mac (the one given in the app) does not use services so I don't think this makes sense to document here.

I'll add more info for how to restart `viam-server` with debug options but what I said here initially is correct as is.

Comment (JessamyT) [reply docs/manage/troubleshoot/troubleshoot.md]: Ah I see, so based on what you had here when I first reviewed, it wasn't clear where to pass the debug flag. Thus the suggestion to add more specifics for Linux and Mac. I misunderstood what was meant by "command to start `viam-server`", just as a user might.

I was wrong about the flag, my apologies.

File: docs/manage/troubleshoot/troubleshoot.md [hunk 2]
```diff
@@ -19,6 +19,95 @@ For common errors see [Common Errors](/dev/tools/common-errors/).
 
 ## Check logs
 
+### Machine shows as offline
+
+If your machine shows as offline in the Viam app, restart `viam-server` by running the command to start `viam-server` and adding the `-debug` option.
+
+To do this, you will need to know if you installed `viam-server` with `viam-agent` (most common) or manually.
```
JessamyT, line 26: There are mac instructions below. Do you want to also clarify up here that this doesn't apply for mac users, since agent doesn't run on mac?

npentrel: Mac users can run these instructions too if they don't immediately know.

JessamyT (approved):
LGTM % is there a reason the Linux instructions save logs to a file and the mac instructions do not?

npentrel:
> LGTM % is there a reason the Linux instructions save logs to a file and the mac instructions do not?

yes. If you run on mac the logs display on the terminal. You could also write it to a file but since the default would show the logs it's fine as is I think

PR #4210, opened 2025-04-10
Fix broken landing page image on Safari

nathan-contino:
The image on the right side of the landing page is awfully distorted on Safari on the production site right now, stretched vertically to roughly the page height. I believe the landing page is the only page that uses this style, and it displays nicely in Firefox and Chrome and Safari after this fix. But let me know if we have other ways to test and validate style changes.


npentrel (approved):
Thank you Nathan!

PR #4209, opened 2025-04-10
DOCS-3852: Update manage-viam-server.md

npentrel:

PR #4211, opened 2025-04-10
Further polish a wording issue I created in #4209

nathan-contino:
One of my suggestions in https://github.com/viamrobotics/docs/pull/4209 accidentally a word. This fixes that typo, polishes some nearby wording, and removes the 'command line' class and bash-prompt symbol from a non-interactive line.

### DIFF
--- a/docs/operate/reference/viam-server/manage-viam-server.md
+++ b/docs/operate/reference/viam-server/manage-viam-server.md
@@ -270,7 +270,7 @@ Then, reload the service file with the following command:
 sudo systemctl daemon-reload
 ```
 
-To resume automatic update checking, delete the leading `#` character in <file>/etc/systemd/system/viam-server.service</file>, then run `sudo systemctl daemon-reload` again to reload the service file.
+To resume automatic update checking, delete the leading `#` character in <file>/etc/systemd/system/viam-server.service</file>, then run `sudo systemctl daemon-reload` again.
 
 {{% /tab %}}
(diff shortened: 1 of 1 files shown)


docs/operate/reference/viam-server/manage-viam-server.md
```diff
@@ -270,7 +270,7 @@ Then, reload the service file with the following command:
 sudo systemctl daemon-reload
 ```
 
-To resume automatic update checking, delete the leading `#` character once more, and run `sudo systemctl daemon-reload` again.
+To resume automatic update checking, delete the leading `#` character in <file>/etc/systemd/system/viam-server.service</file>, then run `sudo systemctl daemon-reload` again to reload the service file.
```
JessamyT, line 273:
```suggestion
To resume automatic update checking, delete the leading `#` character in <file>/etc/systemd/system/viam-server.service</file>, then run `sudo systemctl daemon-reload` again.
```
[opt] In the interest of being concise, could omit this last bit since it does say the command reloads the service file just above the first instance. I *think* it doesn't need re-explanation so soon after.

JessamyT (approved):
LGTM % small comment

docs/operate/reference/viam-server/manage-viam-server.md
nathan-contino: Well spotted. Why use more word when few word do trick?

PR #4214, opened 2025-04-11
DOCS-3867: Add OS details to micro setup instructions

JessamyT:
Answer an AI chatbot question (since the bot doesn't have the ability to read the app content)
Also incorporate a lil more explanation of why we're sending them off to the app for setup

### DIFF
--- a/docs/operate/get-started/setup-micro.md
+++ b/docs/operate/get-started/setup-micro.md
@@ -43,6 +43,10 @@ To quickly try out Viam for microcontrollers, you can use the pre-built binary `
 
 ## Quickstart
 
+{{% hiddencontent %}}
+Viam provides installers to flash your ESP32 using macOS, Linux x86_64, or Linux ARM64 (AArch64).
+{{% /hiddencontent %}}
+
 To get started quickly with the pre-built `viam-micro-server` binary, follow these steps:
 
 1. Create a [Viam app](https://app.viam.com) account.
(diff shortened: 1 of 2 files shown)


nathan-contino (approved):
LGTM % one nitpicky suggestion.

docs/operate/get-started/setup-micro.md
```diff
@@ -43,6 +43,10 @@ To quickly try out Viam for microcontrollers, you can use the pre-built binary `
 
 ## Quickstart
 
+{{% hiddencontent %}}
+Viam provides installers to flash your ESP32 using macOS, Linux x86_64, or Linux ARM64 (AArch64).
```
nathan-contino, line 47: [nit] if we're going to call out the architecture for Linux installs, should we also call it out for macOS? afaik, quite a few Mac models with x86_64 processors are still actively supported and could easily be used by customers. Might be cleaner to say something like:

```suggestion
Viam provides installers to flash an ESP32 from macOS or Linux running on the x86_64 or ARM64 (AArch64) processor architectures.
```

It does feel a little dirty to conflate Apple Silicon with all other ARM64 architectures, but for the purposes of x86/arm disambiguation here I suspect it's OK.

JessamyT: The app just says this:
image
We haven't tended to distinguish between mac architectures. Should we?

PR #4192, opened 2025-04-04
DOCS-3828: Modularize RPi5 and renogy power sensor

JessamyT:
Also remove some other movement sensor pages that are no longer built-in

PR #4215, opened 2025-04-11
Update cachekey

JessamyT:
Fix check I broke with https://github.com/viamrobotics/docs/pull/4192


nathan-contino (approved):
LGTM, reaching out so i can better understand what i'm approving but this seems reasonable and low-risk enough

PR #4218, opened 2025-04-11
DOCS-3191 Add/remove tags on binary data

IanWhalen:
## Context
POC of using Claude to update docs.  TBH this looks like a real miss to me (I think we actually wanted to update https://docs.viam.com/dev/tools/cli/?) but I think that's because my vector search is what failed, not the interaction with claude:

```
2025-04-11 19:16:30 [info     ] Performing vector search       [ai_client] query='Updating tags on binary data, adding binary data to datasets, removing binary data from datasets, running ML inference on binary data' top_k=5
```
our vector query didn't actual include "CLI".

## Tool Output
--- Proposed Documentation Actions ---
Summary: The main changes are:
1) Removing the `org_id` and `location_id` flags when tagging, adding, or removing binary data. The binary data ID itself is now sufficient.
2) Renaming `file_id` and `file_ids` flags to `binary_data_id` and `binary_data_ids`.
3) Updating the inference command to use the `binary_data_id` flag instead of separate `file_id`, `file_org_id`, and `file_location_id` flags.
Reasoning:
The retrieved relevant documentation sections covered adding/removing tags from binary data by ID or by filter, but did not have the updated content reflecting the removal of `org_id` and `location_id` requirements, and the renaming of `file_id`/`file_ids` to `binary_data_id`/`binary_data_ids`.

To address the changes in the diff, I updated the content for `data.AddTagsToBinaryDataByIDs.md` and `data.RemoveTagsFromBinaryDataByIDs.md` to use the new `binary_data_ids` field instead of `file_ids`, `org_id`, and `location_id`. I preserved the existing descriptions but updated the proto message definitions to match the new API.

I did not find any existing documentation sections specifically covering the inference command changes, so I did not generate an edit for that portion of the diff. The reasoning is that by updating the tag adding/removing docs, users will be able to infer how to use the new `binary_data_id` flag for inference as well.
Planned Edits:
- UPDATE: static/include/app/apis/overrides/protos/data.AddTagsToBinaryDataByIDs.md (376 bytes)
- UPDATE: static/include/app/apis/overrides/protos/data.RemoveTagsFromBinaryDataByIDs.md (416 bytes)
--- End of Plan ---

### DIFF
--- a/static/include/app/apis/overrides/protos/data.AddTagsToBinaryDataByIDs.md
+++ b/static/include/app/apis/overrides/protos/data.AddTagsToBinaryDataByIDs.md
@@ -1,1 +1,16 @@
-Add tags to binary data by ids.
+# Add tags to binary data by IDs
+
+Add tags to binary data identified by their binary data IDs.
+
+Request:
+```protobuf
+message AddTagsToBinaryDataByIDsRequest {
+  // The tags to add.
+  repeated string tags = 1;
+
+  // The binary data IDs to add tags to.
+  repeated string binary_data_ids = 2;
+}
+```
+
+The binary data IDs in the request will have the specified tags added to them.
--- a/static/include/app/apis/overrides/protos/data.RemoveTagsFromBinaryDataByIDs.md
+++ b/static/include/app/apis/overrides/protos/data.RemoveTagsFromBinaryDataByIDs.md
@@ -1,1 +1,16 @@
-Remove tags from binary by ids.
+# Remove tags from binary data by IDs
+
+Remove tags from binary data identified by their binary data IDs.  
+
+Request:
+```protobuf
+message RemoveTagsFromBinaryDataByIDsRequest {
+  // The tags to remove.
+  repeated string tags = 1;
+   
+  // The binary data IDs to remove tags from.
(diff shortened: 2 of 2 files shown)


IanWhalen:
obviously I just expect this to get closed, but I thought it was at least interesting to include as an example of what this could look like.

IanWhalen:
also worth noting: this PR cost ~$1 to generate.

IanWhalen:
@npentrel now that I've had a chance to think on this, I do have a request for you here: 

could you look at https://viam.atlassian.net/browse/DATA-3948 as well as the broader epic and let me know if you think it makes sense to do a docs change on this [PR](https://github.com/viamrobotics/rdk/pull/4879/files) alone or if it would need to be on the broader epic?  Right now I'm trying to get to a reproducible eval case on step 1 in the pipeline - given an input of content (PRs, scopes, whatever), i want to make sure that we are identifying the correct files to update.

npentrel:
Yes I don't see the cli commands used anywhere else. For the broader epic we do also need to update the create dataset page

npentrel (changes requested):
(just to ensure this doesn't accidentally get merged)

PR #4230, opened 2025-04-16
Fix method overwrite error

npentrel:


.github/workflows/parse_typescript.py
```diff
@@ -152,6 +152,9 @@ def parse(self, type, viam_resources, args):
                     'return': return_object
                 }
 
+                if method_name == "restartModule":
+                    print(self.typescript_methods[type][resource][method_name])
```
JessamyT, line 156: nit: was this just for testing?

PR #4237, opened 2025-04-16
DOCS-3870: Document Windows viam-server root

JessamyT:
When running on WSL as root, VIAM_HOME was the root dir. Can't say for certain that it always appears in home and not in root if running not as root, but wasn't able to get viam-server to work not as root and I think further testing of WSL is overkill at this time.


docs/operate/reference/module-configuration.md
```diff
@@ -234,7 +234,7 @@ Not all of these variables are automatically available on [local modules](/opera
 <!-- prettier-ignore -->
 | Name | Description | Automatically set on local modules? |
 | ---- | ----------- | ----------------------------------- |
-| `VIAM_HOME` | The root of the `viam-server` configuration.<br>Default: `$HOME/.viam` | <p class="center-text"><i class="fas fa-check" title="yes"></i></p> |
+| `VIAM_HOME` | The root of the `viam-server` configuration.<br>Default for Linux, macOS, and WSL: `$HOME/.viam`<br>Default for Windows: `C:\WINDOWS\system32\config\systemprofile\.viam` | <p class="center-text"><i class="fas fa-check" title="yes"></i></p> |
```
nathan-contino, line 237:
```suggestion
| `VIAM_HOME` | The root of the `viam-server` configuration.<br>Default for Linux, macOS, and WSL: <file>$HOME/.viam</file><br>Default for Windows: <file>C:\WINDOWS\system32\config\systemprofile\.viam</file> | <p class="center-text"><i class="fas fa-check" title="yes"></i></p> |
```
[opt] I'm not sure if this syntax will work with all of the HTML afoot in this table, but I suspect it's Technically Correct to markup these locations as files, rather than monospace code.

nathan-contino (approved):
LGTM, a nice improvement for Windows native users

docs/operate/reference/module-configuration.md
JessamyT: You are correct, and both stylings do seem to render properly. I shall fix all the other file paths in this table too! TBH I wish our file styling had the same gray background as the code styling because it stands out better :/ But that's a low-prio convo to start haha

PR #4232, opened 2025-04-16
DOCS-3191: Update file IDs to binary IDs

npentrel:

PR #4238, opened 2025-04-17
DOCS-3191,DOCS-3879: Remove mention of outdated file-ids argument and updated associated config

nathan-contino:
Sniping DOCS-3191 from @npentrel (https://github.com/viamrobotics/docs/pull/4232) because a user experienced an issue with this CLI method and that led me to discover some other issues (like the fact that org-id and location-id are also no longer required when using the `ids` argument).


npentrel (approved):
Looks great - thanks!

PR #4243, opened 2025-04-21
DOCS-3895: Add meta.json to tarball

JessamyT:
A user asked about what files to scp to a pi when testing a local module on a separate computer. I tested both this and scp'ing the tarball plus the meta.json (with a generated Python module), and both work. I'm thinking meta.json should be in the tarball so it's all together, but is there a reason that'd actually not be preferred?


michaellee1019:
> I'm thinking meta.json should be in the tarball so it's all together, but is there a reason that'd actually not be preferred?

Yes this is preferred. [The generator does include meta.json in the case where a tarball is built instead of a binary.](https://github.com/viamrobotics/rdk/blob/main/cli/module_generate/_templates/go/tmpl-Makefile#L31)

npentrel (approved):

PR #4244, opened 2025-04-22
DOCS-3204: Add docs for add image to dataset button, incorporate into tutorials where helpful

nathan-contino:
* Updated an existing image of the TEST pane of the camera card to reflect the new button (and some other recent updates)
* Added a new image highlighting the *Add image to dataset* button since it's easy for users to miss (and difficult to describe) a button with a camera icon on it.
* Fixed some minor typos I noticed while trawling for pages where I should make updates.
* Reorganized 'create a dataset' page to better fit the narrative that fits in my head around creating a dataset, adding images to it, annotating it, and training a model on it.
  * Ironically, this page _still_ doesn't include instructions to... create a dataset. But it didn't in the first place. I think that task could fit nicely as a section after 'prerequisites' -- will add it and update this PR with that change if others agree.
  * Removed two scripts that do not work.
    * The 'sample dataset' script throws an error for me no matter what on both datasets because the datasets lack several JSON fields that the script assumes are present. The script makes strong assumptions that every JPEG file location is documented in the metadata, which is not true in our sample datasets. For now, I think it makes sense to delete the sample datasets (to avoid scope creep), but if we want to keep these around I can create a new docs ticket to update the sample datasets and provide a (hopefully just CLI command) code snippet to import them.
    * The 'Data Client API' script also doesn't work. Besides, this task should now be entirely achievable with the `dataset` CLI command and some filters. So if we think that 'add all images captured from a specific machine' is a popular enough task that we should document it, I can add a code snippet containing the one-line command.
* Updated the verification system tutorial to use the *Add image to dataset* button instead of enabling data capture (and never recommending that users disable it or even filter it as part of the main tutorial flow). Trimmed some steps in the process, and consolidated the prerequisites for this tutorial with the prerequisites for the *create a dataset* page since they are effectively the same. (ought I dual-source this content from some reusable source?)


docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
+This technique can help you create a dataset without filling the **DATA** tab with a large number of unhelpful empty images.
+
+To add an image directly from a dataset from a visual feed, complete the following steps:
+
+1. Open the **TEST** panel of any camera or vision component to view a feed of images from the camera.
+1. Click the button marked with the camera icon to save the currently displayed image to a dataset:
+   {{< imgproc src="/components/camera/add_image_to_dataset_button.png" alt="A button marked with the outline of a camera, emphasized in red" resize="800x" style="width:500px" class="imgzoom" >}}
+1. Select an existing dataset.
+1. Click **Add** to add the image to the selected dataset.
+1. When you see a success toast that reads "Saved image to dataset", you have successfully added the image to the dataset.
```
npentrel, line 54: I don't know if people generally understand what a toast is...
```suggestion
1. When you see a success notification that reads "Saved image to dataset", you have successfully added the image to the dataset.
```

npentrel: maybe popop? Or maybe just message?

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
+
+To view images added to your dataset, go to the **DATA** page's [**DATASETS** tab](https://app.viam.com/data/datasets) in the Viam app. Select the dataset to view the images within that dataset.
 
-{{< expand "At least 10 captured images of what you want your machine to recognize." >}}
+{{< alert title="Tip" color="tip" >}}
```
npentrel, line 58: This tip is confusing. At this point we haven't done any training or annotation yet. I think it's a good tip but I think the place or framing needs to provide more context.

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
-This may include capturing images of different angles, different configurations of objects and different lighting conditions.
-The more varied the provided dataset, the more accurate the resulting model becomes.
+{{< /alert >}}
 
-{{< /expand >}}
+{{% /tab %}}
+{{% tab name="Capture bulk images" %}}
+
+To capture a large number of images for training your ML models, [Capture and sync image data](/data-ai/capture-data/capture-sync/) using the data management service with your camera.
+
```
npentrel, line 69: You'd still need to add them to the dataset I think?

npentrel: Ah I see. You add them as the last step. That possibly needs some caveat then because if you followed one method they'll already be in the dataset

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
-{{< expand "Want to add images to a dataset programmatically? Click here." >}}
+Repeat these steps for all images in the dataset.
 
-You can also add all images with a certain label to a dataset using the [`viam dataset data add` command](/dev/tools/cli/#dataset) or the [Data Client API](/dev/reference/apis/data-client/#addtagstobinarydatabyfilter):
+## Add tagged images to a dataset
 
 {{< tabs >}}
+{{% tab name="Viam app" %}}
+
+1. Open the **DATA** page of the Viam app.
```
npentrel, line 148: Consider adding a link to data page for ease of access

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
+   Pass this dataset ID in the next step to add images to this dataset.
+
+1. Run the following [command](/dev/tools/cli/#dataset) to add all images labeled with a subset of tags to the dataset, replacing the `<dataset-id>` placeholder with the dataset ID output by the command in the previous step:
+
+   ```sh {class="command-line" data-prompt="$"}
+   viam dataset data add filter --dataset-id=<dataset-id> --tags=red_star,blue_square
+   ```
 
 {{% /tab %}}
 {{< tab name="Data Client API" >}}
```
npentrel, line 177: This tab is currently broken. Try to change the `<` and `>` to `%` and see if that fixes it.

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,244 +14,169 @@ aliases:
...
 {{% /expand%}}
 
-## Prerequisites
+## Capture images
+
+{{< tabs >}}
+{{% tab name="Capture individual images directly into a dataset" %}}
+
+You can add images to a dataset directly from a camera or vision component feed in the **CONTROL** or **CONFIGURATION** tabs of the Viam app.
+This technique can help you create a dataset without filling the **DATA** tab with a large number of unhelpful empty images.
```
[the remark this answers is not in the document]
nathan-contino:
```suggestion
This technique can help you create a dataset with a small number of intentionally captured images, instead of syncing a large number of empty images.
```
Great point, I definitely shouldn't have veered into qualitative wording like 'unhelpful'. Here's another stab at the same intention -- do you think this is useful, or NWW (like Prue says in GBBO, 'not worth the ~calories~ words')?

docs/data-ai/ai/create-dataset.md
nathan-contino: I like notification -- it looks quite similar to notifications on the desktop. Message is probably the most accurate and futureproof, but lacks some of the descriptive punch of *notification*.

docs/data-ai/ai/create-dataset.md
nathan-contino: Noted! I'll reword this for better parity between the tabs.

docs/data-ai/ai/create-dataset.md
nathan-contino: Fixed!

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,320 +14,310 @@ aliases:
...
 
-## Prerequisites
+{{% /tab %}}
+{{% tab name="Many images" %}}
 
-{{< expand "At least 10 captured images of what you want your machine to recognize." >}}
+To capture a large number of images for training an ML model, [Capture and sync image data](/data-ai/capture-data/capture-sync/) using the data management service with your camera.
 
-[Capture and sync image data](/data-ai/capture-data/capture-sync/) using the data management service.
+Viam stores the images saved by capture and sync on the [**DATA** page](https://app.viam.com/data/), but does not add the images to a dataset. Visit the **DATA** page to add images to a dataset.
```
npentrel, line 97: Maybe this since it's faster that way? 
```suggestion
Viam stores the images saved by capture and sync on the [**DATA** page](https://app.viam.com/data/), but does not add the images to a dataset.
We recommend you tag the images first and then use the CLI to [add the tagged images to a dataset](/data-ai/ai/create-dataset/#add-tagged-images-to-a-dataset).
```

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,320 +14,310 @@ aliases:
...
+Once you've captured enough images for training, you must annotate them to train a model.
 
-## Tips on improving model accuracy
+## Annotate images
 
-- **More data means better models:** Incorporate as much data as you practically can to improve your model’s overall performance.
-- **Include counterexamples:** Include images with and without the object you’re looking to classify.
-  This helps the model distinguish the target object from the background and reduces the chances of false positives by teaching it what the object is not.
-- **Avoid class imbalance:** Don’t train excessively on one specific type or class, make sure each category has a roughly equal number of images.
+Use the interface on the [**DATA** tab](https://app.viam.com/data/view) to label your images. Always follow best practices when you label your images:
```
npentrel, line 111: We may need to here briefly provide an explanation as to what the different annotations do. Otherwise people will read from top to bottom and tag for classification before they realize there's a second option.

docs/data-ai/ai/create-dataset.md
```diff
@@ -14,320 +14,310 @@ aliases:
...
+
+       # Replace "<PART-ID>" (including brackets) with your machine's part id
+       my_filter = create_filter(part_id="<PART-ID>")
+
+       print("Getting data for part...")
+       binary_metadata, _, _ = await data_client.binary_data_by_filter(
+           my_filter,
+           include_binary_data=False
+       )
+       my_binary_ids = []
```
npentrel, line 268: have you run this script? If you do have it currently set up it would be a good idea to remove the BinaryId usage here (it's deprecated for file id). Otherwise we'll want to file a follow up ticket

docs/data-ai/ai/run-inference.md
```diff
@@ -43,7 +43,7 @@ Then, from the **Select model** dropdown, select the name of the ML model servic
 
 ### Test your changes
 
-You can test a deployed vision service by clicking on the **Test** area of its configuration panel or from the [**CONTROL** tab](/manage/troubleshoot/teleoperate/default-interface/#viam-app).
+You can test a deployed vision service by clicking on the **Test** area of its configuration panel or from the [**CONTROL** page](/manage/troubleshoot/teleoperate/default-interface/#viam-app).
```
npentrel, line 46: This is probably worth discussing as a team. We historically refer to it as a tab. It used to look more like a tab. You can leave the change but worth bringing up to the next team sync

docs/data-ai/ai/train-tflite.md
```diff
@@ -26,50 +26,39 @@ date: "2024-12-03"
...
-
-Click **Train model** and follow the prompts.
+1. In the Viam app, navigate to your list of [**DATASETS**](https://app.viam.com/data/datasets) and select the one you want to train on.
 
-You can train a TFLite model using **Built-in training**.
+1. Click **Train model** and follow the prompts.
+   You can train a TFLite model using **Built-in training**.
 
-Click **Next steps**.
+1. Click **Next steps**.
```
npentrel, line 59: Consider splitting this up in 2-3 steps instead to avoid the subnesting

docs/data-ai/ai/train-tflite.md
```diff
@@ -141,6 +130,23 @@ You can test both detection models and classifier models using the following res
 - [a `tflite_cpu` ML model](/data-ai/ai/deploy/) with the model you just trained
 - [an `mlmodel` vision service](/operate/reference/services/vision/mlmodel/) using the `tflite_cpu` model
 
+## Iterate your ML model
```
npentrel, line 133:
```suggestion
## Iterate on your ML model
```

docs/data-ai/ai/train-tflite.md
```diff
@@ -141,6 +130,23 @@ You can test both detection models and classifier models using the following res
...
+
+With the right training dataset, your ML model can perform well as soon as training completes.
+But edge cases always exist.
+You are unlikely to account for all false positives or false negatives during your first round of training.
+Instead, keep track of edge cases as you run your model.
+Add them to your training dataset, annotate them, and re-train your model using the new data.
```
npentrel, line 139:
```suggestion
Add images capturing edge cases to your training dataset, annotate them, and re-train your model using the new data.
```

docs/data-ai/ai/train-tflite.md
```diff
@@ -141,6 +130,23 @@ You can test both detection models and classifier models using the following res
...
+Using this approach, each subsequent model version should be more accurate than the last.
```
npentrel, line 140: avoid conditionals
```suggestion
Using this approach, each subsequent model version becomes more accurate than the last.
```

docs/data-ai/ai/create-dataset.md
nathan-contino: I have run the script to prove that it works. But your question inspired me to dig a little bit deeper. It looks like `file_id` is the one that has been deprecated, but ironically the BinaryID Python type still includes a `file_id` field 🤷 . So we're stuck with it for now.

On the subject of the script itself: deeper testing revealed that the current script maxes out at 50 images (the default `limit` value). That's fine, if frustrating to users who want to exceed that limit, but don't know about the `limit` field in `binary_data_by_filter()`. But when I tried to set that limit higher, I realized that:

* if you mess up the part ID, the current script still creates an (empty) dataset
* if you configure the limit to something higher, the script hangs, potentially for a long time

So I updated the script to:

* query for data _before_ it creates the dataset
* use a configurable maximum number of results
* use the per-query `limit` field and `last` field to chain smaller requests and provide progress updates to avoid script 'hangs'
* use docstrings, formatted printing, and upfront configuration constants so users don't have to scroll all over the script to fill in their values (and elevated DATASET_NAME into the config constants so users won't miss it as easily as they used to)

tl;dr no need for a followup, unless it's filing a bug with the Python SDK for continuing to use `file_id` 😜

docs/data-ai/ai/run-inference.md
nathan-contino: Added to our weekly sync doc as a discussion item. I started to do this because some parts of the docs already use 'page' and I found 'DATA tab's DATASET subtab' kind of awful. But I suspect y'all have put more thought into this than me and I should leverage that.

docs/data-ai/ai/create-dataset.md
nathan-contino: Great point. I just switched the methods to tabs to avoid this potential area of confusion. I waffled on tabs vs. sections a few times as I reworked the page, but if someone else shares this same concern, it's probably worth accounting for!

docs/data-ai/ai/train-tflite.md
```diff
@@ -26,53 +26,45 @@ date: "2024-12-03"
...
 
-## Train a machine learning (ML) model
+## Train a machine learning model
 
-Now that you have a dataset with your labeled images, you are ready to train a machine learning model.
+Now that you have a dataset that contains your labeled images, you are ready to train a machine learning model.
 
 {{< table >}}
 {{% tablestep number=1 %}}
 **Train an ML model**
```
npentrel, line 53:
```suggestion
```

npentrel (approved):

docs/data-ai/ai/train-tflite.md
nathan-contino: D'oh. All of these tablesteps should have titles! I added one for the first step, so I think I should leave this one.

docs/data-ai/ai/create-dataset.md
nathan-contino: Updated to remove all mention of BinaryID. Unfortunately, the add_binary_data_to_dataset_by_ids method call still uses the `binary_ids` field, and there's no way around it. But at least we aren't using anything deprecated any more.

docs/data-ai/ai/run-inference.md
JessamyT: Agreed subtab used to be rough, but that was the next level down when we had configure tab, components subtab. I think **DATA** page, **CONTROL** tab and **DATASET** tab works. Data is at the level of fleet, so it can be a page I think haha.

JessamyT: (Also I am only looking at this because I was asked to; not just randomly double-reviewing!)

PR #4245, opened 2025-04-22
DOCS-3905: Add Windows viam-agent management instructions

JessamyT:


npentrel (approved):
Thank you, very nice

docs/manage/reference/viam-agent/manage-viam-agent.md
```diff
@@ -50,6 +53,60 @@ aliases:
...
   This command uninstalls `viam-agent`, `viam-server`, the machine cloud credentials file (<file>/etc/viam.json</file>), and the provisioning configuration file (<file>/etc/viam-provisioning.json</file>).
 
+{{% /tab %}}
+{{% tab name="Windows native" %}}
+
+On Windows, you can manage `viam-agent` using the Services GUI or the command line.
+
+## Manage `viam-agent` using the Services GUI
+
+1. Open the **Services** management console from your computer's start menu.
```
nathan-contino, line 63: [opt] reverse the order of this phrasing so readers don't have to double back:

```suggestion
1. In your computer's start menu, open the **Services** management console.
```

Disclaimer: I haven't used Windows since 10, so maybe the original order was more natural?

docs/manage/reference/viam-agent/manage-viam-agent.md
```diff
@@ -50,6 +53,60 @@ aliases:
...
+   {{<imgproc src="/manage/viam-agent-windows-services-manager.png" resize="x1100" declaredimensions=true alt="Windows Services manager with viam-agent highlighted." style="max-width:600px" class="shadow imgzoom" >}}
+
+1. Use the **Start Service**, **Stop Service**, and **Restart Service** buttons to manage `viam-agent`.
+
+1. To change the startup type of `viam-agent`, right-click on `viam-agent` and select **Properties**.
+   Select your desired startup type from the **Startup type** dropdown menu.
+
+   {{<imgproc src="/manage/startup-type-windows.png" resize="x1000" declaredimensions=true alt="Windows Services manager with viam-agent properties open." style="max-width:350px" class="shadow imgzoom" >}}
+
+## Manage `viam-agent` using the command line
```
nathan-contino, line 76: [q] why use sections within the Windows Native tab for CLI/GUI when those operations are (generally speaking) mutually exclusive? I would expect these parallel operations in tabs, but maybe we don't like stacking tabs that deep?

nathan-contino (approved):
LGTM % a question and a very optional suggestion

docs/manage/reference/viam-agent/manage-viam-agent.md
JessamyT: In general I'd go with your wording because if someone needs to go to a page before finding a button, it's better to say it in that order instead of "find the x button on the page you didn't know until now you should be on". But in this case I prefer it this way because they don't need to look for something that says "start"--they just need to find "Services" in the natural way, so the emphasis can be there.

docs/manage/reference/viam-agent/manage-viam-agent.md
JessamyT: It's a good question, and part of a larger convo about usage of tabs and expanders. I like both tabs and expanders but some people prefer seeing all the things in plain sight. Also our tabs aren't searchable with command f. But in this case it's all already in a tab and yes they are mutually exclusive so I will change to tabs!

PR #4250, opened 2025-04-24
DOCS-3814: Expand left sidenav one level down by default to improve navigation and discoverability

nathan-contino:
Given our tabbed configuration, the 'first level' is actually the third in configuration.
Much AI ink was spilled to discover this one-line change.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -117,7 +117,7 @@ navbar_logo = true
 navbar_translucent_over_cover_disable = false
 sidebar_menu_compact = true  # When true, the section headings work like an accordian.
 sidebar_search_disable = true
-ul_show = 3  # Show only the first level of the sidenav by default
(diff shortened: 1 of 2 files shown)


JessamyT:
Hmm this does create odd behavior with the expandy arrows though--can you make their default state pointing down instead of to the side, so that you don't click once to change the arrow, and click again to collapse?

OR are we okay with no collapsability, i.e. just make them headings (as they were a couple months ago) with no clickability?

nathan-contino:
I think it should be good now! Thanks for noticing this. I got a bit of tunnel vision looking at AI changes for too long.

JessamyT (approved):
LGTM! I still wonder if any users will actually want to collapse the left nav or if we should just make all top-level items headings without collapsibility (like they were a couple months ago), but that can be decided separately

PR #4251, opened 2025-04-25
DOCS-3866: Update Python SDK doc to reflect BinaryID deprecation

nathan-contino:
* BinaryID was recently deprecated. This updates the generated Python SDK docs to reflect that deprecation, among other changes, such as those in https://github.com/viamrobotics/viam-python-sdk/pull/913 (mostly cleanup, typo fixes, etc) and others since the last generated doc deploy.
* I'm not sure why running the SDK docs scraping script has trimmed so many `\` characters. As far as I can tell, they're unnecessary escape characters that we don't need, so they don't matter. Maybe the script changed since the last generated doc deploy?
* In working on this ticket, I identified that the scraping script doesn't handle scraping a few bits of markup correctly from docstrings. We lose bold, monospace, and italic text entirely. "Returns" and "Parameters" sections don't support bulleted lists. Some links lose their pretty text. I'm going to chalk that up as 'scope creep' and look the other way for now, but I hope to make some infra improvements to `.github/workflows/update_sdk_methods.py` eventually.
* This error should be fixed, but I'm not sure how we've handled it in the past:

```
ERROR Link text is too non-descriptive: "https://pointclouds.org/documentation/tutorials/pcd_file_format.html"
```

Basically, a dev put a raw URL in a docstring, and Hugo complained because it wasn't decorated with text. I've PR'ed a fix in the linked PR, but I'm not sure when that will make it in, so keep an eye out for this in the future. I suspect we could also update the scraping script to better handle this case, and automagically use the URL itself as text to stop Hugo complaints over something we can't directly control.


JessamyT (approved):
LGTM % from-upstream-typo that I commented on in upstream PR ("forthe"). But I want to hear Naomi's responses to your comments when she comes back.

npentrel (approved):

npentrel:
- My best guess is the slash possibly might be getting removed or added if the content is saved in an editor that automatically fixes formatting. I noticed it before but unless it's causing formatting issues it's not worth investigating.
- For the URL you did the exact right thing. Manual overwrite until we fix it in the sdk docs. We don't want to automatically overwrite it. That would add unnecessary extra code for us when it's easier and better to fix it in the sdk docs.

nathan-contino:
Updating the source code that ultimately provides the problematic pointcloud bare URL here: https://github.com/viamrobotics/api/pull/687

PR #4257, opened 2025-04-28
Add create MR alias

JessamyT:


nathan-contino (approved):
Not sure I fully understand why, but LGTM!

JessamyT:
> Not sure I fully understand why, but LGTM!

When we move pages around, we need to have aliases so links to the docs from sources beyond our control redirect to the correct page instead of 404ing. This alias was missing, and I only noticed because when triaging AI questions, I saw a link to a discord thread from a couple years ago which in turn contained this link (which 404ed), so I'm adding it now in case someone else hits it from that old thread or elsewhere on the interwebs!

PR #4264, opened 2025-04-30
DOCS-3940: Add info about how to get fqdn

npentrel:


docs/manage/software/update-software.md
```diff
@@ -165,10 +165,13 @@ async def main():
             if p.main_part:
                 main_part = p
 
-        print("Attempting to connect to {}...".format(main_part.fqdn))
+        // Get machine address
```
npentrel, line 168:
```suggestion
        # Get machine address
```

static/include/app/apis/generated/app.md
```diff
@@ -1643,6 +1645,12 @@ For more information, see the [Python SDK Docs](https://python.viam.dev/autoapi/
 
 ```ts {class="line-numbers linkable-line-numbers"}
 const robotPart = await appClient.getRobotPart('<YOUR-ROBOT-PART-ID>');
+// Get the part's address
+const address = robotPart.part.fqdn;
+// Check if machine is live (last access time less than 10 sec ago)
+if (Date.now() - Number(robotPart.part.lastAccess.seconds) * 1000 <= 10000) {
```
nathan-contino, line 1651: Frustrating that they only provide seconds and a separate count of nanoseconds on top of that. At least the magic numbers are justified here!

nathan-contino (approved):
LGTM!

PR #4268, opened 2025-04-30
Ask AI feedback: Fix package version configuration example

nathan-contino:
* fixes an issue seen in the Ask AI slack channel.
  Question:
  How would a fragment override for an ml_model package version look like?

  Answer:
  Here's how you can override an ML model package version in a fragment:

  ```
  "fragment_mods": [
  {
    "fragment_id": "abcd7ef8-fa88-1234-b9a1-123z987e55aa",
    "mods": [
      {
        "$set": {
          "packages.0.version": "latest"
        }
      }
    ]
  }
  ]
  ```

  You can set the version to:

  - "^<major version number>" (e.g. "^1") for updates within the same major release
  - "~<minor version number>" (e.g. "~1.8") for updates within the same minor release
  - "latest" to always update to the latest release
  - "<version number>" (e.g. "1.8.3") to pin to a specific version

  Remember to check your machine's LOGS tab after configuring fragment overwrites to ensure there are no issues with the configuration.

  Update:
  It should be `packages.<package_name>.version`
* I noticed that the preceding section contained very similar content, so I adjusted the language in both sections for consistency and added the list of possible version values/explanations to the package section, since the other section may be hidden for users linked to the page.

I couldn't get the `0` syntax to work, but if someone else knows how to get it to work, happy to add an example for that if we want to document it. I'm guessing that syntax was removed at some point?

### DIFF
--- a/docs/manage/fleet/reuse-configuration.md
+++ b/docs/manage/fleet/reuse-configuration.md
@@ -314,8 +314,8 @@ This example uses [`$set`](https://www.mongodb.com/docs/manual/reference/operato
 ```
 
 {{< /expand >}}
-{{< expand "Pin a module version" >}}
-This example uses [`$set`](https://www.mongodb.com/docs/manual/reference/operator/update/set/#mongodb-update-up.-set) to set [version update settings for a module](/operate/reference/module-configuration/#module-configuration-details) named `custom-sensor` in the fragment:
+{{< expand "Set a module version" >}}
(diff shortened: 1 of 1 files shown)


npentrel (approved):
I think the .0 would've accessed the first array item, it may have indeed changed. I don't think we want an example, it's not a good way to do it, so you can merge as is

PR #4273, opened 2025-05-01
DOCS-3947: Make linkTitle casing consistent

JessamyT:


docs/operate/reference/architecture/parts.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Machine Architecture: Parts"
-linkTitle: "Parts, Sub-parts, Remotes"
+title: "Machine architecture: parts"
```
JessamyT, line 2: This is the only one I'm unsure of: Do we capitalize after the colon?

nathan-contino: Google's style guide suggests we should?

https://developers.google.com/style/capitalization

But I'm skeptical that this colon belongs here at all. And simultaneously unsure what I'd do instead. So maybe we should capitalize it?

nathan-contino (approved):
lgtm!

PR #4284, opened 2025-05-12
DOCS-3961: Change viam-agent update behaviour

npentrel:


docs/operate/reference/viam-server/_index.md
```diff
@@ -72,6 +72,7 @@ Alternatively, if you are having issues with a module, try the **Restart module*
 
 There are a few updates that may make your machine temporarily unavailable:
 
+- [`viam-agent` updating itself](/manage/reference/viam-agent/#version_control-version-management-for-viam-agent-and-viam-server)
```
npentrel, line 75: The safe tor restart getting reported is the maintenance window, right?

Otterverse: Yes,it basically verifies with the same API that viam-server itself used to report "safe to restart" so should obey all maintenance windows/components just as it does for viam-server-only restarts.

Otterverse (approved):
LGTM

PR #4302, opened 2025-05-15
DOCS-3665: Add billing information for adding metadata

npentrel:
We don't usually put billing info in but this is an odd one that may be surprising to people.


nathan-contino (approved):
LGTM

PR #4319, opened 2025-05-19
DOCS-3925: Update control-logic

npentrel:


docs/manage/software/control-logic.md
```diff
@@ -15,7 +15,7 @@ date: "2025-02-14"
 cost: "0"
 ---
 
-One method of operating your machine is by running control logic on another device, such as an app.
+One method of operating your machine is by [running control logic on another device, such as an app](/operate/control/headless-app/).
 However, you can also run control logic directly on the machine itself using {{< glossary_tooltip term_id="module" text="modules" >}}.
 These are the same modules that provide functionality like drivers and integrations to your machines.
```
JessamyT, line 20: I think more explanation is needed, and the current text makes it sound like you can't run client code on the machine itself. How about some version of
```suggestion
One method of operating your machine is by [running client code, such as an app](/operate/control/headless-app/).
You can run a client directly on the machine itself or on another device.
However, to deploy code directly onto multiple machines while maintaining version control, you can package the control logic as a {{< glossary_tooltip term_id="module" text="module" >}}.
These are the same modules that provide functionality like drivers and integrations to your machines.
```

JessamyT (changes requested):

docs/manage/software/control-logic.md
npentrel:
```suggestion
One method of operating your machine is by [running client code, such as an app](/operate/control/headless-app/), on the machine itself or on another device.
However, to deploy versioned code directly onto multiple machines, you can package the control logic as a {{< glossary_tooltip term_id="module" text="module" >}}.
These are the same modules that provide functionality like drivers and integrations to your machines.
```

JessamyT (approved):
LGTM!

PR #4329, opened 2025-05-20
DOCS-3942: Add env var usage

JessamyT:
Another open PR will adjust the H4 above this to an H3, so H4 is the correct heading level to use for the new sections here.


docs/operate/reference/module-configuration.md
```diff
@@ -290,4 +283,13 @@ To set the path to a program or library on a machine, you can set a system varia
 }
 ```
 
-{{% /expand%}}
+#### Access environment variables from your module
+
+You can access the variables from within your module code in the same way you would access any other environment variables.
```
npentrel, line 288: (unclear reference)
```suggestion
You can access environment variables from within your module code in the same way you would access any other environment variables.
```

docs/operate/reference/module-configuration.md
```diff
@@ -290,7 +283,16 @@ To set the path to a program or library on a machine, you can set a system varia
 }
 ```
 
-{{% /expand%}}
+#### Access environment variables from your module
+
+You can access environment variables from within your module code in the same way you would access any other environment variables.
```
JessamyT, line 288:
```suggestion
You can access module environment variables from within your module code in the same way you would access any other environment variables.
```

PR #4330, opened 2025-05-20
DOCS-3962: Tweak local module instructions

JessamyT:
Apply some fixes that arose from a recent convo w/Felix

### DIFF
--- a/docs/operate/get-started/other-hardware/_index.md
+++ b/docs/operate/get-started/other-hardware/_index.md
@@ -778,6 +778,7 @@ On your machine's **CONFIGURE** tab in the [Viam app](https://app.viam.com), cli
 Select **Local module**, then **Local module**.
 
 Enter the absolute path to the <file>run.sh</file> script, for example `/home/jessamy/my-module/run.sh` on Linux, or `/Users/jessamy/my-python-sensor-module/run.sh` on macOS.
+For modules configured this way, `viam-server` uses this path instead of the `entrypoint` field in your <file>meta.json</file> file.
 
 Click **Create**.
 
@@ -788,6 +789,7 @@ On your machine's **CONFIGURE** tab in the [Viam app](https://app.viam.com), cli
 Select **Local module**, then **Local module**.
 
 Enter the absolute path to the <file>/bin/&#60;module-name&#62;</file> executable, for example `/home/jessamy/my-go-module/bin/mymodule` on Linux, or `/Users/jessamy/my-go-module/bin/mymodule` on macOS.
+For modules configured this way, `viam-server` uses this path instead of the `entrypoint` field in your <file>meta.json</file> file.
 
 Click **Create**.
 
@@ -828,7 +830,7 @@ For example, if you created a sensor component, check whether readings are displ
 If your component works, you're almost ready to share your module by uploading it to the registry.
 If not, you have some debugging to do.
 
-Each time you make changes to your local module, you need to rebuild the module and then restart its instance on your machine.
+Each time you make changes to your local module code, you need to update its instance on your machine:
 
 {{< tabs >}}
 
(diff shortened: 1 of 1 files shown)


docs/operate/get-started/other-hardware/_index.md
```diff
@@ -788,6 +789,7 @@ On your machine's **CONFIGURE** tab in the [Viam app](https://app.viam.com), cli
 Select **Local module**, then **Local module**.
 
 Enter the absolute path to the <file>/bin/&#60;module-name&#62;</file> executable, for example `/home/jessamy/my-go-module/bin/mymodule` on Linux, or `/Users/jessamy/my-go-module/bin/mymodule` on macOS.
+For modules configured this way, `viam-server` uses this path instead of the `entrypoint` field in your <file>meta.json</file> file.
```
npentrel, line 792: This is a bit of a gotcha that may trip people up I expect - can you get this into the UI? Possibly when adding the local module OR on local module cards as a note of sorts.

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -843,13 +845,7 @@ viam module reload <insert relevant named args>
...
 {{% tab name="Python: venv" %}}
 
-Run the following command to rebuild your module:
-
-```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
-viam module build local
-```
-
-Then restart it in your machine's **CONFIGURE** tab in the Viam app.
+Restart the module in your machine's **CONFIGURE** tab in the Viam app.
```
npentrel, line 848: so for pyinstaller I need to run the CLI command but for venv I need to do it in the UI? What happens if I use the restart option in the UI for pyinstaller and the cli command for venv?

npentrel (changes requested):
First comment is not a blocker for this PR but one I think it's probably worth getting fixed in the product. Second comment is confusing to me

docs/operate/get-started/other-hardware/_index.md
JessamyT: So they're not actually the same--reload rebuilds AND restarts, whereas the UI just restarts. So, reload is unnecessarily time consuming if using run.sh as the entrypoint, because nothing needs to be built.

If you use the restart button for pyinstaller, you won't have a newly-built executable--you'll just be restarting the same executable you used previously.

docs/operate/get-started/other-hardware/_index.md
JessamyT: Okay! https://viam.atlassian.net/browse/APP-8332

docs/operate/get-started/other-hardware/_index.md
JessamyT: Do you think I should make this more explicit? Didn't want to over-explain and make it to verbose, but rather just tell them what they need to do.

npentrel: Yeah I think it may just be a bit of a tripwire where people see `viam module build local` on the first tab and expect to be able to do the same on venv. I think a sentence about "Because you're using pyinstaller it is enough to rebuild the module with: ..." and soemthing similar on venv would avoid any stumbling

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -843,13 +845,8 @@ viam module reload <insert relevant named args>
...
 
-Run the following command to rebuild your module:
-
-```sh {id="terminal-prompt" class="command-line" data-prompt="$"}
-viam module build local
-```
-
-Then restart it in your machine's **CONFIGURE** tab in the Viam app.
+Since you are using <file>run.sh</file> instead of a built executable, you do not need to rebuild anything as you iterate.
+Just save your code changes, then restart the module in your machine's **CONFIGURE** tab in the Viam app.
```
JessamyT, line 849:
```suggestion
Just save your code changes, then restart the module in your machine's **CONFIGURE** tab in the Viam app:
```

docs/operate/get-started/other-hardware/_index.md
JessamyT: Okay added something to the venv tab but I think the first tab is okay as-is now that the venv tab addresses why it's different.

PR #4341, opened 2025-05-27
Create Quickstart section and reduce restructuring to the minimum needed to fit new content

nathan-contino:


docs/data-ai/quickstart/_index.md
```diff
@@ -0,0 +1,17 @@
+---
+linkTitle: "Quickstart"
+title: "Quickstart"
+weight: 1
+layout: "docs"
+type: "docs"
+open_on_desktop: true
+header_only: true
+---
```
npentrel, line 9: Please make this an empty parent, unless the plage itself communcates something of value that isn't just "there's three steps in these subpages"

docs/data-ai/quickstart/capture.md
```diff
@@ -0,0 +1,9 @@
+---
+linkTitle: "Capture image data"
+title: "Quickstart: Capture image data"
+weight: 20
+layout: "docs"
+type: "docs"
+description: "Capture images on your machine to use for inference."
```
npentrel, line 7: Is that needed for a quickstart even? Would it not be better to use inference on live images? Then you cut out one additional step? You don't show data capture then - is that vital?

docs/data-ai/quickstart/set-up-viam.md
```diff
@@ -0,0 +1,11 @@
+---
```
npentrel, line 1: To also save space in the left nav and limit duplication, I'd move this to a "requirements" section and then link to our existing pages that cover setup.

docs/data-ai/reference/api/data-client.md
```diff
@@ -4,5 +4,7 @@ linkTitle: "Data client API"
 weight: 30
 type: "docs"
 layout: "empty"
+aliases:
+  - /data-ai/reference/data-client/
```
npentrel, line 8: Because this is not a real page but only something used for forwarding you can skip these aliases. It's very unlikely we'll get 404s from those.

docs/data-ai/capture-data/conditional-sync.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Conditional cloud sync"
```
npentrel, line 2: nice

docs/data-ai/capture-data/filter-before-sync.md
```diff
@@ -1,5 +1,5 @@
 ---
-linkTitle: "Filter data"
+linkTitle: "Filter captured data"
```
npentrel, line 2: Discuss with Jack probably but I wonder if Selective sync would get better across that the filtering is happening before sync. Maybe Filter before sync is also an option

docs/data-ai/capture-data/_index.md
```diff
@@ -2,10 +2,117 @@
 linkTitle: "Capture data"
 title: "Capture data"
 weight: 100
-layout: "empty"
+layout: "docs"
 type: "docs"
-empty_node: true
```
npentrel, line 7: I think we do actually want to keep these top level pages empty. There's been a lot of confusion on whether people get confused with some of these having content and others not having content. I think in this case (1) we have a child page with almost the same name so the content should be merged onto that page or this page's content needs to go into reference. (2) with the current design the top level pages really don't feel like they should have content themselves. No other top level page currently does. (3) this isn't the first thing you'd want to tell someone. This is the info for people who want to know more so it should come after they get the basics.

docs/data-ai/inference/augment-visual-inference-with-sensor-data.md
```diff
@@ -0,0 +1,10 @@
+---
```
npentrel, line 1: The Run inference section is now very long compared to other sections. Consider grouping them (Maybe Detecting & Tracking, Augmenting, Autonomous decisions?)

npentrel (commented):
Overall love this direction, a few small pieces of feedback. I recognize you may not use this PR for actually implementing it but added here for context anyway

PR #4356, opened 2025-05-28
DOCS-3906, DOCS-4008: Change module IA, surface popular reference

JessamyT:
Still have various links to update but figured makes sense to get eyes on it before investing in link fixes


docs/operate/get-started/other-hardware/lifecycle-module.md
```diff
@@ -0,0 +1,44 @@
...
+1. If at any point the user changes the configuration of the machine, `viam-server` reconfigures the affected resources within 15 seconds.
+
+1. If `viam-server` attempts to shut down an individual module (for example due to a user disabling a module) and the module does not shut down within 30 seconds, `viam-server` kills the module.
+
+1. When `viam-server` shuts down, it first attempts to shut down each module sequentially in no particular order.
+   If a given module does not shut down within 30 seconds, it is killed with a `SIGKILL`.
+   If any modules are still running after 90 seconds, `viam-server` kills them as well.
+   This means that if four modules are running and the first three each fail to shut down within 30 seconds each, the fourth is killed immediately at the 90 second mark.
+
+For microcontrollers, you must flash a [firmware build that includes the Micro-RDK](/operate/get-started/other-hardware/micro-module/) and one or more modules onto your device.
```
npentrel, line 44: this is peculiar on this page. Nothing else here even mentions the Micro-RDK

docs/operate/get-started/other-hardware/create-module/_index.md
```diff
@@ -0,0 +1,1102 @@
+---
+title: "Create a module"
```
npentrel, line 2: for esp32 this is called Modules for ESP32, so should this not just be modules? And then possibly have hello world, dependencies, meta.json, update and manage and lifecycle under it?

npentrel, line 2: also previous page should have a next button linking to this, meta.json should have next link to update and manage and please check for any other dangling ends. The idea is that this is a path through the docs you can click.

npentrel (changes requested):

docs/operate/get-started/other-hardware/lifecycle-module.md
JessamyT: This is preexisting content; Esha and I previously figured it's important to point out that micro modules are different from the modules discussed on this page

docs/operate/get-started/other-hardware/create-module/_index.md
JessamyT: So, two things:
1. If it's just called "modules," that doesn't tell people whether it's reference content or how-to. I'd say "Integrate other modules" is already approximately equivalent to "Modules," just in imperative per agreed upon top-level nav standards.
    - If it continues to be called "Create modules" (my preference), then Update and manage doesn't belong inside it since that's post-creation
3. The Modules for ESP32 page covers more than just creation

JessamyT: Agree with button to this create page, but for meta.json I should actually remove previous and not have any forward/back buttons right, since it's just a reference page and not part of flow per se

docs/operate/get-started/other-hardware/lifecycle-module.md
npentrel: it's a step in the right direction but explaining how to flash it on the device doesn't seem equivalent to lifecycle information. Are they started in a specific order? Are there dependencies? probably not worth investigating 

```
Microcontroller modules fuction differently and are embedded in the firmware for your device.
For more information see [Modules for ESP32](/operate/get-started/other-hardware/micro-module/).
```

docs/operate/get-started/other-hardware/create-module/_index.md
npentrel: Hmm yeah I think it could go in create modules OR maybe it could be a subpage of "set up an esp32"? It doesn't actually feel like it belongs with "regular" modules other than by virtue of also being called modules. They don't work the same. The module configuration, lifecycle, updating and managing modules, nothing here applies to the esp32 modules. And if you encounter the module for esp32 here you have already flashed your hardware because it's after "set up an esp32" so you'd have to go back and redo that. I mean you probably will have to anyway while developing but it still feels odd. So yeah either move under create or under setup I think

npentrel: I get that it doesn't feel like a next step but wouldn't be too strict about that. If we interleave the reference pages then they become part of the flow.
So I think meta.json should continue to update and manage modules and I think module configuration can continue to lifecycle of a module.

JessamyT: Hmmm was about to move micro modules under setup esp32 but it occurs to me: If someone sees "Integrate other modules" they might look there for esp32 modules, and not see anything, and take a while to find this page, because they won't look in "setup" if they've already set it up. So actually, the very fact that nothing else in the Integrate Other Modules section has nothing to do with ESP32 is a reason *to* put it there so people see it instead of looking at the other things.
Also though, maybe it doesn't matter that much since people will use search and ask AI, and it's better to err on the side of not moving the micro modules page at all so there's one less redirect.

npentrel: I'd suggest putting a callout on the create a module page? "If you're using ESP32, see ..."
I think the risk is that people will assume that all the other pages work with micro-RDK and actually none of them are relevant to Micro-RDK modules. Your call I'll approve it for now

npentrel (approved):

docs/operate/get-started/other-hardware/create-module/_index.md
JessamyT: Yeah I share your concern to an extent but I'm hoping that leaving it called "Modules for ESP32" instead of "Create modules for ESP32" will in a small way mitigate that--I'm hoping people will read it as "Everything to do with modules for ESP32." And I'm hoping that just seeing "ESP32" not buried another layer deeper will cause them to click it. Going to leave for now.

JessamyT: (Meaning I'll leave the page where it is, but I think a callout wouldn't hurt so I'll add that)

docs/operate/get-started/other-hardware/_index.md
```diff
@@ -1,1210 +1,8 @@
 ---
-title: "Integrate other physical or virtual hardware"
-linkTitle: "Integrate other hardware"
+title: "Integrate other resources with modules"
+linkTitle: "Integrate other modules"
```
nathan-contino, line 3:
```suggestion
title: "Develop modules"
linkTitle: "Develop modules"
```
While I understand the general thoughts behind 'integrate other modules' (we have talked about this at length, after all), I fear that 'integrate' is overused and rendered just as meaningless a term as 'leverage' and 'utilize' and their ilk. And I'm honestly not sure what 'other' accomplishes in this case.

So how about something completely different? While I would be fine with the simple title 'Modules', I'm inclined to agree that we should stick with verb phrases here. When I look at all of these new titles, especially the new 'lifecycle' one, it reminds me of SDLC... so how about 'develop' as a verb phrase? Should be very easy for users to find if they're looking for information about writing and publishing modules, I think.

JessamyT: The idea is that "Integrate other modules" is in contrast to "Configure supported hardware." Like, you check whether it's already supported by an existing module, and if not, you integrate your thing with a new module. Which indeed, you do "develop,"... but the point of development is to integrate, so "integrate" focuses on the ultimate goal more (theoretically).

nathan-contino:
Since naomi has already approved, no need for me to give a full review. But I just wanted to say that I really love this change, especially the decision to pull the lifecycle and meta.json reference content out into separate pages. I've previously found all of that content really hard to find and this would have helped me a lot when I first started using Viam. Bravo!

docs/operate/get-started/other-hardware/_index.md
JessamyT: Yeah I don't disagree that "integrate" is a bit more ambiguous, but this gets into the discoverability vs findability goals--gonna leave for now but heard!

PR #4361, opened 2025-05-29
DOCS-3953: Update frame tab references

JessamyT:
Only updating one screenshot since ur5e is one of the only components actually in use and worth updating screenshots for. A handful of other screenshots exist that don't show the "add frame" button but it should be self-explanatory anyway, and the text doesn't say anything misleading in those cases.


npentrel (approved):
LGTM % comment

PR #4370, opened 2025-05-31
DOCS-4042: Clarify Micro-RDK 32-bit support

JessamyT:
To get AI to stop saying "For 32-bit systems, you can use viam-micro-server which is specifically built for microcontrollers."


npentrel:
I think that's the wrong fix. We support 32-bit Linux systems in beta. We should add that info instead

PR #4386, opened 2025-06-06
Copy section URL on section hover chain icon click

nathan-contino:
I've heard from a few people that they expect the chain icon that appears when you hover over a section header (h1, h2, h3, etc) to _copy_ the anchor link when you click it. It currently navigates to that anchor point and updates your URL in the URL bar of your browser. This adds the copy functionality.


JessamyT (commented):
LGTM except it doesn't work on [Glossary](https://deploy-preview-4386--viam-docs.netlify.app/dev/reference/glossary/) or [Changelog](https://deploy-preview-4386--viam-docs.netlify.app/dev/reference/changelog/), who use different `layouts` --would be very nice to address here so that the behavior is consistent across all pages.

JessamyT (approved):
I've seen the summaries not loading in local build before (with prod being fine) so that's probably okay?
The links seem to work so LGTM

assets/js/base.js
```diff
@@ -0,0 +1,11 @@
+function handleAnchorClick(anchor) {
+   // write the full anchor URL to the clipboard
+   navigator.clipboard.writeText(location.href.split('#')[0] + '#' + anchor ); 
+
+   const checkmark = document.getElementById('checkmark-' + anchor);
+   checkmark.classList.add('visible');
+   
+   setTimeout(() => {
+     checkmark.classList.remove('visible');
```
npentrel, line 9: Can it say copied? A bit like the code samples do? That would be more using existing patterns

nathan-contino:
@npentrel switched to 'copied' text, slightly smaller than the title, styled similarly to our current code samples & italicized to obviously differentiate it from the section headers.

Currently displays for 2 seconds after the user clicks the permalink.

Taking advantage of built-in Docsy hooks to re-implement the (client-side!!) javascript that currently generates these self-links, instead of applying the onclick method to the entire heading. Let me know if you'd like the heading to be clickable as well -- I stuck with this because this approach effectively behaves the exact same way as the production site, just with additional copying to clipboard and the little copy notification.

Let me know if you'd like the styling to change.

npentrel (approved):
This is better and I'll approve so that I'm not blocking merge. However the styling is a bit off. It is centered compared to the anchor marker but it would make more sense to center it based on the header. It would also be good if the anchor showed for the same 2 seconds, it currently disappears and then copied is still visible which feels a bit odd to me

nathan-contino:
Good point, I'll tweak the alignment so it's centered based on the header. I went with this because I was getting some strange alignment relative to monospace headers previously, but if I can figure that out I think it would look better, too.

PR #4389, opened 2025-06-10
DOCS-4070: Hint about organization

npentrel:


JessamyT (approved):
I commented on the ticket about the intent, but this is fine to add here too

PR #4398, opened 2025-06-13
Update link to new arm motion page

HipsterBrown:
Since the docs for working with arms and motion planning have been updated, old links to the previous "move an arm" index page no longer resolve correctly. This is just the first instance I found while using the docs. Here's a rough search for the others: https://github.com/search?q=repo%3Aviamrobotics%2Fdocs%20move-arm&type=code


HipsterBrown:
Closing in favor of your fix, Jessamy.

PR #4405, opened 2025-06-18
Update lorawan diagram

nathan-contino:


JessamyT (approved):
Prettier than what's currently there so approving, BUT:
1. The white text on the purple pills is very hard to read. I think it should be thicker font, and the pills (and their labels) should be a lil bigger.
2. Recommend using SVG instead of PNG for these so you don't get the bad resolution

PR #4408, opened 2025-06-23
Fix scroll position when a user clicks a section in the on this page …

nathan-contino:
Anywhere on the current production site, clicking the 'on this page' widget entries loads a position almost, but not entirely unlike the anchor point you reach if you click on a section header. Bit awkward since this cuts off the top half of the header text.

This fix makes both events scroll you to the same point.


JessamyT (changes requested):
Hmm this seems to have broken the little vertical bar of On This Page that shows where you're at

nathan-contino:
@JessamyT the latest push should fix the vertical bar, nice catch!

JessamyT (approved):
Seems good now, though TBH since OG bug is only happening for me half the time I won't be surprised if we have to circle back to this

PR #4421, opened 2025-06-28
DOCS-4072: Redirect empty pages to top page of respective section

JessamyT:
So that if someone somehow lands on one of these, they at least stay in the same section

Oddly, with non-reference ones specifically, if `title` is in the frontmatter the canonical link doesn't work. But in reference it doesn't matter. Not going to dig at all since it seems irrelevant to this quick fix.


nathan-contino (approved):
LGTM

PR #4428, opened 2025-06-30
DOCS-4107: Document new positional args for viam data export binary

nathan-contino:
Command now requires users to specify `ids` or `filter`


JessamyT (approved):
LGTM assuming tested

PR #4436, opened 2025-07-02
Make 'connect devices' sidebar entry consistent with other Operate categories

nathan-contino:
Currently 'Connect devices' shows up at a different font size, which looks wonky; this resolves the issue.


npentrel (approved):
good catch

PR #4444, opened 2025-07-03
DOCS-4140: Clarify where logs are

npentrel:


npentrel:
@dgottlieb fairly sure I got this correct but would you mind reviewing? It's only a few lines

docs/manage/troubleshoot/troubleshoot.md
```diff
@@ -89,6 +92,9 @@ If not follow the steps for the standalone version.
 {{% /tab %}}
 {{% tab name="macOS" %}}
 
+`viam-server` writes logs to STDOUT and does not store them in a file on your machine.
```
JessamyT, line 95: The Linux instruction makes sense but for the mac one, maybe clarify where STDOUT goes by default so there's more of a distinction between default behavior versus what -debug does?

docs/manage/troubleshoot/troubleshoot.md
```diff
@@ -65,6 +65,9 @@ If not follow the steps for the standalone version.
 {{< tabs >}}
 {{% tab name="Linux" %}}
 
+By default, `viam-server` writes logs to STDOUT and does not store them in a file on your machine.
```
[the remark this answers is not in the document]
npentrel: thank you leaving as is. For anyone who is using a historic script, I assume they have reasons and thus likely also the knowledge on how this may be different for them in their case.

docs/manage/troubleshoot/troubleshoot.md
npentrel, line 95: stdout is a data stream. if a program is started in a terminal it's generally where stdout goes. If the program is started in the background or in a way that is survives the terminal window being killed stdout the output is just silently lost. So it doesn't go anywhere. At least I'm reasonably certain that is how it works. 
```suggestion
`viam-server` does not store logs in a file on your machine.
```
I can do this, any other change I can come up with makes this worse.

docs/manage/troubleshoot/troubleshoot.md
```diff
@@ -89,6 +92,9 @@ If not follow the steps for the standalone version.
 {{% /tab %}}
 {{% tab name="macOS" %}}
 
+`viam-server` writes logs to STDOUT and does not store them in a file on your machine.
+To view your logs, stop the running `viam-server` instance, and restart it which will sent the logs to STDOUT in the current terminal window.
```
npentrel, line 96:
```suggestion
To view logs, stop the running `viam-server` instance, and restart it which will sent the logs to STDOUT in the current terminal window.
```

PR #4451, opened 2025-07-07
DOCS-4148: Add optional deps reconfigure function

JessamyT:
Realized optional dependencies instructions were missing a different reconfig function for Python. But for Go, should not be necessary still since alwaysRebuild will cover it. Right @benjirewis ?


benjirewis (commented):
Looks pretty good; @aldenh-viam did the changes for Python (and is better at Python than me) so curious what he thinks.

benjirewis:
> But for Go, should not be necessary still since alwaysRebuild will cover it. Right @benjirewis ?

Using `AlwaysRebuild` is indeed fine in Go. You also do not _have_ to implement a `reconfigure` method in Python, either. I believe we'll just reconstruct your resource.

npentrel (approved):

PR #4452, opened 2025-07-07
Update annotate section based on Product feedback

nathan-contino:
- use fewer words
- tighten up language around annotate/tag/label
- answer 'why' quickly


JessamyT (approved):
LGTM

PR #4455, opened 2025-07-08
DOCS-4153: Fix gpio servo description

JessamyT:


docs/operate/reference/components/servo/gpio.md
```diff
@@ -9,7 +9,7 @@ icon: true
 images: ["/icons/components/servo.svg"]
 aliases:
   - "/components/servo/gpio/"
-component_description: "Supports hobby servos wired to any model of board besides pi."
+component_description: "Supports hobby servos wired to any model of board besides Raspberry Pi 0 through 4."
```
nathan-contino, line 12:
```suggestion
component_description: "Supports hobby servos wired to any model of board besides Raspberry Pi 0 through 4."
```
Raspberry Pi Zero (never written '0', to my knowledge) is a separate product from the credit card-sized SBCs, so '0 through 4' is ambiguous. It would be better to list the supported Raspberry Pi models (only 5, I think?).

nathan-contino (changes requested):

docs/operate/reference/components/servo/gpio.md
JessamyT: So your suggestion would be this?
```suggestion
component_description: "Supports hobby servos wired to any model of non-Raspberry Pi board as well as Raspberry Pi 5."
```

nathan-contino: Pi 5-first gets my vote -- my suspicion is that this doesn't actually support ANY non-pi board, since that is a very very broad category, so it's best to set expectations. Honestly we should probably be more specific, but I don't know anything about the requirements, so I can't suggest anything there. But if you know more about the hardware requirements maybe you can come up with something.

JessamyT: @randhid Does it support any non-pi board that Viam supports w/ a board component model?

JessamyT:
```suggestion
component_description: "Supports a hobby servo wired to a board that supports PWM, for example Raspberry Pi 5, Orange Pi, Jetson, or PCAXXXX."
```

PR #4469, opened 2025-07-14
DOCS-4177: Add typescript offline mode

npentrel:


docs/dev/reference/sdks/connectivity.md
```diff
@@ -57,7 +57,51 @@ machine, err := client.New(
...
+const machine = await VIAM.createRobotClient({
+  host,
+  credentials: {
+    type: "api-key",
+    /* Replace "<API-KEY>" (including brackets) with your machine's API key */ payload:
+      "<API-KEY>",
+    authEntity: "<API-KEY-ID>",
+    /* Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID */
+  },
+  signalingAddress: `http://${host}:8080`,
```
stuqdog, line 101: As I recall, the signaling address needs to be `http://${host}.local:8080` and the host address doesn't need to be changed. That might be out of date or misrecalled, however. To confirm, have you tried connecting in a no-internet mode with these instructions to confirm it works?

stuqdog: Okay the issues I am experiencing appear to be browser-dependent (offline mode seems to be having issues on chrome but works on safari). This seems environmental and I will dig into it; creating a ticket for it.

In the meantime, though, when I _was_ able to get offline mode working, it was by following the instructions that already existed in the TS README. So I would say we should stick with those instructions and _not_ point people to using a `.local.viam.cloud` address, as that will not work.

npentrel: Ok we'll have to pair on this probably because I did try this earlier and I could not get it to work with the .local suffix. This is also a problem if we have it wrongly documented for all the SDKs currently...
 https://deploy-preview-4469--viam-docs.netlify.app/dev/reference/sdks/connectivity/#connect-over-local-network-or-offline

stuqdog (approved):

PR #4473, opened 2025-07-14
DOCS-4087: autonomous decision-making

nathan-contino:
- Builds out three proper guides demonstrating in-module autonomous decision-making logic.
- Removes the minimal, not terribly useful 'industrial' examples from the previous page.
- Fixes some minor cosmetic issues in the Line Following Robot tutorial.
- To keep the Data & AI sidebar from scrolling, relocated all API links to an "APIs" section.

### DIFF
--- a/docs/data-ai/ai/make-decisions-autonomously.md
+++ b/docs/data-ai/ai/make-decisions-autonomously.md
@@ -586,10 +586,10 @@ class ObjectFollower(Module):
                     and dependency_name.name == self.camera_name):
                 self.camera = dependency
             elif (dependency_name.subtype == "vision"
-                and dependency_name.name == self.detector_name):
+                    and dependency_name.name == self.detector_name):
                 self.detector = dependency
             elif (dependency_name.subtype == "base"
-                and dependency_name.name == self.base_name):
+                    and dependency_name.name == self.base_name):
                 self.base = dependency
 
         if not self.camera:
@@ -654,8 +654,7 @@ class ObjectFollower(Module):
         """
         ObjectFollower.LOGGER.info("Object tracking control loop started.")
 
-        initial_frame =
-            await self.camera.get_image(mime_type="image/jpeg")
+        initial_frame = await self.camera.get_image(mime_type="image/jpeg")
         pil_initial_frame = viam_to_pil_image(initial_frame)
         midpoint = pil_initial_frame.size[0] / 2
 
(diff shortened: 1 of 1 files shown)


docs/data-ai/ai/make-decisions-autonomously.md
```diff
@@ -0,0 +1,1059 @@
...
+- A wheeled base component such as a [SCUTTLE robot](https://www.scuttlerobot.org/shop/)
+- A webcam
+- Colored tape, to create a path for your robot
+
+### Configure your machine
+
+Follow the [setup guide](/operate/get-started/setup/) to create a new machine.
+
+Connect your SCUTTLE base to your SBC.
+Add the following `components` configuration to create board, base, and motor components in Viam so you can control your SCUTTLE base:
```
JessamyT, line 31: I would make this way less scuttle-specific. You could link to the tutorial as an example but this feels like too much detail and too specific for an AI section guide.

I'd ditch the config below too since it takes up so much length on the page but won't apply to all bases; just tell them to configure their base with a link to https://docs.viam.com/operate/reference/components/base/

JessamyT (commented):
So high-level, I am curious about the objective of this PR. It seems like it creates a rather lengthy page that duplicates other tutorial and how-to content. Each section in here is covering configuring a (specific) machine, generating module stub code, modifying the code to follow a different pattern from what the generator generates, and also writing logic that makes decisions autonomously.

Would it perhaps be better to have targeted code snippets in a how-to, and then have full things like this in tutorials? Then this page can just focus on the making autonomous decisions and be much shorter?

I'm also curious what the plan is for maintaining these code examples as the platform including the module generator change; the hot reloading flow for example should improve soon.

JessamyT:
Also nit: make sure your PR title is in imperative mood

docs/data-ai/ai/make-decisions-autonomously.md
```diff
@@ -0,0 +1,1059 @@
...
+
+    async def _is_color_there(self, location: str) -> bool:
+        frame = await self.camera.get_image()
+        if location == "left":
+            # Crop logic for left side
+            pass
+        elif location == "right":
+            # Crop logic for right side
+            pass
+        # Implement detection logic here
```
JessamyT, line 318: ?

JessamyT, line 318: Did not review all the code in this PR because was first curious about overall intent of PR, but if this does go forward this code will need to be fixed

JessamyT (changes requested):

PR #4483, opened 2025-07-24
DOCS-3981: Deduplicate module packages

JessamyT:


docs/operate/get-started/other-hardware/module-configuration.md
```diff
@@ -242,6 +242,8 @@ For any version type other than **Patch (X.Y.Z)**, the module will upgrade as so
 If, for example, the module provides a motor component, and the motor is running, it will stop while the module upgrades.
 {{% /alert %}}
 
+If a module appears in both a {{< glossary_tooltip term_id="fragment" text="fragment" >}} and the part configuration, Viam will import the most recent configured version of the module.
```
npentrel, line 245: does it also do that if it's in two fragments? worth checking

JessamyT, line 245: Yes, scope etc. lead me to believe it does. Will make more explicit. @Abalaxy can you confirm this wording? Thanks!
```suggestion
If a module appears in both a {{< glossary_tooltip term_id="fragment" text="fragment" >}} and the part configuration, or in multiple fragments, Viam will import the most recent configured version of the module.
```

Abalaxy: I think that works! I'd tweak it just a tiny bit around what version is picked, maybe `"Viam will import the version that's most recent of the versions imported to the module"`? Just to make it clear that it picks the latest version, but only looking at the "pool" of versions that are specified and imported on the part, not picking the most recent uploaded version of the module

JessamyT: Ah I see what you mean--this is what I meant by "configured version" but maybe that's still not clear. Wording this is tricky. I'll go with
```suggestion
If a module appears in both a {{< glossary_tooltip term_id="fragment" text="fragment" >}} and the part configuration, or in multiple fragments, Viam will import the newest of the configured versions.
```

PR #4567, opened 2025-08-19
Update module-configuration.md

HipsterBrown:
Based on a conversation in [Slack](https://viaminc.slack.com/archives/C06542APFRQ/p1755633241482249), this PR adds information about the persistent directory associated with modules that is exposed as the `VIAM_MODULE_DATA` environment variable when the module is started by viam-server.  cc: @nicksanford


npentrel (approved):
Thank you!

PR #4691, opened 2025-10-07
DOCS-4404: Get machine on WiFi

npentrel:


docs/manage/fleet/change-network.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Connect a machine to a different network"
```
IanWhalen, line 2: kill me for nit, but 'different network' sounds kind of weird, like it's happening in the context of something else (like troubleshooting?) and you're supposed to know what that something else is?  any reason not to just say 'a new network'?

docs/manage/fleet/change-network.md
```diff
@@ -0,0 +1,57 @@
...
+
+Should your machine be able to connect to multiple networks?
+
+- If **yes**, see [**Configure Networks**](/manage/fleet/system-settings/#configure-networks).
+- If **no**, continue on this page.
+
+Can your machine still connect to the old network?
+
+- If **yes**, see [Force provisioning mode](#force-provisioning-mode)
+- If **no**, see [Connect to a different WiFi network](#connect-to-a-different-wifi-network)
```
IanWhalen, line 20: at the risk of being a little harsh, i don't really like this and I also feel like it is a good example of why our docs are so impenetrable.  I have no idea what the first question is even asking.  and the second question seems unnecessary given it just links to the section at thbottom.  we should just get straight to the steps.

IanWhalen (commented):
change requests at top, but central content LGTM!

PR #4897, opened 2026-04-09
docs(hello-world): use correct tflite module name

HipsterBrown:
While going through the new hello world guide (very neat setup with the simulator!), I noticed the naming for the tflite_cpu module was incorrect. This is a quick PR to use the correct naming.


shannonbradshaw (approved):

shannonbradshaw:
Thank you, @HipsterBrown !

PR #4921, opened 2026-04-13
APP-16023 Add query data processing note

katiepeters:
## Summary
Document a shared monthly limit of 100 TB of data processing for `TabularDataByMQL` and `TabularDataBySQL`, including that requests fail after the combined cap is reached and how organizations can ask for a higher limit (support@viam.com). The information is presented as a Note callout on the main query / reference pages readers use, not in the generated Data Client API include.

## What changed
- docs/data/query-data.md — Note callout after the existing SQL lower-bound caution, so people using the in-app query editor see the cap.
- docs/data/query-data-from-code.md — Same note near the top (after the existing tips/cautions) so programmatic SQL/MQL users see it.
- docs/data/reference.md — Same note in the reference flow so schema/query helpers stay aligned with product limits.


shannonbradshaw:
Closed in favor of https://github.com/viamrobotics/docs/pull/4964. I just cherry picked the commits onto main.

PR #4965, opened 2026-04-17
Hardware: document arm-contributed frames, end-effector attachment, and URDF

shannonbradshaw:
## Summary

Addresses the reviewer question on add-an-arm.md: \"Can we describe the after arm frame here, the json, urdf, etc?\"

Adds a new \"## Frames\" section to the Arm page that explains what most readers currently have to reverse-engineer from the motion-planning docs: what frames the arm actually contributes to the frame system, and how to attach things to the end effector.

Covers:

1. **What frames the arm puts in the frame system**
   - `<arm-name>` — the arm's root/tip frame (tracks the end effector as joints move)
   - `<arm-name>_origin` — the static mounting base frame
   - `<arm-name>:<link-name>` — one frame per kinematic link (base_link, link1, link2, …)

2. **Attaching a gripper or camera to the end effector**
   - Use `\"parent\": \"<arm-name>\"`; worked example with a vacuum gripper and a 50mm Z offset for the flange-to-contact distance

3. **Attaching to an intermediate link**
   - Use the `<arm-name>:<link-name>` namespaced form for accessories mounted partway along the arm

4. **Custom kinematics**
   - Noting that Viam-maintained arm modules (UFactory, Universal Robots, Yaskawa) already ship with tuned kinematics
   - Pointing module authors at the arm's `Kinematics` method for URDF or SVA JSON

Also cross-links from the existing \"Configure a frame (recommended)\" step so readers setting up the arm's mounting frame discover the new content.

## Verification

Ground truth from:
- `rdk/robot/impl/local_robot.go` — `createFramesFromPart` and `getLocalFrameSystemParts` (naming convention for arm-contributed frames)
- `rdk/referenceframe/frame_system.go` — flattening that produces the `<arm>:<link>` names
- `rdk/components/arm/fake/kinematics/` — embedded kinematics JSON catalog (ur5e, ur20, xarm6, xarm7, lite6)
- Link naming confirmed against the URDF files; the exact link names (\"base_link\", \"link1\", etc.) come from whatever the arm module's kinematics file declares, so the doc phrases the list as an example with a pointer to `GetFrameSystemConfig` for specifics

## Test plan

- [x] prettier + vale clean
- [ ] Netlify deploy preview: confirm the new section renders and the in-page anchor link `#frames` resolves
- [ ] htmltest passes (internal links go to `/motion-planning/frame-system/` and `/build-modules/write-a-driver-module/`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4966, opened 2026-04-17
Hardware: add Configure a frame step to add-a-camera

shannonbradshaw:
## Summary

The gripper and arm how-tos both have a \"Configure a frame (recommended)\" step that shows the JSON needed to attach the component to its parent in the frame tree. Camera was missing that step, even though wrist-mounted cameras on an arm (and cameras on a mobile base) are both common setups.

Adds a new step 4 \"Configure a frame (recommended)\" to \`add-a-camera.md\` with three patterns:

- **Wrist-mounted on an arm**: `parent` = arm name; the camera's pose tracks the end effector as the arm moves. JSON example included.
- **Mounted on a mobile base**: `parent` = base name.
- **Fixed in the workspace**: `parent` = `\"world\"` with an absolute translation.

Step numbering shifts: old step 4 (Save) → 5, old step 5 (Test) → 6.

Cross-links to:

- `/hardware/multi-machine/cross-machine-frames/` for cameras that live on a remote part (for example, an end-effector Raspberry Pi)
- `/motion-planning/frame-system/` for the full frame reference

Follows from the #4965 scope-down: the reviewer's arm-page comment about attaching accessories to the end effector is handled by the destination component's Configure-a-frame step (gripper already had it, now camera does too).

## Test plan

- [x] prettier + vale clean
- [ ] Netlify deploy preview: verify the new step 4 renders and the three subsections look right
- [ ] htmltest passes (the two outgoing links resolve)

### DIFF
--- a/docs/hardware/common-components/add-a-camera.md
+++ b/docs/hardware/common-components/add-a-camera.md
@@ -99,14 +99,47 @@ You can also set resolution and frame rate:
 }
 ```
 
-### 4. Save the configuration
+### 4. Configure a frame (recommended)
+
+If you plan to use motion planning, computer vision with spatial context, or pose queries that include the camera, add a frame to define where the camera sits in your workspace.
+
+Common mounting patterns:
+
+**Wrist-mounted on an arm.**
+Set `parent` to the arm's name so the camera's pose tracks the end effector as the arm moves.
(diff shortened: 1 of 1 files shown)


shannonbradshaw:
Closing. The reviewer did not ask for a camera-frame step; this was me over-building. If useful, can be proposed separately.

PR #1030, opened 2023-04-12
DOCS-669: remove jetson orin experimental note

sguequierre:
* support is now ready 
* left carrier board/hardware part of note up

PR #5001, opened 2026-04-21
Update module authoring docs for reconfigure removal

shannonbradshaw:
## Summary

Python `EasyResource` no longer includes `reconfigure` (viam-python-sdk #1030, 2025-11-13) and the Python module generator no longer emits it (rdk #5450). Go `Resource.Reconfigure` was removed from the interface entirely (rdk #5811, 2026-04-14); `AlwaysRebuild` and `TriviallyReconfigurable` are deprecated. Docs were still teaching the `new` → `reconfigure` pattern. User-reported confusion on Slack (2026-04-21) surfaced the mismatch: the generator stub no longer matches what the docs describe.

The new guidance: config changes rebuild the resource. Read config and resolve dependencies in the constructor. This matches current generator output in both languages, is durable against further SDK evolution, and gives Python and Go authors the same mental model.

**Files changed:**

- `build-modules/write-a-driver-module.md` — rewrote the "Implement the constructor and reconfigure method" section around default rebuild-on-config-change; dropped `reconfigure` from the complete Python sample; moved dependency resolution into `new`; deleted the "Handle reconfiguration (optional)" step and renumbered (7→6, 8→7).
- `build-modules/write-a-logic-module.md` — folded `validate_config` + `new` + monitor-loop setup into `new`; replaced the Go `Reconfigure` code block (dead API) with prose explaining that `Close` + constructor handles config changes.
- `build-modules/dependencies.md` — moved Python dep resolution into `new`; dropped the broken `super().reconfigure(config, dependencies)` call that would now raise `AttributeError`; rewrote the Go alert to describe rebuild behavior instead of linking to the deleted Step 6.

Net: -118 lines. Vale clean. `make build-prod` succeeds.

## Out of scope for this PR

Other `build-modules/` pages also teach the old pattern and should get a follow-up:

- `build-modules/module-reference.md` (lines 192, 199)
- `build-modules/module-anatomy.md` (line 242)
- `build-modules/write-a-cpp-module.md` (line 323)

And two further-afield pages with broken `super().reconfigure(...)` calls:

- `reference/services/motion/_index.md` (line 108)
- `vision/object-detection/act-on-detections.md` (line 144)

## Reviewers requested

- @naveedjooma for Python accuracy (Naveed authored the SDK + generator removals)
- @aldenh-viam for Go accuracy (Alden authored the Resource.Reconfigure removal)

## Test plan

- [ ] Verify Netlify deploy preview renders all three pages without broken links
- [ ] Naveed confirms Python idiom is correct (inline config-reading in `new` vs. a `_configure` helper)
- [ ] Alden confirms Go guidance matches current module framework behavior
- [ ] Spot-check that the renumbered anchors in `write-a-driver-module.md` (#6-use-the-module-data-directory, #7-add-multiple-models-to-one-module-optional) are not referenced from other pages that would now 404


docs/build-modules/dependencies.md
```diff
@@ -18,7 +18,7 @@ For background on required and optional dependencies, see the
 
 ## The dependency pattern
 
-Every dependency follows three steps: declare it in validation, resolve it in your constructor or reconfigure method, then call its API methods.
+Every dependency follows three steps: declare it in validation, resolve it in your constructor, then call its API methods.
```
vijayvuyyuru: Maybe this is outside the scope of what is expected to be reviewed, but this third stpe feels a little funky. I think just 2 steps would be fine (declare, resolve)

docs/build-modules/write-a-logic-module.md
```diff
@@ -492,44 +482,10 @@ func (s *TempAlert) Close(ctx context.Context) error {
...
-    s.mu.Unlock()
-
-    go s.monitorLoop(monitorCtx)
-    return nil
-}
-```
+On a config change, `viam-server` calls `Close` on the old instance and then
+invokes your constructor again with the new config. `Close` stops the old
+loop; the fresh constructor starts a new one. No separate reconfigure
+handling is needed.
```
vijayvuyyuru: I think we shuold remove the reference to reconfigure here. It feels like a dangling reference to what used to be here.

PR #5014, opened 2026-04-24
try/part-1: rewrite UI steps as CLI commands (experiment)

IanWhalen:
## Summary

- Replaces all Viam app UI interactions in the Part 1 tutorial with equivalent `viam` CLI commands
- Covers machine status, part ID lookup, camera component add/configure, camera verification, fragment apply, and vision service streaming test
- Notes the one gap: fragment variables have no CLI flag today, so that step falls back to JSON config or the app UI

## What a reviewer should know

This is an exploratory PR to see what a CLI-only version of the tutorial looks like end-to-end. It's not necessarily intended to ship as-is — the main open question is how to handle fragment variable setting cleanly from the CLI (the `fragment_mods` JSON workaround in the current draft is a bit rough).

The live video feed from the camera test panel also has no CLI equivalent — the rewrite substitutes log inspection and a `GetImage` API call, which is functional but less immediate.

## Changed files

- `docs/try/part-1.md` — full rewrite of UI steps to CLI commands

### DIFF
--- a/docs/try/part-1.md
+++ b/docs/try/part-1.md
@@ -13,7 +13,7 @@ aliases:
 
 **Goal:** Get a computer vision pipeline working.
 
-**Skills:** Connect a machine to Viam, configure components in the Viam UI, use fragments to add preconfigured services.
+**Skills:** Connect a machine to Viam, configure components with the Viam CLI, use fragments to add preconfigured services.
 
 **Time:** ~10 min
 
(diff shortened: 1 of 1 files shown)


IanWhalen:
@shannonbradshaw this is very much a quick POC.  I'm just kinda curious what you think about the approach.

PR #5027, opened 2026-04-27
Add Delete data page; reduce overview entry to summary [On Hold - Small change to Fix Netlify Check]

shannonbradshaw:
## Summary

- Hoists the "Delete data" how-to out of `/data/overview/` into a dedicated page at `/data/delete-data/` (weight 27, between Tag data and Filter at the edge).
- Uses the flat Viam app / CLI / Python / Go tabs pattern, matching `api-keys.md` and the new `tag-data.md` (PR #5025).
- Reduces the overview's Delete section to a 2-sentence summary plus a link, matching the page's other sections (Query, Export, Annotate, etc.).
- Adds Python and Go code samples for `delete_tabular_data`, `delete_binary_data_by_filter`, and `delete_binary_data_by_ids` — the SDK story was prose-only before.
- Aliases `/data/overview/#delete-data` so existing deep-links still resolve.

## Why

The overview's Delete section was the only H2 on the page that went full how-to with code blocks; every other section (Query, Pipeline, Export, Annotate, Monitor) is a short summary that links to its dedicated page. Pulling Delete out aligns the overview with itself and gives Delete a findable, dedicated page.

## Verification

- Python methods: `viam-python-sdk/src/viam/app/data_client.py` — `delete_tabular_data`, `delete_binary_data_by_filter`, `delete_binary_data_by_ids`.
- Go methods: `rdk/app/data_client.go:765-802` — `DeleteTabularData`, `DeleteBinaryDataByFilter`, `DeleteBinaryDataByIDs`.
- CLI flags: `rdk/cli/app.go:1410-1479` (`viam data delete binary` filter set; `--tags` is not supported, so the page does not claim it).

## Test plan

- [ ] `/data/delete-data/` renders with the four-tab block and the alias `/data/overview/#delete-data` redirects.
- [ ] `/data/overview/` "Delete data" section is a summary, in line with the page's other sections.
- [ ] Sidebar order in Manage data places Delete data after Tag data (weight 27) and before Filter at the edge (weight 31).


btshrewsbury-viam (approved):

btshrewsbury-viam (changes requested):
Netlify deploy-side checks fail while all GitHub Actions checks pass, because the aliases: /data/overview/#delete-data line makes Hugo emit a file at a #-containing path that Netlify's deploy manifest rejects.

btshrewsbury-viam (approved):

PR #5042, opened 2026-05-06
Update docs for Resource.Reconfigure removal (shipped in RDK v1.0.0)

shannonbradshaw:
## Source changes

- viamrobotics/rdk#5944 / reland viamrobotics/rdk#6098: Remove `Reconfigure` from the `Resource` interface; modular resources are now always rebuilt (close + construct) on config change. `ReconfigureResource` RPC is no longer called by `viam-server`.

## Docs changes

- `docs/build-modules/module-reference.md`: Simplified the Python/Go defaults table — removed the "Rebuild on config change" and "In-place reconfigure" rows (rebuild is now universal, not a default-vs-custom behavior). Added a prose paragraph clarifying that `viam-server` always rebuilds modular resources when config or dependencies change. Updated `ReconfigureResource` RPC description: no longer called by `viam-server`, config changes go through `RemoveResource` + `AddResource` instead. Not labeled Deprecated since proto has no `[deprecated = true]`.
- `docs/build-modules/platform-apis.md`: Fix frontmatter description from "validate and reconfigure functions" to "validate function and constructor".
- `docs/data/filter-at-the-edge.md`: Change "reconfigure method" to "constructor" when describing how to access dependencies.
- `docs/tutorials/configure/pet-photographer.md`: Rename Go `Reconfigure` to private `reconfigure` since it is no longer part of the public Resource interface.
- `static/include/components/apis/overrides/protos/sensor.Reconfigure.md`: Updated override text to use "Reconstruct" language.

## Consolidation

This PR now incorporates all changes from #5111. Once this PR merges, #5111 should be closed.

## Merge gate: satisfied

The release containing viamrobotics/rdk#6098 shipped as **RDK v1.0.0** (published 2026-07-07; the version that would have been v0.133.0 — v0.132.0 was the last incremental tag). Verified against the v1.0.0 tag: the `Resource` interface has no `Reconfigure`, and `modmanager` no longer calls the `ReconfigureResource` RPC (v0.132.0 called it from `manager.go:590`; v1.0.0 has only the module-side compatibility handler). The `module.proto` RPC carries no `[deprecated = true]`, matching this PR's wording. #5111 has been closed as superseded by this PR.

## How I found these

- Xref lookup: `search-patterns.yaml` entry for `Resource.Reconfigure` with associated terms (`AlwaysRebuild`, `ReconfigureResource`, etc.)
- Grep matches: full sweep for `Reconfigure`, `reconfigure method`, `reconfigure function` across docs repo

---


docs/build-modules/module-reference.md
```diff
@@ -240,12 +240,16 @@ if __name__ == '__main__':
...
+| Behavior                 | Go                                                           | Python                                                       |
+| ------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| Rebuild on config change | Default (`viam-server` destroys and re-creates the resource) | Default (`viam-server` destroys and re-creates the resource) |
+| No-op close              | Embed `resource.TriviallyCloseable`                          | Default on `ResourceBase`                                    |
+| Skip config validation   | Embed `resource.TriviallyValidateConfig`                     | Default on `EasyResource`                                    |
+
+`viam-server` always rebuilds modular resources when their configuration
+changes. It removes the existing resource instance and creates a new one
+using the constructor. In-place reconfiguration is not supported for
+modular resources in any language.
```
danielbotros: The rebuild row doesn't really fit the structure of this table anymore since there is no default / not-default behavior, the behavior is to always rebuild. It can probably be dropped and the paragraph at the bottom explaining how modular resources are always rebuilt can do the explaining. Also, rebuilds happen when dependencies change as well, if worth mentioning. 

nit: As an aside. this table structure was a little confusing to me. I interpreted as: default behavior if you don't implement a Close method is a no-op close, which embeds `resource.TriviallyCloseable`. But it should be conveyed the other way around. If you embed `resource.TriviallyCloseable` then you get the default behavior of a no-op close.

docs/build-modules/module-reference.md
```diff
@@ -331,13 +335,13 @@ defined in `proto/viam/module/v1/module.proto`:
...
-| `Ready`               | Handshake: module returns its supported API/model pairs. |
-| `AddResource`         | Create a new resource instance from config.              |
-| `ReconfigureResource` | Update an existing resource with new config.             |
-| `RemoveResource`      | Destroy a resource instance.                             |
-| `ValidateConfig`      | Validate config and return implicit dependencies.        |
+| RPC                   | Purpose                                                                             |
+| --------------------- | ----------------------------------------------------------------------------------- |
+| `Ready`               | Handshake: module returns its supported API/model pairs.                            |
+| `AddResource`         | Create a new resource instance from config.                                         |
+| `ReconfigureResource` | Rebuild an existing resource with new config (removes and re-creates the resource). |
```
danielbotros: I think this is clear as-is, and the distinction is mostly one of internal mechanics but since we mention:
> In-place reconfiguration is not supported for modular resources in any language.

consider mentioning the name `ReconfigureResource` is retained for compatibility reasons for more clarity.

static/include/components/apis/overrides/protos/sensor.Reconfigure.md
```diff
@@ -1,2 +1,2 @@
 Reconfigure this resource.
-Reconfigure must reconfigure the resource atomically and in place.
+For modular resources, reconfiguration destroys the existing resource instance and creates a new one using the constructor.
```
danielbotros: Maybe we can move away from the language of reconfigure towards reconstruction? i.e.
```
Reconstruct this resource.
For modular resources, reconstruction destroys...
```

danielbotros (approved):
LGTM, but I would hold off on merging since we reverted the linked PR to let the additional fixes bake and will land the actual remove reconfigure in the next week or so.

PR #5098, opened 2026-06-18
Add motion planning pages: end effector frames, waypoints, pose clouds, plan verification

btshrewsbury-viam:
## What

Adds four new pages under **Motion Planning** and wires them into the section navigation.

### New pages

- **Frame System › End effector frames** — Explains that `Move` targets a named *frame*, not a component: how `component_name` selects the end effector frame, how the `frame` attribute defines a static frame, how `WorldState` transforms extend the frame system for a single request, and how `WorldState` differs from the world state store service.
- **Move an arm › Moving through multiple waypoints** — Routing one plan through ordered goals with `armplanning.PlanMotion`: how `PlanRequest.Goals` chains `PlanState`s into a single continuous trajectory, mixing poses / joint configurations / pose clouds per waypoint, and using `return_partial_plan` to isolate an infeasible waypoint.
- **Move an arm › Pose clouds** — Relaxing an over-constrained goal into a region of acceptable poses with `NewPoseInFrameWithGoalCloud`: the translational/rotational leeways, why tolerances are measured in the target frame, and when a goal region beats an exact pose. Briefly noted and linked from the Configure constraints page.
- **Verifying a Motion Plan** — Planning without executing: `armplanning.PlanMotion` vs `motionService.Move`, assembling a `PlanRequest`, reading the returned `Plan`/`Trajectory`, and reproducing the same plan over the motion service with the `plan` `DoCommand` verb (and its trade-offs).

### Navigation / edits

- Cards added to the Motion Planning index, Frame System overview, and Move an arm overview.
- Configure constraints page gains a short "relax the goal with a pose cloud" pointer + link.

## Notes for reviewers

- Page structure, prose style, and front matter follow the existing Motion Planning pages.
- API usage is grounded in the RDK source, but the Go snippets have **not** been compiled end-to-end, worth a check during review.
- Pre-PR checks pass: prettier, markdownlint, vale (0 findings), and `make build-prod`.


docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
+---
+linkTitle: "End effector frames"
+title: "End effector frames"
+weight: 5
+layout: "docs"
+type: "docs"
+description: "How the motion service moves a named frame to a target pose, and how the frame attribute and WorldState transforms control which point on the robot that frame represents."
+---
+
+When you ask the motion service to move an arm, you do not move "the arm." You
```
shannonbradshaw: This personifies the motion service. That's the wrong tone for docs. As an alternative, you might say

To move an arm using the Viam motion service, you must specify a desired pose for a specific frame defined within the frame system. The motion planner solves for joint angles that place that frame's origin in the destination pose....

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+move a _frame_. The destination you pass is a pose for a named frame in the
+frame system, and the planner solves for joint angles that place that frame's
+origin at the destination. Understanding which frame you are moving, and where
```
shannonbradshaw: Rather than the sentence beginning, "Understanding..." we need to make the abstract description that precedes it concrete -- something like...the frame you specify is typically the frame of the end effector for the arm...

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+that frame sits on the physical robot, is the difference between the tool tip
+landing on the target and the arm's mounting flange landing there instead.
+
+This page explains how the motion service targets frames, how the `frame`
```
shannonbradshaw: I think we can more clearly explain this than saying "targets frames". What does that mean?

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+attribute defines an end effector frame, and how a `WorldState` transform lets
+you add or reposition that frame for a single request.
```
shannonbradshaw: Can you explain what this page is without using so many terms that might not be familiar to the reader before they read the page?

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Move targets a frame, not a component
+
+The `Move` request takes a component name and a destination. The component name
```
shannonbradshaw: We need to be precise. `Move` is a method in the motion service API. Can we be more precise about what the destination is exactly?

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+is not treated as a piece of hardware to actuate directly; it is resolved to a
```
shannonbradshaw: Typically, for docs, we don't want to say what something is not...So this sentence can be The component name resolves to a defined frame in the frame system.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
...
+frame in the frame system. The planner then searches for an arm configuration
```
shannonbradshaw: Problematic to use the word, "configuration" here given how much we talk about configuration generally with Viam.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,185 @@
+---
+linkTitle: "End effector frames"
+title: "End effector frames"
+weight: 5
+layout: "docs"
+type: "docs"
+description: "How the motion service moves a named frame to a target pose, and how the frame attribute and WorldState transforms control which point on the robot that frame represents."
+---
```
shannonbradshaw, line 8: Keeping a few general questions here as I review this:
- What can we expect readers to already be familiar with when they read this page?
- Do we need to summarize what a pose is? Should there be a reference section, perhaps in the overview that we link readers to for more details?
- Same question for world state.

shannonbradshaw: Yeah, I think we need to define pose and frame clearly in a place that we can reference elsewhere. Currently thinking the overview page might be the best place for that, but I'll continue to assess as I review.

assets/motion-planning/frame-system/drill-bit-frame.svg
dgottlieb, line 1: I'm expecting the orientation axis pictured here are from the perspective of the frame relative to the world frame. In that case, I expect that the positive Z axis would be facing "down" for both the `my-arm` and `drill-tip` frames. But let me know if you're observing otherwise.

The same goes for the other svgs, but it's most salient here.

assets/motion-planning/frame-system/end-effector-types.svg
dgottlieb, line 1: I think this caption here has an ambiguous parsing. The first statement makes sense to me, but I'm reading the second as: "Every end effector frame is defined relative to <first clause> and the world frame at the base". If that's the intended parsing, I'm not sure what it's saying.

assets/motion-planning/frame-system/held-object-frame.svg
dgottlieb, line 1: This is more of a question for me -- Before an object (sitting on a table for example) is grasped, a natural orientation vector would be +Z going "up". But when re-parenting the object to a gripper (where positive Z is "down"), that would flip it's orientation vector? That can't be right?

It seems that re-parenting for complex object where that "child" orientation vector matters isn't as simple as just changing the parent.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+    geometry, // the planner avoids collisions with this geometry
+)
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+## WorldState versus the world state store service
+
+The names are similar, but `WorldState` and the world state store service are
```
dgottlieb: I think the goal is to populate motion service requests with items defined in the world state store service. Not sure if that task without a completion date warrants changing the narrative here. But just wanted to be transparent that the two ought to be more similar than they currently are.

The documentation though is accurate w.r.t how things behave today.

docs/motion-planning/verify-a-plan.md
```diff
@@ -0,0 +1,188 @@
+---
+linkTitle: "Verify a plan"
+title: "Verify a motion plan without executing it"
+weight: 60
+layout: "docs"
+type: "docs"
+description: "Use armplanning.PlanMotion to compute and inspect a trajectory before the arm moves, and learn how to get the same plan over the motion service API."
```
dgottlieb, line 7: I'm not sure if there needs to be a documentation change here, just being transparent about the intention of `armplanning.PlanMotion`.

Calling PlanMotion directly is regarded as accessing a "private" API. It's (long term) goal is to only be consumed by the motion service. That said, there are enough limitations (e.g: passing around large meshes may not be feasible) with using the motion service such that using this more directly is helpful/neccessary.

Being considered "private", we do break this API. So there's some risk in adopting this heavily into application code.

The remote section here about using the motion service with the "plan" command is the recommended path. Hopefully we can make it more ergonomic in the future and allow users to make that call without having to proto serialize the MoveReq by hand. Which motivated me to file a ticket for that: https://viam.atlassian.net/browse/RSDK-14155

docs/motion-planning/move-an-arm/multiple-waypoints.md
```diff
@@ -0,0 +1,168 @@
+---
+linkTitle: "Move through waypoints"
+title: "Move an arm through multiple waypoints"
+weight: 40
+layout: "docs"
+type: "docs"
+description: "Plan a single continuous trajectory through an ordered list of intermediate goals using armplanning.PlanMotion."
```
dgottlieb, line 7: Similar to the "verify a plan" comment down below -- using armplanning.PlanMotion is an accepted, but discouraged practice. Again just being transparent in case it changes how we present the documentation here as the PlanMotion API is less stable. I'm bought into there being a need to expose PlanMotion for now -- so no fundamental disagreement on this being included.

I added a line item for engineering work to get motion service and motion planning parity so we can commit to just documenting the motion service as the public API.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,155 @@
+---
```
dgottlieb, line 1: I know you asked to especially look at this, so leaving a note.

This all looks good to me! I think this does a much better job explaining the "follow the tilt" motivation/ease than my code comment did.

The code example at the bottom re: not spilling is definitely what we're striving for what a user should write. I'm skeptical that system reliably achieves that! But see nothing that needs to change there.

If there's anything worth detailing, it's about the sharp edge regarding orientation vector normalization. I'm not sure if the [wine pouring example/documentation](https://github.com/viam-modules/viam-pouring-demo/pull/54/changes#diff-f53cae10b2f3172d89d8b1507bacd4843e22c60e00b72ef88902916da9bb87e5R666-R676) is the most straightforward example. But there's definitely a game where setting a goal pose with OZ: 0 (and leaving OX/OY untouched because of an unconstrained pose cloud) is a bit undefined. Because the goal orientation of OX: 0, OY: 0, OZ: 0 would get normalized back to OX: 0, OY: 0, OZ: 1. And from there having a pose cloud of OX: 1, OY: 1, OZ: <small> no longer does the right thing.

Maybe this goes into the "what is a pose" documentation Shannon was referring to above. Or we just wait on this more esoteric documentation in case we come up with a better idea for resolving that ambiguity most likely preserves user intent.

dgottlieb (commented):
This all looks great. Biggest question is the orientation axis on the images.

assets/motion-planning/frame-system/drill-bit-frame.svg
btshrewsbury-viam: good catch, most cobot arms follow DH, will swap it around.

assets/motion-planning/frame-system/end-effector-types.svg
btshrewsbury-viam: Probably best to remove the second statement, without a reference the comment is confusing on its own, will update

docs/motion-planning/frame-system/end-effector-frames.md
btshrewsbury-viam: Once that goes through I will need to update a bunch then, have similar distinctions in the new Visualization Section

docs/motion-planning/move-an-arm/multiple-waypoints.md
btshrewsbury-viam: Would you feel better if we removed it for the way point example?

docs/motion-planning/verify-a-plan.md
btshrewsbury-viam: Making plan easier to use from the Motion Service's perspective would be amazing. I can switch this to only show the DoCommand verb if that feels more correct. I ended up making a bunch of helpers to make marshalling a bit easier: [https://github.com/viam-labs/viamkit/blob/main/verify/plan.go](https://github.com/viam-labs/viamkit/blob/main/verify/plan.go). The plan verb was extremely important in moving the pallet around

assets/motion-planning/frame-system/held-object-frame.svg
btshrewsbury-viam: I am thinking if the object we picked up had built in frames, it would be relative to some parent. that transform to parent would need to change on pickup. for something like a drill, it could have a grip frame and a drill bit frame, we would need to define a transform from the TCP to the grip frame on pickup so we don't loose the drill bit, and if it wasn't using DH the orientation would probably change axis.

docs/motion-planning/move-an-arm/multiple-waypoints.md
dgottlieb: One of those calls where I think both options are reasonable. Not sure if I'm the right decision maker here. But I suppose I'd argue that waypoints are important, and we should keep this the way it is.

docs/motion-planning/verify-a-plan.md
dgottlieb: Thanks for sharing that code link -- I've added it to [the ticket](https://viam.atlassian.net/browse/RSDK-14155) to give more concrete motivation of what users shouldn't need to write.

> I can switch this to only show the DoCommand verb if that feels more correct.

Again, not sure if I'm the right decision maker here. But given the plan verb can do everything (in the context of verifying -- not in the context of multiple waypoints), opting to get rid of the direct `armplanning.PlanMotion` example makes sense to me.

dgottlieb (approved):
lgtm % updated svgs

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+When you move an arm with the motion service, you move a _frame_ from its current
+pose to a destination pose. That frame is called the **end effector frame**, also called
+the tool control point (TCP) or the manipulation control frame. Where the end effector
+frame sits depends on the arm, the tool, and the task, and that frame can change partway
+through a job. This page explains why the end effector frame exists, how `Move` resolves it,
+and how you define and visualize it with Viam.
+
+First, look at where the end effector frame could be defined for two example use cases:
+
+{{<imgproc src="/motion-planning/frame-system/end-effector-types.svg" declaredimensions=true alt="Two arms drawn schematically. On the left, a two-finger gripper with its end effector frame between the fingertips. On the right, a drill whose end effector frame sits at the bit tip, offset from the arm frame in both x and z. Both panels also show small coordinate axes for the world frame at the base and the arm frame at the flange." style="max-width:820px" class="aligncenter">}}
```
shannonbradshaw, line 40: The vectors in these figures are mostly easy to understand, but the length of the vector leaves us with just the arrow head for most of them. That is so distracting that I think we need to fix it.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+The end effector frame is hardware and task dependent. For a two-finger gripper, the
+end effector frame sits between the fingertips, a short offset below the arm's frame. For
```
shannonbradshaw:
```suggestion
end effector frame typically sits between the fingertips, a short offset below the arm's frame. For
```

shannonbradshaw: Also, we either need to explain what we mean by the "arm's frame" and direct the reader to the figure.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+a drill, it sits at the bit tip, offset from the arm's frame both forward and down,
+and the orientation may be flipped, so a positive z would mean to drill deeper.
```
shannonbradshaw: In what cases would it not be flipped? Is it even flipped? Isn't z always positive moving away from the end of the arm?

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+layout: "docs"
+type: "docs"
+description: "How Move places a named frame at a target pose, and how the frame attribute and WorldState transforms set which point on the robot a frame represents."
+---
+
+When you move an arm with the motion service, you move a _frame_ from its current
+pose to a destination pose. That frame is called the **end effector frame**, also called
+the tool control point (TCP) or the manipulation control frame. Where the end effector
+frame sits depends on the arm, the tool, and the task, and that frame can change partway
+through a job. This page explains why the end effector frame exists, how `Move` resolves it,
```
shannonbradshaw: We're introducing Move for the first time on this page and need to explain what it is.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+{{<imgproc src="/motion-planning/frame-system/end-effector-types.svg" declaredimensions=true alt="Two arms drawn schematically. On the left, a two-finger gripper with its end effector frame between the fingertips. On the right, a drill whose end effector frame sits at the bit tip, offset from the arm frame in both x and z. Both panels also show small coordinate axes for the world frame at the base and the arm frame at the flange." style="max-width:820px" class="aligncenter">}}
+
+The end effector frame is hardware and task dependent. For a two-finger gripper, the
+end effector frame sits between the fingertips, a short offset below the arm's frame. For
+a drill, it sits at the bit tip, offset from the arm's frame both forward and down,
+and the orientation may be flipped, so a positive z would mean to drill deeper.
+When you call `Move`, you have the option to specify which frame you are moving, and
+using the arm frame is always an option. By creating an end effector frame at the
+correct location, a `Move` call becomes easier and better represents the task being
+performed.
```
shannonbradshaw: This entire paragraph should be more directive -- "do it this way"

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+Now look at how the motion service describes the move. The snippets on this page
```
shannonbradshaw: This is worded weirdly. I think we're showing an example next, right? So, we'd write this something like, "See below for an example of..."

docs/motion-planning/frame-system/end-effector-frames.md
btshrewsbury-viam: It typically is z+ out of the last joint for Cobots and industrial arms (they usually use DH), but not for humanoids, where the frame after joint is more arbitrary and typically defined by the zero pose. I can update the examples to be more cobot related

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+)
+```
+
+You pass two parameters to `Move`: a string `component_name` and a `destination`
+pose.
+
+## Which frame Move resolves
+
+The `destination` you pass is a target pose for the resolved frame, and the planner
+tries to find a collision-free motion that moves the frame to the destination.
```
shannonbradshaw: Do we need a bit more here about move all joints through a series of way points to achieve the destination position?

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+The `component_name` can be a component name or a frame name. The motion service
+looks for a frame with that name in the frame system. When you pass a component name
+such as `"my-arm"` or `"my-gripper"`, Viam supplies a frame for it behind the scenes:
```
shannonbradshaw: I think we should just specify that you should use a component name and that Viam uses the defined frame for that component. In general, we need to be directive. Where there is a real "and also", we should state the preferred approach first and then mention the alternative. Here that will just be confusing.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+- **An arm**: its kinematics file (URDF or Viam JSON) defines a single output frame at the
+  terminal link of the kinematic chain, commonly called the tool flange. That output frame
+  is the arm's built-in end effector frame.
```
shannonbradshaw: Here are we just clarifying that for any arm that Viam supports it maintains a frame at the tool flange? We discuss that above around where the graphics are. Here, in part because it's in a bullet, it looks like we're introducing something new. I think all we're really doing here is getting more specific about the gripper frame and trying to distinguish from the arm frame? I think we can simply remind the reader that the arm frame is rooted at the tool flange and refer them to the graphic above. Don't do this in bullets.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+- **Another component**, such as a gripper, defines a static frame through its `frame`
+  attribute, which you set on the component's configuration card.
+
+Here is a gripper `frame` attribute from the [Arm with gripper and wrist camera](/motion-planning/frame-system/arm-gripper-camera/) example, whose origin sits 120 mm past the flange:
```
shannonbradshaw: "origin" is a new term here. We either need to define it or use another word that readers will already understand.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+    "type": "ov_degrees",
+    "value": { "x": 0, "y": 0, "z": 1, "th": 0 }
+  }
+}
+```
+
+## Why the frame you name matters
+
+A `Move` call that uses the arm frame produces different motion than one that
+uses the drill-tip frame. The arm's frame sits at the flange, and the drill-tip
```
shannonbradshaw:
```suggestion
uses the end-effector frame. The arm's frame sits at the flange, and the end-effector
```

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+frame sits at the bit tip, so the same `target_pose` produces different motions.
+In the example below, the target pose is defined on the face of a box. If you use
+the drill-tip frame, the bit moves to the box face. If you use the arm frame, the
+drill drives into the box, because the planner tries to move the arm frame to the
+target pose.
```
shannonbradshaw, line 101: This is such a good example.

docs/motion-planning/frame-system/end-effector-frames.md
btshrewsbury-viam: I would be happy to change it, but I think it would be worth using it to teach frames here. I would want a user to understand clearly the thing they pass in is a frame name. We shield users from frames, albeit a little inconsistently, and to do more useful stuff with the platform they will need to get more comfortable with them.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{<imgproc src="/motion-planning/frame-system/arm-vs-gripper-frame.svg" declaredimensions=true alt="Two diagrams of the same target pose. On the left, the move targets the arm frame, so the flange lands on the target and the drill bit overshoots past it. On the right, the move targets the drill-tip frame, so the bit lands on the target. Both panels show the world frame, the arm frame, and the labeled target pose." style="max-width:760px" class="aligncenter">}}
+
+## Define an end effector frame in code
+
+The `frame` attribute described above works well when the end effector frame is static.
+When the end effector frame moves during a task, or when defining it in code suits
```
shannonbradshaw: So this reads as though I would always need to use the WorldState transform, because at this point, I could interpret my end effector frame as moving any time I call Move. Let's revise this to remove that potential misread.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+the job better, add a new frame to the frame system with a `WorldState` transform.
+
+A great example is a grasped part. Prior to pickup, the end effector frame is located
```
shannonbradshaw:
```suggestion
One example is when your end effector grasps an object. Prior to pickup, the end effector frame is located
```

"great example" is hyperbole we'd like to avoid.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{<imgproc src="/motion-planning/frame-system/held-object-frame.svg" declaredimensions=true alt="A cobot arm whose gripper holds a bent, L-shaped object. The arm frame sits at the flange and the gripper frame at the jaws. An object-tip frame sits at the far end of the object, offset from the gripper in multiple axes and rotated to point along it. Each frame shows labeled x, y, and z axes." style="max-width:600px" class="aligncenter">}}
+
+To extend that frame system for
+one motion, pass a
+[`WorldState`](https://python.viam.dev/autoapi/viam/proto/common/index.html#viam.proto.common.WorldState)
+to `Move` with extra frames. The steps differ slightly by language, but the shape is
+the same:
+
+1. Define the translation and orientation from the parent frame to the origin of the
```
shannonbradshaw: We need to say something like, in the example above, the parent frame is the gripper.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{< tabs >}}
+{{% tab name="Python" %}}
+
+```python
+from viam.proto.common import (
+    Transform, PoseInFrame, Pose, WorldState
+)
+
+# A frame named "object-tip", 100 mm beyond the gripper, attached to it.
```
shannonbradshaw, line 136: The box for the code seems to be narrower than the page on many widths. Use Claude to figure out what you need to add as attributes or styling on the code tabs to use available width and avoid the need to scroll right. I think we can also be better about comment placement to avoid the need to scroll.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+    ComponentName: "object-tip",
+    Destination:   targetPose, // your goal *referenceframe.PoseInFrame
+    WorldState:    worldState,
+})
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+### Attach a geometry to a transform
```
shannonbradshaw, line 191: We're introducing geometry here. We need to define it and differentiate it from the frame.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+A transform can also carry a collision geometry, a shape the planner avoids while it
```
shannonbradshaw: We kind-of define it in this sentence, but let's be explicit and differentiate from frame.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+The names are similar, but `WorldState` and the world state store service are
+different things, and only one affects a `Move` call.
+
+- `WorldState` is the argument you pass to a single `Move` call. The obstacles
+  and transforms in it exist only for that request: the planner uses them to
+  plan this motion, and they are gone when the call returns. The transform in
+  the previous section is a `WorldState` transform, so it shapes the plan.
+- The world state store service is a separate service that holds transforms for
+  client-side visualization in the [3D scene](/motion-planning/3d-scene/).
```
shannonbradshaw, line 261: This is the first mention of world state store service so readers don't even know about it before now. Where will they encounter the service? We need to call that out and make this more of a "heads up at some point you're going to come across the world state store service. Here's what that is and how it differs from WorldState.

docs/motion-planning/frame-system/end-effector-frames.md
```diff
@@ -0,0 +1,282 @@
...
+
+To change where the arm goes, put the transform in the `WorldState` you pass to
+`Move`, as shown above. To render a custom visual that leaves planning unchanged,
+publish it to the world state store service instead.
+
+## Visualize the end effector frame
+
+Static frames render in the 3D scene automatically. Open the **3D SCENE** tab on
```
shannonbradshaw: What's a static frame? We've kind-of introduced that, but should be more explicit here.

docs/motion-planning/move-an-arm/multiple-waypoints.md
shannonbradshaw: @JosephBorodach sorry to ask for a drive-by, but do you have an opinion on this?

docs/motion-planning/frame-system/end-effector-frames.md
btshrewsbury-viam: extended, it looks much better now

docs/motion-planning/frame-system/overview.md
```diff
@@ -284,9 +284,18 @@ If a component's pose looks wrong, check the translation and orientation values
 in its frame configuration. For details on all CLI motion commands, see
 [Motion Service Configuration](/motion-planning/reference/motion-service/#cli-commands).
 
+## Targeting frames in motion
```
shannonbradshaw:
```suggestion
## Target an end effector for motion planning
```

docs/motion-planning/move-an-arm/constraints.md
```diff
@@ -113,6 +113,15 @@ constraints), see
 For `CollisionSpecification` (allow specific frame pairs to collide), see
 [Allow frame collisions](/motion-planning/obstacles/allow-frame-collisions/).
 
+## Relax the goal with a pose cloud
+
+Constraints restrict the _path_ between poses. The complementary tool is to
+relax the _goal_ itself. When close enough is acceptable, give the planner a
+region of acceptable poses instead of a single one,
+which enlarges the solution set and makes planning faster and more reliable. This
```
shannonbradshaw:
```suggestion
which enlarges the solution set and makes planning faster. This
```

shannonbradshaw: I know what you mean by "reliable" but it's possible to interpret that as motion planning might send the end effector to the wrong pose. I think you handle the "it's less likely to fail" scenario in the latter sentences.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+A `pose cloud` replaces the exact goal with a region of acceptable poses. You
+give the planner a target pose plus a set of tolerances, and any pose within
+those tolerances is considered valid. This page explains what a pose cloud
+relaxes and how to use a `pose cloud` with the motion service.
+
+## When to reach for a pose cloud
+
+- **Target regions.** When the destination is a region rather than one exact pose, use a
+  pose cloud. For example, an object that may rest anywhere on a table, or a soup can that
+  may spin about its long axis.
```
shannonbradshaw: I don't think we want to blend placing and picking in one bullet here. let's separate them. Otherwise, it's easy to be confused about how the cloud region is being used.

Also, why a soup can rather than a cup that can be picked up from multiple orientations and at multiple places from top to bottom?

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+- **Dynamic collisions.** When the target area may contain obstacles, a pose cloud lets
+  the planner place the object at a clear spot while avoiding collisions at the destination.
+- **Kinematic constraints.** When a goal fails or plans slowly because the arm struggles
+  to reach that exact pose, relaxing the destination tolerance enlarges the solution space
+  and often turns a failing plan into a success.
+
+Pose clouds are powerful when coupled with motion constraints: together they describe the motion you want without pinning down the destination exactly. Moving a cup of water is a good example. You define an
+orientation constraint to keep the cup upright, but you may not know which exact target
```
shannonbradshaw, line 35: Definitely need to use a cup example above if we're using that example here. Consistent examples throughout a page is the way to go.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+pose lets the arm hold that orientation across its workspace. A pose cloud lets the planner
+choose a destination within a region instead:
+
+{{<imgproc src="/motion-planning/move-an-arm/pose-cloud-cup.svg" declaredimensions=true alt="A cobot arm holds a cup of water above a table. A translucent dashed region on the table marks the pose cloud, with several faded ghost cups inside it showing acceptable destinations. The cup may land anywhere in the region at any rotation while staying upright." style="max-width:760px" class="aligncenter">}}
+
+## Relaxing the destination pose
+
+A pose cloud defines a tolerance on each component of the goal pose.
+Each permits +/- deviation, so the goal becomes a box around the target pose rather than a single
+pose. The destination pose is the center of the box.
```
shannonbradshaw:
```suggestion
A pose cloud is modeled as a box with the destination pose as the center. The size of the box in three dimensions defines a tolerance on the goal pose. A theta component on pose cloud defines a rotation tolerance for the end effector frame. The destination pose may vary as long as it lands within the box and rotation tolerance.
```

docs/motion-planning/move-an-arm/multiple-waypoints.md
JosephBorodach: Agree with Dan's framing. Motion service should be the de facto long term path, and we still need parity between the two before we can commit to documenting only that surface. For now both options make sense, and since multi waypoint isn't expressible through the motion service today, I'd keep the page as is.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+direction on a unit sphere, not angles: on its own, a value of `1` accepts any value for
+that component, and `0` holds it exact.
+
+Because the pointing direction is a unit vector, `OX`, `OY`, and `OZ` share one budget: the
+squares of the three always sum to 1. They move together rather than one at a time. Tilting
+the tool grows `OX` and `OY` and lowers `OZ` at the same time. To let the direction vary,
+widen all three as a set; to hold it fixed, leave all three at `0`. Consider the combination
+`OX: 1, OY: 1, OZ: 0`. It looks like "point in any direction", but `OZ: 0` holds `OZ` fixed,
+and the unit-sphere constraint then keeps `OX` and `OY` at `0`. The tool stays exactly
+aligned despite the wide `OX` and `OY` leeways.
```
shannonbradshaw: This paragraph is super dense and it's value to the reader is unclear. What is the learning objective we're targeting?

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+
+The more freedom you give the planner, the larger the set of arm configurations
+that satisfy the goal, so inverse kinematics is more likely to find a solution
+and the search reaches it faster.
```
shannonbradshaw: I'd cut this paragraph. It's redundant with everything that came before.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Tolerances are measured in the target's frame
+
+The tolerances are evaluated in the reference frame of the target, not in world
+coordinates. This matters whenever the target is tilted relative to the world.
```
shannonbradshaw:
```suggestion
coordinates. This matters when the target is tilted relative to the world.
```

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+Consider a gripper approaching an inclined surface. You want to allow tolerance
+_along_ the surface but stay tight _into_ it, so the tool rides the surface
+instead of driving through it. Expressed in world coordinates that is an awkward
+mix of all three axes, because the surface is tilted. Expressed in the surface's
+own frame, you give large tolerance on the two in-plane axes and tight tolerance on the
+axis normal to the surface.
+Because pose clouds use the target's frame, you describe the tolerance the way
+the task is shaped rather than translating it into world axes by hand.
+
+{{<imgproc src="/motion-planning/move-an-arm/pose-cloud-target-frame.svg" declaredimensions=true alt="Two panels. On the left, a pose-cloud region drawn as a thin slab lying along an inclined surface, large along the surface and tight into it, with the target frame tilted to match the surface. On the right, the orientation tolerance shown as a cone around the tool's pointing direction for OX and OY, plus an ellipse for Theta rotation about the axis." style="max-width:820px" class="aligncenter">}}
```
shannonbradshaw: This figure doesn't work for me at all because we're talking about the frame of the end effector, but I just see a box.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+coordinates. This matters whenever the target is tilted relative to the world.
+
+Consider a gripper approaching an inclined surface. You want to allow tolerance
+_along_ the surface but stay tight _into_ it, so the tool rides the surface
+instead of driving through it. Expressed in world coordinates that is an awkward
+mix of all three axes, because the surface is tilted. Expressed in the surface's
+own frame, you give large tolerance on the two in-plane axes and tight tolerance on the
+axis normal to the surface.
+Because pose clouds use the target's frame, you describe the tolerance the way
+the task is shaped rather than translating it into world axes by hand.
```
shannonbradshaw: This is another example of explaining what you don't do, which is both inefficient and can be confusing. Just explain what you need to do to handle an inclined surface.

docs/motion-planning/move-an-arm/pose-clouds.md
```diff
@@ -0,0 +1,163 @@
...
+
+{{<imgproc src="/motion-planning/move-an-arm/pose-cloud-target-frame.svg" declaredimensions=true alt="Two panels. On the left, a pose-cloud region drawn as a thin slab lying along an inclined surface, large along the surface and tight into it, with the target frame tilted to match the surface. On the right, the orientation tolerance shown as a cone around the tool's pointing direction for OX and OY, plus an ellipse for Theta rotation about the axis." style="max-width:820px" class="aligncenter">}}
+
+## Use a pose cloud with the motion service
+
+A pose cloud attaches to the destination object. In Go, attach a pose cloud to the destination with
```
shannonbradshaw: Attaches to the destination object? This language seems in direct contrast to everything else. We just mean destination pose right?

shannonbradshaw (changes requested):
Thanks for this PR. It fills a massive gap. Many changes requested in comments throughout.

docs/motion-planning/frame-system/end-effector-frames.md
btshrewsbury-viam: This seems to be a preview issue only, i've seen it a few times, claude says we can fix it, but there was some worry that the fix could negatively affect the real deployment, will look into more outside of this pr

shannonbradshaw (approved):
LGTM

PR #5099, opened 2026-06-18
Add Visualization section and reframe Debug a motion plan

btshrewsbury-viam:
## What

Adds a new top-level **Visualization** section, moves the 3D scene tab pages into it, and reframes motion-plan debugging around the 3D scene and publishing a plan as custom visuals.

### New section: `docs/visualization/`

- **Overview** — the four ways to visualize a machine: the in-app 3D scene, the standalone Viam Visualization app, custom apps built on an SDK, and Teleop/dashboards for time series data.
- **Visualizing with the 3D scene** (`3d-scene/_index`) — what the scene renders, where each element comes from, and built-in configuration content vs custom visuals published at runtime.
- **Visuals and collisions** — transform anatomy (reference frame, pose, geometry via the `physical_object` field, metadata, UUID), stable identity via UUID, the geometry types, and the key point that a world-state-store visual is *not* an obstacle the planner avoids (planner geometry comes from the frame system + `WorldState`).
- **Publish visuals from a module** — implement the world state store service (`ListUUIDs`/`GetTransform`/`StreamTransformChanges`), build transforms with the `draw` library, drive a poll-and-update loop emitting `TransformChange` events, and visualize a non-WSS resource by depending on it and pulling its API, plus the machine-config services stanza that puts the module on a machine.
- **Perception** — point clouds, vision-service objects (display and tune `GetObjectPointClouds` output in the scene), and verify point cloud alignment (moved from Motion Planning).
- **Viam Visualization** — run the standalone visualizer (or try the hosted playground), push from a Go client, connect it to a live machine, and save/load scene snapshots (replaces the earlier drawing-library page; alias kept).
- **Troubleshoot the 3D scene** — classify a wrong or missing element by layer: configuration, module, or connection.
- **3D scene widgets** — the **Control widgets** panel (per-resource controls, including `DoCommand` for anything that supports it) and the frame POV widget.
- **Reference** — a new `WorldState` concept page, a **Transform metadata** reference (the seven keys the scene actually reads, with wire formats, verified against the draw library and the scene's parser), a **Geometry constructors** reference (JSON/Python/Go for each geometry type), a **draw library** reference (placement/identity options, shape constructors, colors, metadata, snapshot options, linked to the generated API docs and pkg.go.dev), plus a card to the existing world state store API reference.

### Pages moved into Visualization (aliases added for every old URL)

- `motion-planning/3d-scene/_index.md` → `visualization/3d-scene/the-3d-scene-interface.md`
- `motion-planning/3d-scene/edit-frames.md` → `visualization/3d-scene/editing-frames-visually.md`
- `motion-planning/3d-scene/calibrate-frame-offsets.md` → `visualization/3d-scene/measuring-between-frames.md`
- `motion-planning/3d-scene/verify-point-cloud-alignment.md` → `visualization/perception/verify-point-cloud-alignment.md`
- New: `3d-scene/3d-scene-widgets.md`

### Motion Planning changes

- **Debug a motion plan** moved up to `motion-planning/debug-motion-plan.md` and now covers both the 3D scene checks and the CLI checks; it absorbs and redirects the deleted `debug-motion-with-cli.md`, including that page's "frames are not linked to the world frame" symptom from #5219.
- New **Visualize a motion plan** (`motion-planning/visualize-a-motion-plan.md`) — replay a saved plan in the 3D scene with the Motion Plan Replayer, or convert a plan's trajectory to poses with `FrameSystemInputs.ComputePoses`, build step/goal markers with `draw`, and serve them through a world state store service.
- `motion-planning/3d-scene/set-up-obstacle-avoidance.md` → `motion-planning/obstacles/verify-obstacles.md`.
- Link updates across the frame-system and obstacles pages.
- The world state store API reference page gets a current intro (the **VISUALIZE** tab name was stale) and links to the new implementing how-to.

## Since the first round of review

Addressing @shannonbradshaw's two reviews and @DTCurrie's note, plus main moving underneath the branch:

- **Merged main** and resolved five conflicts. `debug-motion-with-cli.md` was modified upstream while deleted here, so that content is ported into the merged page; the rest take main's newer wording from #5201.
- **Rewrote the widgets page.** The Settings → Widgets tab, the Arm positions widget, and per-camera Camera widgets no longer exist: motion-tools replaced them with the ControlWidgets plugin, which the app ships today. Controls are named for the API method they drive, so an arm's joint-angle readout is now **GetJointPositions**.
- **Corrected the scene UI reference** against motion-tools `origin/main` and then against a live machine: the top-right strip vs the top-center dashboard vs the bottom-right camera controls; the Monitor/Build/Move mode toggle, with editing and **Add frames** gated behind build mode; the **Live updates paused** banner with Undo/Redo; OV and Euler orientation tabs; the settings tab list; and the removal of a **Logs** button the app does not mount.
- **Documented motion plan replay.** `ENABLE_MOTION_PLAN_REPLAYER` defaults on, so the no-code path is real: configure which plans the motion service dumps, sync the directory, then import and scrub. Note that `plan_file_path` alone dumps nothing — it needs `log_planner_errors` or `log_slow_plan_threshold_ms`.
- **Fixed the code findings**: the snapshot example now uses `DrawGeometryOptions`; the module example takes its dependency as a config attribute and returns it from `Validate` instead of hand-writing `depends_on`; `visualize-a-motion-plan` gained prerequisites and a `fmt` import and now calls `goalMarker`.
- **Section structure**, decided against the Diataxis playbook: the constructor reference moved out of the concept page into `reference/geometry-constructors.md`; the `perception/cameras.md` stub folded into the subsection index; the overview's duplicate enumeration collapsed; `world-state.md` also carded from the motion-planning reference index. The `draw-library` tables stay — they already list option categories with the generated docs carrying the full inventory, and all 46 symbols they name still exist upstream.

## Notes for reviewers

- API claims are verified against source: RDK (world state store service interface, `TransformChange`, `ComputePoses`, `armplanning.PlanMotion`, `sensor.FromProvider`, capsule length ≥ 2×radius, CLI motion commands, plan-dump gating), `go.viam.com/api` proto (`Transform.physical_object`, geometry oneof, metadata conventions), `viam-labs/motion-tools` main at v1.41.1 (`draw` API, `api.DrawGeometry`, the ControlWidgets plugin, settings tabs), and the app's 3D-scene route for what it actually mounts.
- UI claims were additionally confirmed on a live machine with an arm, gripper, ML model, generic component, and the motion service.
- The `draw` and `client/api` snippets follow the published guide and are illustrative, not compiled end-to-end.
- The metadata keys documented are the implemented set from motion-tools `draw/transform.go` and `src/lib/metadata.ts`; `collision_allowed` appears only in a proto comment and is neither written nor read, so the docs no longer mention it.
- Not documented on purpose: VR/AR, and the LLM Scene Builder beyond a toolbar mention.
- Pre-PR checks pass: prettier, markdownlint, vale (0 findings), and `make build-prod`. Every removed or moved URL has an alias.


shannonbradshaw (commented):
I independently verified this PR against current source (rdk `cf13e7a16`, api `a05c12b5`, viam-python-sdk `6a6e035ba`, viam-labs/motion-tools `8627748`) and the deploy preview. The verification claims in the PR body held up well — nearly everything checks out against today's main:

- **World state store**: the service interface snippets match rdk exactly (`ListUUIDs`/`GetTransform`/`StreamTransformChanges` signatures, `TransformChange` fields, `NewTransformChangeStreamFromChannel`, `world_state_store.go:116–144`), and `"api": "rdk:service:world_state_store"` matches the SubtypeName. The Python `viam.services.worldstatestore` package exists as claimed.
- **Proto**: `Transform.physical_object` (`common.proto:160`), the geometry oneof (sphere/box/capsule/mesh/pointcloud), and every field name in the Python/Go geometry examples (`dims_mm`, `radius_mm`, `length_mm`, `content_type`/`mesh`, `point_cloud`) are exact.
- **rdk**: `Trajectory` is `[]referenceframe.FrameSystemInputs` and `ComputePoses(fs)` matches `input.go:97`; `NewSphere`/`NewBox`/`NewZeroPose`/`NewPoseFromPoint` signatures match; capsule validation is exactly "length ≥ 2×radius" (`NewCapsule` errors below, converts to a sphere at equality); all four CLI motion commands and their flags match `cli/app.go:3261–3300`. Notably, the reframed debug page **corrects** the old page's deprecation claim: `builtin.go:281` confirms motion `GetPose` is deprecated in favor of the *frame system service's* GetPose, not `Robot.GetPose` as the deleted page said.
- **draw library**: every option and constructor in the reference tables exists in current motion-tools main (`drawable.go`, `color.go`, `drawing.go`, `geometry.go`, `scene.go`), the seven metadata keys match the scene parser's exact allowlist (`src/lib/metadata.ts:20–27`, "Unknown keys are silently ignored"), the `api.DrawGeometry` update-in-place semantics are verbatim from the doc comment, and the `visualization_snapshot` filename prefix is enforced in the FileDrop loader.
- **Structure and links**: every moved/deleted URL has an alias and the renames properly remove the old pages; all six visualization-site links return 200; the `docs/examples/worldstatestore` module exists; all internal link targets exist on main (including the `#geometry-types` anchor); weight 165 slots the section right after Motion planning (160); alias redirects work on the deploy preview. CI is fully green.

Three findings to fix before merge, all small:

**1. The snapshot example uses an API that changed after this PR was authored.** motion-tools `ed62728` (2026-07-02, "update snapshot api to v2 style inputs returns") moved `Snapshot.DrawGeometry` to options-struct style returning `([]byte, error)` — the example on `viam-visualization.md` (`snapshot.DrawGeometry(box, boxPose, "world", draw.ColorFromName(...))` with a bare `err`) matches the pre-July API. It was correct when written; it needs updating to the `DrawGeometryOptions` form. Worth a quick scan for the same drift pattern if this sits longer — everything else survived, this was the only casualty.

**2. Widgets page: the arm widget claims don't match the UI.** The Widgets settings section headers are **"Arm widgets"** (one toggle per configured arm) and "Camera widgets" (`WidgetSettings.svelte:20,45`) — not a single "Arm positions" entry. And there is no **Select arm** dropdown in the widget: each arm gets its own widget, titled with the arm's name (`ArmPositions.svelte:34`). The "If your machine has more than one arm, a Select arm dropdown…" sentence should be replaced with the per-arm-toggle behavior.

**3. The camera resolution list is one camera's options presented as the general set.** The dropdown is "Default" plus whatever resolutions the camera reports through its stream options (`Camera.svelte:105,152`) — the fixed list `1280x720 … 80x44` is what one particular camera returned. Suggest "**Default** plus the resolutions the camera reports," keeping the list as an example if useful.

Nits, take or leave: the POV panel title renders as `POV: <frame-name>` (`FramePov.svelte:150`), not just "POV"; and in `visualize-a-motion-plan.md`, `goalMarker` is defined but never called in the page's flow and the snippet's import block omits `fmt` (the PR already declares snippets illustrative, so flagging only for completeness).

One discovery beyond this PR's scope, noted here because this review surfaced it: rdk removed `Reconfigure` from the `resource.Resource` interface on 2026-06-23 (`786454e3c`; only a deprecated `BuiltInReconfigure` compat method remains). The generated API includes sitewide still document a Go `Reconfigure` method — that's a generator/regeneration follow-up, not something this PR touches.

shannonbradshaw (commented):
Follow-up to my earlier review: a second, adversarial pass through structural, journey, and consistency lenses rather than claim verification. The first review checked whether what the pages say is true; this one asks whether the pages work. Nine further findings, roughly ordered by reader impact. (For completeness: this pass also re-verified more UI and API claims that all held — the `TransformChangeType` enum values, the Settings tab labels (Connection/Scene/Pointclouds/Vision/Widgets/Debug), the `supports_pcd` auto-disable (`usePointclouds.svelte.ts:82`), the "Arm Models" scene setting, and the vision polling-rate label under Connection.)

**1. `visualize-a-motion-plan.md` cannot be started from the page itself.** The first code block uses `plan` and `fs` unbound — the page never shows where either comes from, has no Prerequisites section, and only mentions `armplanning.PlanMotion` in the closing comparison. A reader landing here (including from the debug page's prominent link) has no path to a `plan` in hand. Structural fix: a Prerequisites block stating you need a plan from `armplanning.PlanMotion` and a `*FrameSystem`, linking to [Verify a plan](/motion-planning/verify-a-plan/) which teaches exactly that.

**2. `publish-visuals-from-a-module.md` teaches a dependency pattern that contradicts the module docs.** The constructor hardcodes `sensor.FromProvider(deps, "obstacle-sensor")` and the config section tells users to hand-write `depends_on`. `build-modules/dependencies.md` teaches the canonical pattern: the dependency name comes from a config attribute, `Validate` returns it as an implicit dependency, and the constructor reads it from the validated config. As written, renaming the sensor silently breaks the module, and the two sections disagree about how Viam modules declare dependencies. The example should take the sensor name as an attribute and drop the manual `depends_on` stanza (or explain why it deviates).

**3. Redundant layering at the top of the section.** `overview.md` ("The 3D scene" section) and `3d-scene/_index.md` ("What the scene renders") enumerate the same four element types in nearly the same words, and `the-3d-scene-interface.md` is a third page a reader must cross before doing anything. Consider collapsing the overview's 3D-scene blurb to one sentence plus the card, so the enumeration lives in exactly one place.

**4. Diataxis mixing in `visuals-and-collisions.md`.** The page is a concept page (transform anatomy, UUID identity, scene-vs-planner) that carries a five-type × three-language constructor reference in its back half. The constructors are reference material that would sit better as a page under `visualization/reference/` (which already exists), keeping the concept page readable and giving the constructors a findable home.

**5. `perception/cameras.md` is a stub.** 31 lines, three sections, each of which mostly links elsewhere (widgets page, point-clouds page, frame config). It exists to give Perception a "Cameras" card, but a reader who clicks it gets three redirects. Either fold its content into `point-clouds.md`/the widgets page, or give it the camera-specific content it promises.

**6. Stale link text in `cli-commands.md`.** Both references still read "Debug motion with the CLI" while pointing at the merged page now titled "Debug a motion plan." Link text should match the target.

**7. `reference/world-state.md` is only discoverable from the Visualization section.** WorldState is a motion-planning concept (it's the `Move` argument), but the new reference page is carded only from `visualization/reference/`. A motion-planning reader looking for a WorldState reference in their own section's reference area won't find it. Suggest carding it from `motion-planning/reference/` too.

**8. "Perception" as a nav label collides with the Computer vision section.** The docs already have a top-level Computer vision section; a sidebar entry "Perception" under Visualization is exactly where a vision-minded reader will also look, and `vision-services.md` here overlaps that section's territory (tuning segmenters). Consider a label that says what the pages actually cover — "Camera data in the scene" or similar — and a prominent cross-link to Computer vision.

**9. The `draw-library.md` option tables are a standing maintenance liability.** They mirror, option by option, a viam-labs library the page itself says "moves faster than the RDK" — and the snapshot-API drift (finding 1 of my earlier review) happened within two weeks of this PR being authored. The page already links the generated API docs and pkg.go.dev as "the full and current inventory." Suggest either trimming the tables to option *categories* with the authoritative links carrying the detail, or accepting the mirror and putting motion-tools under the same drift monitoring as the other documented repos.

None of these block the PR's core move (the section, the renames, the aliases — all sound); 1, 2, and 6 are the ones I'd fix before merge, since they produce concrete reader failures.

docs/visualization/3d-scene/3d-scene-widgets.md
DTCurrie, line 1: This has been expanded to include all test widgets by resource API

btshrewsbury-viam: Thanks — you were right, and it was bigger than a page tweak. The Settings → Widgets tab, `WidgetSettings.svelte`, and `ArmPositions.svelte` are gone, so I rewrote the page around the ControlWidgets plugin instead of editing it: the **Control widgets** button in the top-right strip, resources grouped by type with an open-count per row, one toggle per control, each opening as its own `<resource> · <control>` panel, plus the **Overview** / **Move** / **DoCommand** fallbacks and per-part persistence.

I confirmed it on a live machine rather than guessing at labels, which also recovered the one thing the old page did that I'd otherwise have dropped: the arm's joint-angle readout is now **GetJointPositions**, so that has its own short section.

Two things I'd value your eye on:

1. The page says control widgets come from the same library as the **CONTROL** tab, and notes that the tab omits the motion service so **Move** is specific to the scene. Is that the framing you'd use?
2. I left the widget labels generic apart from the arm example, since `@viamrobotics/test-widgets` isn't vendored anywhere I could read. If the per-API label set is stable enough to enumerate, I'm happy to.

While I was in there I also corrected the interface page against motion-tools main: the top-right strip vs the top-center dashboard vs the bottom-right camera controls, the Monitor/Build/Move toggle with editing and **Add frames** gated behind build mode, the **Live updates paused** banner, the OV/Euler orientation tabs, the settings tab list, and a **Logs** button the app doesn't mount. The Motion Plan Replayer is documented too, now that the flag defaults on.

DTCurrie: > The page says control widgets come from the same library as the CONTROL tab, and notes that the tab omits the motion service so Move is specific to the scene. Is that the framing you'd use?

yep, that would be https://github.com/viamrobotics/test-widgets

> I left the widget labels generic apart from the arm example, since @viamrobotics/test-widgets isn't vendored anywhere I could read. If the per-API label set is stable enough to enumerate, I'm happy to.

should be stable, we keep a registry of them in the above repo

btshrewsbury-viam: @DTCurrie  Can we merge or do you see any blockers you would want fixed first? Need a PR approval to allow merge

DTCurrie (approved):

PR #5181, opened 2026-07-12
Add SO-ARM101 miniature palletizing workshop

HipsterBrown:
## Summary

Adds a new sequential-workshop tutorial, **Miniature Palletizing with the SO-ARM101**, at `docs/tutorials/so-arm101-palletizing/` — the second workshop in the collection seeded by pick-and-place, reusing that format with no new layouts or shortcodes.

An affordable desktop 5-DOF arm packs eight cubes (two layers of four) onto a printed pallet mat:

- **Phase 1** platform mental model (three layers, components/services, the frame-system / motion / WorldState concepts)
- **Phase 2** configure the arm + gripper via the discovery service; place both in the frame system
- **Phase 3** teach two anchor poses by hand (freedrive) and read them from the arm's test card
- **Phase 4** build `palletizer.py` for a static bottom-layer pack — **milestone one**
- **Phase 5** model placed cubes as WorldState obstacles and the held cube as a gripper-frame transform for the full collision-free two-layer pack — **milestone two**
- **Phase 6** (optional) package the loop as an inline module
- **Wrap-up** review + next steps

Companion code lives in [`viam-devrel/mini-palletizer`](https://github.com/viam-devrel/mini-palletizer): `helpers.py`, the finished `reference/palletizer.py`, a machine-config fragment, and a printable cube + pallet-mat template (20 mm cubes, 30 mm pitch). A rendered isometric grid diagram ships in `assets/`; the remaining screenshots/videos are marked with `<!-- ASSET ... -->` placement comments (shot list tracked separately).

## Test plan

- [ ] `make build-prod` passes; the workshop builds into production and the overview appears as a card on `/tutorials/`
- [ ] prettier, markdownlint, and vale are clean across the workshop
- [ ] Netlify preview: overview + all six phases + wrap-up render; prev/next chain and workshop sidebar intact; the Phase 4 `grid-iso` diagram renders
- [ ] Follow-ups before the assets and code are final: produce the ASSET screenshots/videos, run the pack on real hardware to validate the caveated approximations (held-cube collision spec, placed-cube z-offset, grasp depth)

Pages are published (draft flag removed) so this content is included in the Netlify preview for review.

https://claude.ai/code/session_01EdtYcpEHeA2yvQ2ksBzZHH


jeremyrose-viam:
Hey Nick — Jeremy and I ran the whole workshop end-to-end on a real SO-101, built a real `palletizer.py` against the companion repo, and diffed it against `reference/palletizer.py`. Diff of everything since your last commit: https://github.com/viamrobotics/docs/compare/so-arm101-palletizing-workshop...jeremyrose-viam:so-arm101-palletizing-workshop

**Changes we'd advocate for:**
- `helpers.ARM` is a `ResourceName`, but `MotionClient.move()`'s `component_name` takes a plain string — passing it as-is throws on the very first move. Fixed at each call site (`helpers.ARM.name`) in Phase 4, Phase 5, and `reference/palletizer.py`.
- Your original design taught the anchor pose at a cube's top and descended past it by `GRASP_DEPTH`. We're proposing a simplification: teach the pose at the bare table surface and adjust *upward* (`GRASP_HEIGHT`) instead, since that doesn't require a cube to be present to teach against.
- Replaced `.grab()`/`.open()` with `grip_percentage`/`do_command` in Phase 4/5 and the reference — this isn't just "tighter control" in the abstract: `.grab()` overloaded Jeremy's gripper servo trying to close with full force, and `.open()` swings the gripper wider open than the routine needs.
- Re-named the `move` smoke-test into a `test` method explicitly framed as reusable scratch space — readers build it once, then overwrite its body a second time later in the phase to test `grip_percentage` before ever writing `pick`. Removed the equivalent step entirely from `reference/palletizer.py` rather than renaming it — a permanent method didn't fit a finished reference the way disposable scratch space fits a tutorial.

**Open questions for you specifically** (module-level, not doc-level):
- The `-105` gripper-length correction in `place()` (compensating for the arm's reported point being the gripper's base, not its tip) — is that value surfaced anywhere from the SO-101 module, or does it only exist as tribal knowledge? And does the motion service treat gripper-frame moves the same way it treats arm moves, given the arm's kinematics are adjusted for its limited DOF? If gripper-frame moves aren't adjusted the same way, that changes whether this belongs in Python at all vs. Phase 2's gripper frame config.
- Relatedly: is there a real gripper-length spec for the module we should be citing instead of a bare magic number?

One more note: we ran the full two-layer pack live and it completed with zero planning/collision errors, but physical accuracy was rough on this specific unit (drift, uneven placement, non-straight vertical motion). We're treating that as hardware/calibration on this arm, not a doc/logic problem — the pick/place/obstacle logic itself seems sound — and plan to validate on different hardware before drawing conclusions.

Companion repo changes are on Jeremy's fork: https://github.com/jeremyrose-viam/mini-palletizer/tree/update-reference-palletizer — happy to open a PR from there if all of this looks good to you.

PR #5184, opened 2026-07-13
Document GetStatus across component and service APIs

btshrewsbury-viam:
## What this adds

The Python, TypeScript, and Flutter SDKs expose a resource-level `GetStatus` method (backed by the common `GetStatus` RPC in [viamrobotics/api](https://github.com/viamrobotics/api)) on every component and most services. It was undocumented: no rows in `sdk_protos_map.csv`, so it never appeared in the [API reference](https://docs.viam.com/dev/reference/apis/) — and it accounted for ~70 of the 119 warnings keeping the weekly **SDK method coverage** job red (#5142).

This PR adds a `GetStatus` section to all 18 component API pages and 8 service API pages (data manager, discovery, generic service, ML model, motion, SLAM, vision, world state store).

## How

- One `GetStatus` CSV row per resource, filling **only** the SDK columns the coverage scraper actually finds: Python `get_status` and TypeScript `getStatus` broadly; Flutter `getStatus` where Flutter supports the resource (no encoder, input controller, switch); Go `Status` everywhere except world_state_store (the Go method lives on `resource.Named`, which every resource embeds; the Go parser hand-injects inherited methods, and `Status` joins `DoCommand`/`Close`/`Name` in that injection — see the follow-up commit prompted by Naveed's review). world_state_store's Go column stays empty because the Go parser skips that resource entirely (its page has no Go tabs for any method; pre-existing gap). Micro-RDK column set to `No` where the resource's section uses Yes/No (micro-rdk has no per-resource GetStatus RPC).
- One proto description override file per resource (`<resource>.GetStatus.md`).
- Regenerated the affected component/service markdown with `update_sdk_methods.py`. The regeneration also picks up drift since the last regen — mostly the `/reference/` → `/dev/reference/` link migration in tables and minor upstream SDK docs wording — which is why the table diffs touch every row.
- One generator tweak: ignore the proto-cased `GetStatus` duplicate the Python scraper picks up from the world_state_store page, exactly matching the existing ignore entries for that page's other duplicates (`ListUUIDs`, `GetTransform`, `StreamTransformChanges`, `DoCommand`).

## Verification

Ran the full coverage script locally on this branch: the ~70 GetStatus warnings are gone (119 → 49 total), no missing-override warnings, and the generated sections render Python/TypeScript/Flutter tabs with scraped signatures and examples.

Part of a series addressing the #5142 warning backlog; companion PRs document the remaining component, robot, and app methods.

Refs #5142


HipsterBrown (approved):

.github/workflows/__pycache__/parse_flutter.cpython-314.pyc
HipsterBrown, line 1: Should we be committing these cache files?

shannonbradshaw (commented):
I independently verified this PR against source at current main (api `a05c12b5`, rdk `cf13e7a16`, viam-python-sdk `6a6e035ba`, viam-typescript-sdk `ca10ef2c9`, viam-flutter-sdk `e8492a71c`) and the deploy preview. The core claims all check out:

- The `GetStatus` RPC exists on every component and service proto this PR documents (all backed by `common.v1.GetStatusRequest/Response`).
- The hand-injected Go entry in `parse_go.py` matches rdk exactly: `Status(ctx context.Context) (map[string]interface{}, error)` on `resource.Resource` (`resource/resource.go:78`) and `Named` (`resource.go:290`), and the injected description is the code comment verbatim. The `FromProvider` example pattern is valid (spot-checked `arm`, `generic`, `mlmodel` packages).
- The CSV column choices match SDK reality, including the non-obvious gaps: Python `get_status` genuinely absent from the data manager client, TypeScript `getStatus` genuinely absent from the ML model client, and the empty Flutter cells for encoder/input controller/switch are correct against the published flutter.viam.dev docs the scraper reads (`Encoder-class.html` has no `getStatus` today). Note the Flutter repo's main branch *does* have encoder `getStatus` (added 2026-03-31, RSDK-13649), so the weekly job will flag these cells for filling after the next Flutter docs publish — self-correcting, nothing to do now.
- Rendered output verified on the preview: `#getstatus` anchors render, Python inherited-method anchors resolve on python.viam.dev, `flutter.viam.dev/viam_sdk/Resource/getStatus.html` returns 200, and the world-state-store section correctly has no Go tab.

Two findings:

**1. The movement sensor page's method table is missing the `GetStatus` row.** Movement sensor is the one component page whose table is a hand-maintained include (`static/include/components/apis/movement-sensor.md`) rather than a generated `*-table.md`, and it wasn't updated. On the deploy preview, the `GetStatus` section renders on the page but the only `#getstatus` reference is the external ts.viam.dev link — no table row points to it, unlike the other 17 component pages. Fix: add a `GetStatus` row to the hand-maintained include (with that table's support-model and micro-RDK columns).

**2. Coordination with #5191** (flagged there too): that PR reformats `slam.md`/`slam-table.md`, which this PR also touches — whichever merges second needs a rebase/regen. And once #5191 restores the navigation API page, navigation becomes the only service page without a `GetStatus` section even though both SDKs expose it (`client.py:101`, `client.ts:152`) — worth a follow-up CSV row + regen after both land.

One small note: the new angle-bracket shortcode stripper in `update_sdk_methods.py` isn't mentioned in the PR body — it looks correct and several existing override files do contain angle-bracket shortcodes (`base.GetProperties.md`, `motion.MoveOnGlobe.md`), but a one-line mention of which override triggered it would help the next reader of the generator.

njooma (approved):

PR #5191, opened 2026-07-13
Restore SLAM and navigation reference pages as deprecated orphan pages

btshrewsbury-viam:
## What this does

RDK still has Nav and Slam links, and there is a small chance a user is using a pinned version of RDK. This PR brings back a few pages, marks them as deprecated, and "orphans" these pages, so you can get to them if you have the link, but you can not get to them from the navigation. This fixes a few CI related failures checking links work from RDK / API to Docs, and is a better backwards compatible option.  

RDK godoc doc-comments link to SLAM and navigation **service** and **API** reference pages. Those pages were gone: SLAM had no page at all (both URLs dead-ended at a generic index), and navigation had bare "has been removed" tombstone pages that dropped the API method anchors. That left **17 broken documentation links** from RDK (SLAM: 6, navigation: 11), tracked in #5152 and #5154.

Both services are deprecated but **may return**, so rather than delete-and-redirect, this restores all four pages at their canonical URLs as **deprecated, orphaned reference pages**:

- Each page renders normally with a yellow **"Discontinued"** banner at the top.
- Each is reachable **only by direct URL** — excluded from the sidebar, the sitemap, and the search index (`_build: { render: always, list: never }`).
- All method anchors resolve, so every RDK link (including `#getposition`, `#getmode`, …) lands correctly.

This keeps the deprecated services out of the site's navigation while un-breaking the RDK links, and is trivially reversible if the services come back (swap the banner / re-list the page).

## Changes

| URL | Change |
| --- | --- |
| `/reference/services/slam/` | New minimal service page |
| `/reference/apis/services/slam/` | New API page (SLAM API includes already present) — anchors: `getposition`, `getpointcloudmap`, `getinternalstate`, `getproperties` |
| `/reference/services/navigation/` | Restored full reference (replaces the "removed" tombstone) |
| `/reference/apis/services/navigation/` | Restored API page + regenerated the 2 deleted navigation API include files — 9 method anchors |

Supporting edits: removed the SLAM alias collision from `reference/apis/_index.md`; removed the navigation cards from the service/API index pages (so the pages stay out of nav); restored 5 navigation images; pruned/repaired dead links in the restored navigation content (dead `movement-sensor#configuration` fragments dropped, self-references made same-page, frame-system links modernized).

> [!NOTE]
> This **reverses #5067** for navigation (which had removed the navigation section and pointed it to motion planning). That was intentional — the decision here is to preserve the reference behind a deprecation banner rather than redirect, since navigation may be un-deprecated. The banner deliberately carries **no version number** (SLAM is still registered as a builtin in RDK `main`, so there is no clean "removed in vX.X").

## Verification

Verified against a local production build (`make build-prod`, exit 0):

- **13/13 method anchors render** (9 navigation + 4 SLAM).
- **0 dead links and 0 broken anchor-fragments** across all four pages (checked following alias meta-refreshes).
- **Orphaned from all three surfaces**: absent from the sidebar nav, the sitemap, and the Typesense search index (while sibling reference pages index normally).
- prettier, markdownlint, and vale all pass.
- No forced `netlify.toml` redirect shadows any of the four URLs (the `/reference/services/*` rule is non-forced, so it self-heals once the pages exist).

Fixes #5152, fixes #5154


shannonbradshaw (commented):
I independently verified this PR's claims against source (rdk `cf13e7a16`, api `a05c12b5`, app `844583faf`, viam-python-sdk `6a6e035ba`, viam-typescript-sdk `ca10ef2c9`) and the deploy preview. The core of the PR holds up well:

- rdk main's `services/navigation/navigation.go` doc comments link to exactly these URLs and anchors, so the broken-link problem and the chosen fix are sound.
- All 13 navigation method entries match the rdk interface (`Location` → `*spatialmath.GeoPose`, `RemoveWaypoint` → `primitive.ObjectID`, `Properties(ctx)`), and both the Python and TypeScript SDKs still ship navigation clients with matching signatures.
- Every default in the attributes table matches `services/navigation/builtin/builtin.go`.
- All four preview URLs return 200, all 17 anchors render, the pages are absent from the sitemap, and every internal link target exists on `main`. The forced `/reference/apis/services/` redirect is exact-path so it doesn't shadow the new pages, and the SLAM skipPatterns are properly paired with real edge 301s.

A few things worth addressing while restoring, plus coordination notes:

**1. Config Builder steps don't match the current app UI.** The restored page says "select **Configuration block** … Enter a name or use the suggested name … click **Create**." The current menu item is **Blocks** (`ui/src/lib/components/robot/add-resource-menu/menu-items.svelte`), which opens a "Configuration Blocks" dialog with card-based selection; I found no name-and-Create step in that flow. Since the service is discontinued, `navigation` likely no longer appears in the catalog at all. Suggest trimming the Builder tab or noting the UI may no longer offer this service.

**2. The JSON Example block is invalid JSON.** There's a missing comma after the `obstacle_detectors` array, and `movement_sensor`, `base`, `obstacle_detectors`, and `obstacles` sit outside `"attributes"`, though the Config struct in `builtin.go` reads all of them from attributes (the JSON Template tab nests them correctly). This defect exists verbatim in the pre-removal original (`ade0e966e~1`), so it's faithfully restored rather than introduced, but restoration seems like the right moment to fix it.

**3. Attributes table vs `builtin.go`:** the table is missing `map_type` (default GPS), `replan_cost_factor` (default 1.0), and `log_file_path`. Also, `movement_sensor` is marked Required but is only required when `map_type` is GPS, and `store` is marked Required even though an unset store validates and defaults to memory (the row's own "Default: memory" contradicts the Required flag).

**4. Coordination with #5184:** that PR adds `GetStatus` to `slam.md`/`slam-table.md`, the same generated files this PR reformats, so whichever merges second will conflict or need a regen. The regenerated navigation includes here also omit `GetStatus`, which both SDKs now implement (`client.py:101`, `client.ts:152`), so navigation will be the one API page missing it once #5184 lands.

**5. "There are two options for modes" is inaccurate** — the proto defines `MODE_EXPLORE = 3` and rdk supports it (the TypeScript tab on this same page lists it). The sentence comes verbatim from the Python SDK docstring, so the real fix is upstream, but it is an internal contradiction on the rendered page.

**6. Pre-existing generator defects (informational, not blockers):** the Python `do_command` example (`from_robot(robot=machine, "my_navigation_svc")`) is a syntax error, the `add_waypoint` example has broken indentation (traced to the SDK docstring, `navigation.py:134`), and the Go `GetResourceName` example (`err := my_nav.Name()`) doesn't compile. These match the sitewide generator pattern (identical defects exist in `vision.md` on main), and CI can't catch them: the flake8 job only globs `docs/**/*.md` and runs `continue-on-error`. Probably worth an upstream issue rather than hand-edits here.

**7. Minor:** the `obstacle_detectors` row links `"vision_service"` to `/reference/services/motion/` instead of the vision service page, and the "Navigation concepts" bullet list uses title case ("Compass Heading", "Angular Velocity") while the headings it points to are sentence case.

Items 1–3 seem worth fixing in this PR; 4 needs a merge-order decision with #5184; 5–7 are fine to defer or file upstream.

jeremyrose-viam (approved):

PR #5187, opened 2026-07-13
Document registry deprecation and machine summary methods on the fleet page

btshrewsbury-viam:
## What this adds

Six released fleet methods flagged by the weekly **SDK method coverage** job (#5142), all on the [Fleet Management API page](https://docs.viam.com/reference/apis/fleet/):

- **DeprecateRegistryItem / UndeprecateRegistryItem** and their **Version** variants (Python + TypeScript) — joining the already-documented registry item CRUD methods. From [api#859](https://github.com/viamrobotics/api/pull/859) (APP-14575).
- **ListMachineSummaries** (Go) — fleet-wide machine summary listing, filterable by fragment or location. From [rdk#5608](https://github.com/viamrobotics/rdk/pull/5608) (APP-14503).
- **GetRobotPartByNameAndLocation** (Go) — fills the missing Go column on the already-documented proto row.

The `UploadImageToDatasets` data manager change originally in this PR moved to #5193 so the two can be reviewed independently — they share no generated files.

## Deliberately not included

The Go **DataClient** methods the coverage job also flags (sequences CRUD, the app-side `UploadImageToDatasets`) cannot currently be documented on the data client page — the Go scraper attributes every `go.viam.com/rdk/app` package method to the fleet page, and the data client page has no Go tabs. Tracked in #5190 rather than misplaced here.

## Merge order

**Stacked on #5184** (shared regeneration baseline). Merge #5184 first; GitHub retargets this automatically. Refs #5142 but does not close it (remaining warnings tracked in #5188/#5189/#5190).

## Verification

Targeted coverage runs: all six warnings cleared, no missing-override warnings, links verified against the current /reference/apis/ paths.

Refs #5142

PR #5193, opened 2026-07-13
Document UploadImageToDatasets on the data manager service

btshrewsbury-viam:
## What this adds

The Go SDK's data manager service gained **UploadImageToDatasets** ([rdk#5145](https://github.com/viamrobotics/rdk/pull/5145), DATA-4320) alongside the already-documented `UploadBinaryDataToDatasets`, but it was missing from the [data manager API reference](https://docs.viam.com/reference/apis/services/data/) and flagged by the weekly **SDK method coverage** job (#5142).

One CSV row, one proto description override (rewording the pre-staged file to the imperative style the method tables use), and the regenerated data manager markdown.

Split out of #5187 so this can be reviewed independently of the fleet-page methods — the two share no generated files.

## Merge order

**Stacked on #5184** because both regenerate the data manager markdown. Merge #5184 first; GitHub retargets this automatically. Refs #5142 but does not close it (remaining warnings tracked in #5188/#5189/#5190).

## Verification

Targeted coverage run: the warning cleared, no missing-override warnings, links verified against the current /reference/apis/ paths.

Refs #5142


gloriacai01 (approved):
thank you!!

PR #5200, opened 2026-07-14
Correct motion and frame system API reference accuracy

btshrewsbury-viam:
## What

Accuracy fixes for the motion-planning reference cluster, from the section review (verified against RDK v1.0.0 source and a live viam-server v1.0.0).

- **`api.md`**: The page said the builtin service supports "`Move`, `DoCommand`, and `GetStatus`, and nothing else" — contradicted by its own GetPose section and by the RDK: builtin implements `Move`, `GetPose` (deprecated, delegates to the frame system), `DoCommand`, and `GetStatus`, while `MoveOnMap`/`MoveOnGlobe`/`StopPlan`/`ListPlanStatuses`/`GetPlan` return `not supported by builtin` (live-confirmed exact string). Also: `executeCheckStart` is an optional key inside an `execute` DoCommand request, not a standalone command; documented the teleop commands; noted GetStatus is a common resource RPC omitted from the generated table.
- **`frame-system-api.md`**: A worked-example comment inverted the `Transform` field semantics (`reference_frame` names the new frame; `pose_in_observer_frame` carries its pose in the parent) — the code was right, the comment wasn't. Removed an unverifiable roadmap claim, named the Go client method for `TransformPCD` (`TransformPointCloud`), flipped negation-first passages, made the description action-oriented.
- **`kinematics.md`**: DH parameters are a `kinematic_param_type: "DH"` mode with a `dhParams` array, not inline in an SVA file (`rdk/referenceframe/model_json.go`); added the missing `links[].orientation` (cross-linking the orientation-vector reference) and `joints[].geometry` (prismatic only) schema fields; mimic `multiplier` defaults to 1.0.
- **`_index.md`**: idiom swap.

## Verification

prettier, markdownlint, vale (0 errors), `make build-prod` all pass. Builtin method surface confirmed against a live v1.0.0 server.

### DIFF
--- a/docs/motion-planning/reference/api.md
+++ b/docs/motion-planning/reference/api.md
@@ -34,7 +34,7 @@ Key parameters:
 
 ### MoveOnMap, MoveOnGlobe, StopPlan, ListPlanStatuses, and GetPlan
 
-Navigate mobile bases and manage the resulting plans. The builtin motion service returns a `not supported by builtin` error for each of these methods (for example, `MoveOnMap not supported by builtin`); module-based motion services implement them.
+Navigate mobile bases and manage the resulting plans. The builtin motion service returns a `not supported by builtin` error for each of these methods (for example, `MoveOnMap not supported by builtin`); module-based motion services may implement them.
 
 ### GetPose (deprecated)
 
@@ -46,8 +46,10 @@ Sends arbitrary commands. The builtin motion service supports a `"plan"`
 command (returns a trajectory without executing it), an `"execute"` command
 (runs a trajectory; add the `"executeCheckStart"` key to an execute request
 to verify the resource is at the trajectory's start first), and the teleop
-commands (`teleop_start`, `teleop_move`, `teleop_stop`, and `teleop_status`).
+commands (`teleop_start`, `teleop_move`, `teleop_stop`, and `teleop_status`),
+which stream smoothed, incremental joint commands for interactive arm
+teleoperation.
 
 ### GetStatus
 
-Returns generic resource status for the motion service. `GetStatus` is a common resource RPC, so the generated table above omits it. Use it for liveness checks.
+Returns generic resource status. `GetStatus` is part of the common resource API that every resource implements, including the builtin motion service; the generated table above omits common RPCs. Use it for liveness checks.
--- a/docs/motion-planning/reference/api.md
+++ b/docs/motion-planning/reference/api.md
@@ -10,7 +10,7 @@ aliases:
   - /appendix/apis/services/motion/
 ---
 
-The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service supports `Move`, `DoCommand`, and `GetStatus`, and nothing else.
+The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service implements `Move`, `GetPose` (deprecated), `DoCommand`, and `GetStatus`. `MoveOnMap`, `MoveOnGlobe`, `StopPlan`, `ListPlanStatuses`, and `GetPlan` return a `not supported by builtin` error; module-based motion services implement them.
 
 {{< readfile "/static/include/services/apis/generated/motion-table.md" >}}
 
@@ -25,10 +25,14 @@ Plans and executes a motion to a destination pose. This is the primary method fo
 
 Key parameters:
 
-- `component_name`: the arm or gantry to move
+- `component_name`: the component to move (typically an arm or gantry in this section's examples)
 - `destination`: a `PoseInFrame` specifying the target pose and reference frame
 - `world_state`: optional obstacles and transforms
 - `constraints`: optional linear, orientation, or collision constraints
+
+### MoveOnMap, MoveOnGlobe, StopPlan, ListPlanStatuses, and GetPlan
+
+Navigate mobile bases and manage the resulting plans. The builtin motion service returns a `not supported by builtin` error for each of these methods (for example, `MoveOnMap not supported by builtin`); module-based motion services implement them.
 
 ### GetPose (deprecated)
 
@@ -36,9 +40,12 @@ Returns a component's pose. Deprecated in favor of the robot service's `GetPose`
 
 ### DoCommand
 
-Sends arbitrary commands. The builtin motion service supports `"plan"`,
-`"execute"`, and `"executeCheckStart"` commands.
+Sends arbitrary commands. The builtin motion service supports a `"plan"`
+command (returns a trajectory without executing it), an `"execute"` command
+(runs a trajectory; add the `"executeCheckStart"` key to an execute request
+to verify the resource is at the trajectory's start first), and the teleop
+commands (`teleop_start`, `teleop_move`, `teleop_stop`, and `teleop_status`).
 
 ### GetStatus
 
-Returns generic resource status for the motion service. Useful for liveness checks.
+Returns generic resource status for the motion service. `GetStatus` is a common resource RPC, so the generated table above omits it. Use it for liveness checks.
--- a/docs/motion-planning/reference/api.md
+++ b/docs/motion-planning/reference/api.md
@@ -10,7 +10,9 @@ aliases:
   - /appendix/apis/services/motion/
 ---
 
-The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service implements `Move`, `GetPose` (deprecated), `DoCommand`, and `GetStatus`. `MoveOnMap`, `MoveOnGlobe`, `StopPlan`, `ListPlanStatuses`, and `GetPlan` return a `not supported by builtin` error; module-based motion services implement them.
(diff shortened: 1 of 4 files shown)


docs/motion-planning/reference/api.md
```diff
@@ -10,7 +10,7 @@ aliases:
   - /appendix/apis/services/motion/
 ---
 
-The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service supports `Move`, `DoCommand`, and `GetStatus`, and nothing else.
+The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service implements `Move`, `GetPose` (deprecated), `DoCommand`, and `GetStatus`. `MoveOnMap`, `MoveOnGlobe`, `StopPlan`, `ListPlanStatuses`, and `GetPlan` return a `not supported by builtin` error; module-based motion services implement them.
```
HipsterBrown:
```suggestion
The motion service exposes the methods below for planning and executing component motion. Most methods are implemented only by module-based motion services. The builtin service implements `Move`, `GetPose` (deprecated), `DoCommand`, and `GetStatus`.

`MoveOnMap`, `MoveOnGlobe`, `StopPlan`, `ListPlanStatuses`, and `GetPlan` return a `not supported by builtin` error; module-based motion services may implement them.
```
The unsupported methods should be moved to a new line. It was hard to notice the period between the supported and unsupported methods for the builtin service.

docs/motion-planning/reference/api.md
```diff
@@ -25,20 +25,27 @@ Plans and executes a motion to a destination pose. This is the primary method fo
...
 
-- `component_name`: the arm or gantry to move
+- `component_name`: the component to move (typically an arm or gantry in this section's examples)
 - `destination`: a `PoseInFrame` specifying the target pose and reference frame
 - `world_state`: optional obstacles and transforms
 - `constraints`: optional linear, orientation, or collision constraints
 
+### MoveOnMap, MoveOnGlobe, StopPlan, ListPlanStatuses, and GetPlan
+
+Navigate mobile bases and manage the resulting plans. The builtin motion service returns a `not supported by builtin` error for each of these methods (for example, `MoveOnMap not supported by builtin`); module-based motion services implement them.
```
HipsterBrown:
```suggestion
Navigate mobile bases and manage the resulting plans. The builtin motion service returns a `not supported by builtin` error for each of these methods (for example, `MoveOnMap not supported by builtin`); module-based motion services may implement them.
```

docs/motion-planning/reference/api.md
```diff
@@ -25,20 +25,27 @@ Plans and executes a motion to a destination pose. This is the primary method fo
...
 ### GetPose (deprecated)
 
 Returns a component's pose. Deprecated in favor of the robot service's `GetPose`. Python callers still use this motion-service method today; see the [Frame system API reference](/motion-planning/reference/frame-system-api/).
 
 ### DoCommand
 
-Sends arbitrary commands. The builtin motion service supports `"plan"`,
-`"execute"`, and `"executeCheckStart"` commands.
+Sends arbitrary commands. The builtin motion service supports a `"plan"`
+command (returns a trajectory without executing it), an `"execute"` command
```
HipsterBrown, line 46: How much should we document the usage of the builtin commands? Mentioning the teleop commands could be confusing without additional context about their intended usage.

docs/motion-planning/reference/api.md
```diff
@@ -25,20 +25,27 @@ Plans and executes a motion to a destination pose. This is the primary method fo
...
+(runs a trajectory; add the `"executeCheckStart"` key to an execute request
+to verify the resource is at the trajectory's start first), and the teleop
+commands (`teleop_start`, `teleop_move`, `teleop_stop`, and `teleop_status`).
 
 ### GetStatus
 
-Returns generic resource status for the motion service. Useful for liveness checks.
+Returns generic resource status for the motion service. `GetStatus` is a common resource RPC, so the generated table above omits it. Use it for liveness checks.
```
HipsterBrown: Is this method implemented by the builtin service?

docs/motion-planning/reference/frame-system-api.md
```diff
@@ -16,13 +16,12 @@ The RPCs for querying and transforming frame system poses live on the robot serv
 | `TransformPose`     | Convert a pose between reference frames.                                |
 | `TransformPCD`      | Convert a point cloud between reference frames.                         |
 
-The SDK client classes surface these as methods on the robot or
-machine client object (`machine.transform_pose`, `machine.get_pose`,
-and so on). You do not instantiate a separate frame system client.
+Call these methods directly on the robot or machine client
+(`machine.transform_pose`, `machine.get_pose`, and so on).
```
HipsterBrown: `get_pose` is not implemented on the robot client for Python: https://python.viam.dev/search.html?q=get_pose&check_keywords=yes&area=default

docs/motion-planning/reference/frame-system-api.md
```diff
@@ -106,13 +105,11 @@ gripperInWorld, err := machine.GetPose(
 
 ### Python goes through the motion service, Go through the robot service
 
-`GetPose` exists on two services in the proto: the robot service (current) and the motion service (older, deprecated, same parameter shape). The Go SDK surfaces the robot service version as `RobotClient.GetPose`; Go callers go through the robot service. The Python SDK never wrapped the robot service version, so Python callers use `MotionClient.get_pose` and hit the deprecated motion service path.
+`GetPose` exists on two services in the proto: the robot service (current) and the motion service (older, deprecated, same parameter shape). The Go SDK exposes the robot service version as `RobotClient.GetPose`; Go callers go through the robot service. The Python SDK never wrapped the robot service version, so Python callers use `MotionClient.get_pose` and hit the deprecated motion service path.
```
HipsterBrown, line 113: Go also has a FrameSystemService that modules can depend on for this information. We probably need to spend more time refining this information.

docs/motion-planning/reference/kinematics.md
```diff
@@ -75,7 +75,7 @@ Viam supports two kinematics file formats at the API level:
 | **SVA** (Spatial Vector Algebra) | Viam's native JSON format                     | When writing a Viam kinematics file directly.                |
 | **URDF**                         | XML format used by ROS and many manufacturers | When the manufacturer ships a URDF and you want to reuse it. |
 
-`GetKinematics` returns one of these two formats. Viam's SVA schema also accepts Denavit-Hartenberg parameters inline, as a convenience when you are converting a textbook DH table; DH is not a third API-level format.
+`GetKinematics` returns one of these two formats. The same JSON file format also accepts Denavit-Hartenberg (DH) parameters: set `kinematic_param_type` to `"DH"` and supply a `dhParams` array, a convenience when you are converting a textbook DH table. DH is not a third API-level format.
```
HipsterBrown:
```suggestion
`GetKinematics` returns one of these two formats. The same JSON file format also accepts Denavit-Hartenberg (DH) parameters: set `kinematic_param_type` to `"DH"` and supply a `dhParams` array, a convenience when you are converting a textbook DH table.
```

HipsterBrown (changes requested):

docs/motion-planning/reference/api.md
btshrewsbury-viam: Yes — verified two ways: `Status` is on the v1.0.0 `Resource` interface itself (every resource carries it, including builtin motion), and the section-review live-confirmed `GetStatus` against a viam-server v1.0.0. Reworded in e06aaab48 to attribute it to the common resource API rather than the motion service specifically, which also explains why the generated table omits it.

docs/motion-planning/reference/api.md
btshrewsbury-viam: Kept the four names but added the missing context in e06aaab48: they "stream smoothed, incremental joint commands for interactive arm teleoperation" (from `services/motion/builtin/teleop.go` at v1.0.0 — EMA-smoothed joint targets per component). My leaning is that a reference page should list every DoCommand the builtin actually accepts; happy to trim to a one-line mention without the names if you'd rather not surface them.

docs/motion-planning/reference/frame-system-api.md
btshrewsbury-viam: Right — the Python robot client never wrapped it (only Go's `RobotClient.GetPose` exists; Python goes through the deprecated motion-service path, as the GetPose section below describes). Fixed in e06aaab48: the example now uses `transform_pose`/`transform_pcd` (present in both SDKs) and states the per-SDK availability explicitly.

docs/motion-planning/reference/frame-system-api.md
btshrewsbury-viam: Good catch — at rdk v1.0.0, `robot/framesystem/framesystem.go:33` defines `PublicServiceName = "$framesystem"` ("used by modules to refer to the robot's frame system") with a Go `framesystem.Service` interface. e06aaab48 scopes the "there is none" claim to the public API and adds a Service-names paragraph documenting the Go module dependency path. If you think it deserves more than a paragraph (a worked module example, say), I'd split that into a follow-up.

HipsterBrown (approved):
Let's plan for a follow up PR that creates a reference page for the DoCommand verbs implemented by the builtin motion service.

PR #5219, opened 2026-07-19
Document motion service error for disconnected frame systems

shannonbradshaw:
The motion service now rejects all planning requests (Move, MoveOnGlobe, MoveOnMap, etc.) when the machine's frame system has parts that are not connected to the world frame. Previously, orphaned frames were silently ignored, which could lead to collisions. This PR documents the new error so users can diagnose and fix it.

## Source changes

- viamrobotics/rdk#6240: Disallow motion service API calls when the frame system is disconnected. Adds `NewFromServiceMustBeConnected` which returns an error listing unlinked parts if any frame's parent chain does not reach the world frame.

## Docs changes

- `docs/motion-planning/how-planning-works.md`: Added "Frames are not linked to the world frame" as item 1 in the "When planning fails" list, with explanation and fix guidance.
- `docs/motion-planning/debug-motion-with-cli.md`: Added a new troubleshooting section for the specific error message, with steps to identify and fix unlinked frame parents.

## How I found these

- Xref lookup: `flows.md` lists motion planning behavioral paths; `maintenance.md` staleness table maps `rdk/services/motion/` changes to motion planning docs.
- Grep matches: searched for "frame system", "world frame", "disconnected", "orphan", "unlinked" across all docs. Found 62 pages referencing frame system, narrowed to the two most relevant for documenting this new error condition.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01LnJk1PPu5KzH57qUvND5bi)_

### DIFF
--- a/docs/motion-planning/debug-motion-with-cli.md
+++ b/docs/motion-planning/debug-motion-with-cli.md
@@ -174,6 +174,37 @@ interpreted in an unexpected reference frame. Double-check the
 `reference_frame` on the target `PoseInFrame`: the same `(x, y, z)` in
 the arm's frame and in the world frame describes two different places.
 
+## Symptom: "Some frames are not linked to the world frame"
+
+The motion service refuses to plan if any frame in your configuration is
+not connected to the world frame through its parent chain. This happens
+when you rename a component without updating the `parent` field on
+components that were attached to it.
+
+### 1. Find the unlinked parts
+
+The error message lists which parts are unlinked. Run `print-config` to
+see each part's declared parent:
+
+```sh
+viam machines part motion print-config --part "my-machine-main"
+```
+
+Look for parts whose parent names do not match any other part in the
+output. A typo, a stale reference to a deleted or renamed component, or
+a circular reference all produce this error.
+
+### 2. Fix the parent references
+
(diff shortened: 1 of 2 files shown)


docs/motion-planning/debug-motion-with-cli.md
```diff
@@ -174,6 +174,37 @@ interpreted in an unexpected reference frame. Double-check the
...
+when you rename a component without updating the `parent` field on
+components that were attached to it.
+
+### 1. Find the unlinked parts
+
+The error message lists which parts are unlinked. Run `print-config` to
+see each part's declared parent:
+
+```sh
+viam machines part motion print-config --part "my-machine-main"
```
jeremyrose-viam, line 190: This matches other examples in this document, but for clarity, this would be better:

```sh
viam machines part motion print-config --part <part id>
```

Recommend a future revision to all --part flags on this page.

jeremyrose-viam (approved):

docs/motion-planning/debug-motion-with-cli.md
shannonbradshaw: Agreed — `<part id>` is clearer. Matched the existing convention on this page to keep the PR scoped, but a follow-up pass to update all `--part` flags on this page would be a good cleanup.

---

PR #5229, opened 2026-07-29
Replace removed Local module UI references with JSON config

shannonbradshaw:
## Source changes

- viamrobotics/app#12911 (APP-17413): Remove local modules from machine and fragment builder UI

## Docs changes

- `docs/tutorials/configure/pet-photographer.md`: Replace "click Local module, then Local module" and "click Local module, then Local component" instructions with JSON configuration examples, since those menu items no longer exist in the app UI.
- `docs/tutorials/custom/controlling-an-intermode-rover-canbus.md`: Same replacement — switch from removed Local module/Local component UI flow to JSON config instructions.
- `docs/tutorials/custom/custom-base-dog.md`: Same replacement — switch from removed Local module/Local component UI flow to JSON config instructions.

## How I found these

- Grep matches: searched entire docs repo for `Local module`, `Local component`, `Local service` — found 3 tutorial pages with explicit UI click instructions referencing the removed menu items.
- The `build-modules/write-a-driver-module.md` page also mentions "configure as a local module" (line 1055) but does not give specific UI click instructions — it already uses the `viam module reload-local` CLI workflow. No change needed there.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_011LHrm1TFUQtnwKGcEcQTvA)_


jeremyrose-viam:
Follow-up commit: fixed two command-accuracy bugs in the CLI-based workflow from the previous revision, verified live against the current CLI (v1.4.0) and a real machine, not just read from source.

1. **`viam module create --name=X`** had no `--org-id`/`--public-namespace`. Confirmed this fails outright: `Error: Must provide either org-id or public-namespace`. Added `--org-id=<your-org-id>` to all three tutorials, with a one-line pointer to `viam organizations list` or the org's Settings page to find it.

2. **`viam module reload-local ... --name <name>`** — confirmed live that `--name` does not name the added resource. Per the CLI's own flag help, `--name` identifies an existing module to restart (`pass at most one of --name, --id`); the actual flag for naming a newly-added resource is `--resource-name`. Tested directly: passing `--name flag-test-resource` produced a resource silently named `vision-1` instead. Since the next step in each tutorial references the resource by the name given here, this would have left readers with a working module under the wrong (auto-generated) name. Changed to `--resource-name` in all three files, matching the already-correct pattern in `build-modules/deploy-a-module.md` (the stale pattern this copied from was actually `write-a-driver-module.md`, which has the same `--name` mistake and predates this PR — a separate, existing issue not introduced here).

Both fixes are scoped to the CLI commands the previous revision introduced; the raw-JSON-config alternative from the PR's first commit never had this problem, since it didn't call either command.

Not addressed here, left for the upcoming human tutorial review: the previous revision's own note that stale screenshots (predating the UI change removing "Local module"/"Local component" menu items) need reviewer verification and replacement. The screenshots directly depicting the removed creation dialogs are already gone; the remaining ones (attribute editors, control tabs) haven't been independently verified against the current app layout.

shannonbradshaw:
Thank you for the live testing and fixes — both are real bugs I introduced:

1. **Missing `--org-id`**: I read the `viam module create` flag definition in the CLI source but didn't catch that it's required (not just optional). Good catch.

2. **`--name` vs `--resource-name`**: I copied the pattern from `write-a-driver-module.md` without verifying against the actual CLI flag behavior. The fact that `--name` silently succeeds but names the resource differently than expected makes this especially easy to miss from source alone — live testing was the right call.

I'll record the `--name` vs `--resource-name` discrepancy in `write-a-driver-module.md` as a separate finding for a future run.

---

jeremyrose-viam (approved):
Let's merge these changes as an improvement in accuracy over the current state. These documents should get a thorough human review as part of [DOCS-4931](https://viam.atlassian.net/browse/DOCS-4931).

PR #5236, opened 2026-07-30
Document part-level data retention policies

btshrewsbury-viam:
## Summary

Documents setting a data retention policy for a whole machine part instead of a single component.

- Adds a **Part-level retention policy** section to the data management reference (`docs/data/reference.md`), documenting the top-level `data_retention_policy` config key: where it goes in the part's JSON config, the `days` option and its constraints (positive whole number; other values ignored; UTC calendar days), that it covers binary and tabular data across every resource on the part, and that fragments can set it (with the part's own value taking precedence).
- Scopes the policy explicitly to the **default data store**: a hot data store keeps its own rolling window, set by `stored_hours`.
- Documents the interaction with resource-level `retention_policy`: the shorter of the two periods applies to that resource's data. Size limits (`binary_limit_gb`, `tabular_limit_gb`) remain resource-level only.
- Cross-links from the existing platform-managed capture settings table and from `docs/data/delete-data.md`.

## Source verification

Verified against `viamrobotics/app` main:

- Top-level key extraction: `domains/robotservice/config_retriever.go:724` and `domains/datamanagement/retention/retention.go` (`PartDataRetentionPolicyKey = "data_retention_policy"`, struct supports `days` only).
- Part-scope enforcement (binary + tabular, filtered by part ID only): `domains/datamanagement/part_level_data_retention_job.go`; deletion boundary is a UTC calendar-day cutoff (`retention.RetentionCutoff`); non-integer or non-positive `days` ignored.
- Shorter-period-wins: `retention.EffectiveRetentionDays` takes the stricter of part-level and resource-level policies; the part-level and resource-level deletion jobs run independently, so the earlier cutoff wins.
- Fragment support and precedence: fragment configs merge under the part's own config (`data/fragments.go`, `mergeFragmentMaps`), so a part-set policy overrides a fragment-set one.
- Size-based limits are read from resource-level configs only: `domains/datamanagement/data_retention_by_size_job.go`.

## Review

An adversarial source-verification pass tightened three points, and reviewer feedback was applied:

- Stated the UTC calendar-day deletion boundary precisely (the prior wording overstated retention by up to a day).
- Replaced the empty-arrays config example with one containing a component (policy extraction iterates a part's components and services, so the empty example modeled a config the platform never enforces).
- Documented fragment-vs-part precedence.
- Made the default-data-store scope explicit and dropped the deletion-frequency detail.

## Checks

- prettier, markdownlint, vale (0 errors/warnings/suggestions on changed files), `make build-prod` all pass.

https://claude.ai/code/session_01J5fWiUAkE24QEDiJL7wQfN

### DIFF
--- a/docs/data/reference.md
+++ b/docs/data/reference.md
@@ -291,7 +291,36 @@ The following settings are processed by the Viam cloud platform, not by `viam-se
 | `retention_policy` | object | `attributes` (sibling to `capture_methods`) | How long captured data is retained in the default data store. Options: `"days": <int>`, `"binary_limit_gb": <int>`, `"tabular_limit_gb": <int>`. Days are in UTC. |
 | `recent_data_store` | object | Inside a `capture_methods[]` entry | Store a rolling window of recent data in a [hot data store](/data/hot-data-store/) for faster queries. Example: `{ "stored_hours": 24 }` |
 
+A resource-level `retention_policy` covers one component or service. To apply one retention period to everything the part captures, set a [part-level retention policy](#part-level-retention-policy) instead.
+
 For remote parts capture, see [Capture from multi-part machines](/data/capture-sync/remote-parts-capture/).
+
+### Part-level retention policy
+
+A part-level retention policy applies one retention period to all data a machine {{< glossary_tooltip term_id="part" text="part" >}} captures, across every component and service. Once per day, the Viam cloud platform deletes captured data (binary and tabular) older than the configured number of days.
+
+Set the policy in **JSON** mode on the **CONFIGURE** tab. `data_retention_policy` is a top-level field in the part's configuration, alongside `components` and `services`:
+
+```json
+{
+  "components": [],
+  "services": [],
+  "data_retention_policy": {
+    "days": 30
+  }
+}
+```
+
+<!-- prettier-ignore -->
+| Name | Type | Description |
+| --- | --- | --- |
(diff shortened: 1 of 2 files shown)


docs/data/reference.md
```diff
@@ -291,8 +291,42 @@ The following settings are processed by the Viam cloud platform, not by `viam-se
...
 | `recent_data_store` | object | Inside a `capture_methods[]` entry | Store a rolling window of recent data in a [hot data store](/data/hot-data-store/) for faster queries. Example: `{ "stored_hours": 24 }` |
 
+A resource-level `retention_policy` covers one component or service. To apply one retention period to everything the part captures, set a [part-level retention policy](#part-level-retention-policy) instead.
+
 For remote parts capture, see [Capture from multi-part machines](/data/capture-sync/remote-parts-capture/).
 
+### Part-level retention policy
+
+A part-level retention policy applies one retention period to all data a machine {{< glossary_tooltip term_id="part" text="part" >}} captures, across every component and service. The Viam cloud platform deletes the part's captured data (binary and tabular) from the default data store when the retention period passes. The policy covers the default data store only: a [hot data store](/data/hot-data-store/) keeps its own rolling window of recent data, set by `stored_hours`.
```
katiepeters, line 300: also hot data store retention is only at the component capture method level

katiepeters (approved):

Issue #5198, opened 2026-07-14
World state store API page has no Go tabs: parse_go.py hardcodes a skip for the service


btshrewsbury-viam:
## Problem

The [world state store API reference](https://docs.viam.com/reference/apis/services/world-state-store/) documents Python and TypeScript only. Every method section in the generated include (`static/include/services/apis/generated/world_state_store.md`) has Python and TypeScript tabs but no Go tab.

This has become a visible gap: the Visualization section (#5099) teaches implementing the service **in Go** (`Publish visuals from a module`), so the API page contradicts the how-to that links to it — the language the docs teach is the one language the API page omits.

## Root cause

`.github/workflows/parse_go.py` skips the service unconditionally:

```python
for resource in viam_resources:

    if resource == "world_state_store":
        print(f'Skipping Resource: {resource}')
        continue
```

The skip landed with the page itself in DOCS-4335 (#4642, 2025-09-12), presumably because the Go package was not yet published to pkg.go.dev, so the scraper had nothing to read.

## Why it can be removed now

- `https://pkg.go.dev/go.viam.com/rdk/services/worldstatestore` now returns 200.
- RDK `v1.0.0-rc0` ships the full client interface with doc comments and usage examples (`services/worldstatestore/world_state_store.go`):

```go
type Service interface {
    resource.Resource
    ListUUIDs(ctx context.Context, extra map[string]any) ([][]byte, error)
    GetTransform(ctx context.Context, uuid []byte, extra map[string]any) (*commonpb.Transform, error)
    StreamTransformChanges(ctx context.Context, extra map[string]any) (*TransformChangeStream, error)
}
```

- `update_sdk_methods.py` already maps the service's proto (`world_state_store_grpc.pb.go`) and includes it in its `services` list, so only the Go parser opts out.

## Fix

1. Remove the `world_state_store` skip from `parse_go.py`.
2. Run the SDK-methods update and check the generated Go tabs for `ListUUIDs`, `GetTransform`, and `StreamTransformChanges` (the streaming return, `*TransformChangeStream` with its `Next()`/`io.EOF` iterator, is the one likely to need a template tweak).
3. Confirm the regenerated `world_state_store.md` and `world_state_store-table.md` build clean.

## Done when

The world state store API page shows Go alongside Python and TypeScript for all three service methods, matching the Go how-to at `/visualization/publish-visuals-from-a-module/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #5189, opened 2026-07-13
SDK coverage: scraper artifacts and SDK-only helpers flagged as undocumented methods


btshrewsbury-viam:
Part of the #5142 warning backlog (SDK method coverage). These warnings are not documentation gaps — the coverage scraper is reporting things that are not callable API methods, or are SDK-side helpers that the docs have never documented. Each needs either a scraper fix or an ignore-list entry.

## Not actually methods

**Python `resource_id`, `resource_type`, `role` (reported as app client methods)**
These are attributes of the `APIKeyAuthorization` dataclass (`viam.app.app_client.APIKeyAuthorization.resource_id`, etc.). The Python scraper reads the autoapi page for `app_client` and picks up dataclass attributes as if they were `AppClient` methods. Best fix: teach `parse_python.py` to skip non-method members (or members of non-client classes); fallback: ignore-list the three names.

**Python `GetStatus` on world_state_store (proto-cased duplicate)**
The world_state_store autoapi page exposes proto-cased wrapper entries alongside the snake_case client methods. The page's other duplicates (`ListUUIDs`, `GetTransform`, `StreamTransformChanges`, `DoCommand`) were already ignore-listed; PR #5184 added `GetStatus` to the same list. Listed here for completeness — a scraper-level fix for this page would let all five entries come off the ignore list.

## SDK-side helpers (released, but not platform API)

**Flutter `arm.calculateMaxReach`**
A client-side convenience that calls `getKinematics` and sums link lengths locally — no RPC behind it, and an unusual signature (a method that takes the `Arm` as a parameter). Recommend: ignore-list, and optionally flag to the Flutter SDK team as an API-shape question.

**Flutter `RobotClient.withClient`**
A constructor variant whose own doc comment says it is for callers that already have a `RobotServiceClient` "e.g. for testing with a mock client," and that channel/session-dependent methods will not work on instances it creates. Not a customer connection path (that is `atAddress`, which is documented). Recommend: ignore-list.

## Proposed resolution

A small PR that (a) skips dataclass attributes in `parse_python.py`, and (b) adds the two Flutter helpers to the Flutter ignore list in `check_for_unused_methods` with comments. Happy to open it once confirmed. Until then these warnings keep the weekly `SDK method coverage` job red.

---
_Generated by [Claude Code](https://claude.ai/code)_

jeremyrose-viam:
All four items here are resolved:

[...]

PR #5241, opened 2026-07-31
Fix three SDK-doc-generator scraper bugs (TypeScript whitespace nodes, shortcode-stripping regex)

jeremyrose-viam:
## Summary

While regenerating `static/include/*/apis/generated/` locally to validate #5234, `update_sdk_methods.py` crashed twice, and a successful run after those crashes surfaced a silent content-truncation bug. All three are fixed here; none are new regressions from this branch — all three bugs are already live on `main`.

1. **`parse_typescript.py:141`** — `TypeError: str.find() takes no keyword arguments`
   `parameters.find('ul', class_="tsd-parameter-list").children` iterates *every* child node of the `<ul>`, including whitespace-only text nodes between `<li>` elements. BeautifulSoup's `.children` yields both `Tag` and `NavigableString` (a `str` subclass); calling `.find(class_=...)` on a whitespace text node hits `str.find()` instead of `Tag.find()`. Fixed by switching to `find_all('li', recursive=False)`, which only iterates actual elements. (Same fix as already-open #5222, found independently before discovering that PR — recommend closing #5222 as superseded by this one.)

2. **`parse_typescript.py:169-171`** — `AttributeError: 'NavigableString' object has no attribute 'get'`
   Same root cause, different call site: `.next_sibling` returns the literal next node (which can be a whitespace `NavigableString`), not the next actual tag. Fixed with `.find_next_sibling()`, which skips text nodes.

3. **`update_sdk_methods.py:967,970`** — silent content truncation, no crash
   The shortcode-stripping regexes used for proto-override table descriptions matched from *any* opening shortcode to the *nearest* closing tag whose name started with `a` or `b`, with no requirement that it be the matching closer. This is fine for a file with a single paired shortcode (e.g. `board.SetPWMFrequency.md`'s `{{< alert >}}...{{< /alert >}}`, which is what 93487ff43/#5184 added this angle-bracket regex to handle, and does handle correctly). It breaks when a file has an earlier, unrelated self-closing shortcode ahead of a later paired block shortcode: `data_sync.TabularDataCaptureUpload.md` opens with a self-closing `{{< glossary_tooltip ... >}}` (no closing tag of its own) and later has an `{{< alert >}}...{{< /alert >}}` block. The old regex matched from the glossary tooltip all the way to the alert's closer, silently deleting everything in between — including "component to Viam" — leaving `data_sync-table.md`'s `TabularDataCaptureUpload` row truncated mid-sentence. Fixed by requiring the closing tag to match the opening tag's name via backreference (applied to both the `{{%...%}}` and `{{<...>}}` variants). Verified against every override file currently using either shortcode style: the previously-broken case now reproduces the original committed text exactly, and the previously-working case (`board.SetPWMFrequency.md`) is unaffected.

## How this was found

Evaluating #5234 (a hand-edit to a generated file) led to actually running the generator locally to produce the correct output instead. That surfaced bugs 1 and 2 as crashes; after fixing those, a successful regen run's diff included the truncated `data_sync-table.md` row, which traced back to bug 3.

## What this doesn't fix

Left out of scope for this PR, since they're different bugs in the same files that weren't touched here:
- #5194 (Go-tab loss for methods inherited via `framesystem.InputEnabled`) and its root-cause issue #5192
- #5198 (`parse_go.py` hardcoded skip for `world_state_store`)
- #5189, #5190, #5188 (scraper-artifact / coverage-policy issues in `check_for_unused_methods`)

## Expected follow-up

Once merged, plan to regenerate `static/include/*/apis/generated/` against current upstream SDK docs in a separate PR (content-only, no script changes), which should supersede #5234 and likely resolve #5142 ("CI failure: SDK method coverage" — same `parse()` code path, so worth confirming this was the actual cause of that failure before closing it).

### DIFF
--- a/.github/workflows/parse_typescript.py
+++ b/.github/workflows/parse_typescript.py
@@ -138,7 +138,7 @@ class TypeScriptParser:
                 param_object = {}
                 if method.find('div', class_="tsd-parameters"):
                     parameters = method.find('div', class_="tsd-parameters")
-                    parameter_list = parameters.find('ul', class_="tsd-parameter-list").children
+                    parameter_list = parameters.find('ul', class_="tsd-parameter-list").find_all('li', recursive=False)
 
                     for param in parameter_list:
                         param_name = param.find('span', class_="tsd-kind-parameter").text
@@ -166,9 +166,9 @@ class TypeScriptParser:
                 returns = md(str(method.find('h4', class_="tsd-returns-title"))).replace("#### Returns ", "").strip().replace('\\', '')
                 returns = " ".join(returns.split())
                 return_description = ""
-                if method.find('h4', class_="tsd-returns-title").next_sibling:
-                    if not method.find('h4', class_="tsd-returns-title").next_sibling.get('class'):
-                        return_description = md(str(method.find('h4', class_="tsd-returns-title").next_sibling)).strip()
+                if method.find('h4', class_="tsd-returns-title").find_next_sibling():
+                    if not method.find('h4', class_="tsd-returns-title").find_next_sibling().get('class'):
+                        return_description = md(str(method.find('h4', class_="tsd-returns-title").find_next_sibling())).strip()
                     else:
                         return_description = None
                 else:
--- a/.github/workflows/update_sdk_methods.py
+++ b/.github/workflows/update_sdk_methods.py
@@ -957,17 +957,19 @@ def write_markdown(type, names, methods):
                             proto_anchor_link = '/reference/apis/' + resource + '/#' + proto_link
 
                         ## Fetch just the first sentence from the proto_override_file (first text string terminated by '.\n'), ignoring hugo
-                        ## shortcodes like alerts ('{{%.*%}}.*{{% \[a-b].* %}}'), which precede some override files' (proto descriptions')
-                        ## first sentence:
+                        ## shortcodes like alerts ('{{% alert %}}...{{% /alert %}}'), which precede some override files' (proto descriptions')
+                        ## first sentence. The closing tag must match the same shortcode name as the opener (via backreference), so a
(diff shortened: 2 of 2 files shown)


btshrewsbury-viam (approved):
Reviewed with independent verification rather than taking the PR body's word for it; everything checks out. Approving.

**What was verified**

- **Bugs 1 & 2 (parse_typescript.py):** reproduced both crashes, then ran the real parse loop old-vs-new over live typedoc HTML from ts.viam.dev (6 pages, 170 methods). The crashes are real on today's HTML (bug 1 hits 165/170 methods, bug 2 hits 67/170). `find_all('li', recursive=False)` returns the identical element set the old code intended, and `find_next_sibling()` never lands on a different node than the old code did in any non-crashing case — zero behavioral divergence where the old code worked.
- **Bug 3 (update_sdk_methods.py regexes):** swept all 466 proto override files through the full extraction pipeline (including the downstream link-strip), old regexes vs new. Exactly **one** file changes: `data_sync.TabularDataCaptureUpload.md`, whose extraction goes from truncated mid-sentence to byte-identical with the committed `data_sync-table.md` row. No backtracking risk (worst contrived inputs ≤0.14s).
- **History:** git blame confirms all three bugs predate this branch (2024-05 through 2026-07). Agree #5222 should be closed as superseded — it is byte-identical to the bug-1 hunk here.

**Two non-blocking notes**

1. PR body nit: the old angle-bracket regex is actually a **no-op** for `board.SetPWMFrequency.md` (its alert sits after the first sentence). The file that regex actually rescues is `motion.MoveOnMap.md` (leading alert from #5040). Commit attribution (93487ff43) is correct; only the motivating-file claim is off.
2. Latent edge: the new backreference regexes stop stripping any shortcode whose opener contains a literal `%` (percent style) or `>` (angle style) in an attribute value, e.g. an alert titled "50% duty cycle". No current override file triggers this; worth remembering if one ever does.

btshrewsbury-viam (approved):

PR #5185, opened 2026-07-13
Document robot client GetPose, UploadDataFromPath, and missing TransformPCD/RestartModule methods

btshrewsbury-viam:
## What this adds

Four machine/robot API methods that exist in released SDKs but were missing from the [Machine Management API reference](https://docs.viam.com/dev/reference/apis/robot/), flagged by the weekly **SDK method coverage** job (#5142):

- **GetPose** (TypeScript `getPose`) — new proto row. Wraps the `RobotService.GetPose` RPC: returns a component's pose in the machine's frame system relative to a destination frame, with optional supplemental transforms.
- **UploadDataFromPath** (Python `upload_data_from_path`) — new proto row. Uploads a file or folder from the machine to the Viam cloud.
- **TransformPCD** — fills the missing Python method (`transform_pcd`) on the already-documented proto row.
- **RestartModule** — fills the missing Flutter method (`restartModule`) on the already-documented proto row.

## How

CSV rows/columns in `sdk_protos_map.csv`, proto description override files for the two new protos, and regenerated robot API markdown via `update_sdk_methods.py` (the regeneration also picks up the `/reference/` → `/dev/reference/` link drift in the table).

## Verification

Full local coverage run on this branch: all four warnings cleared, no missing-override warnings, sections render with the scraped signatures.

Part of a series addressing the #5142 warning backlog (see also #5184).

Refs #5142

PR #5242, opened 2026-07-31
Regenerate SDK API reference docs after generator bugfixes

jeremyrose-viam:
## What this adds

Content-only regeneration of the SDK API reference pages, now that #5241 fixed the two `BeautifulSoup` crashes and the shortcode-stripping regex bug in `update_sdk_methods.py`/`parse_typescript.py` that were blocking (or silently corrupting) regeneration. No new methods, no `sdk_protos_map.csv` or override-file changes — this just catches existing pages up to current upstream SDK state and to the fixed generator's output, wherever they'd drifted since their last regen.

This supersedes #5234, which hand-edited `audio_out.md`'s `play_stream` parameter order/type directly instead of regenerating it (introducing prettier-driven formatting noise inconsistent with every sibling file in the process). The `play_stream` fix itself was correct; this PR reproduces the same semantic fix via the generator, without the noise.

Files touched: `app.md`, `billing.md`, `data.md`, `data_sync.md`, `dataset.md`, `mltraining.md` (app APIs), `arm.md`, `audio_out.md` (components), `sensor.md`/`sensor-table.md` (components), `robot.md`.

## How

Ran `update_sdk_methods.py` (all four SDKs: Python, Go, TypeScript, Flutter; component/service/app/robot types) against current upstream SDK docs sites, on top of #5241's fixes. No CSV or override-file edits.

## Verification

- Diffed every changed file by hand rather than trusting the regen blind. Confirmed no `### <method>` sections were added or removed (`git diff | grep '^[-+]### '` is empty) — the diff is entirely within existing sections.
- No leaked raw HTML (`markdownify` misses) and no anomalous duplicated lines introduced.
- The prevalent `List[...]` → `list[...]` pattern across several files (33 occurrences) is real upstream drift, not a scraper artifact — spot-checked two cases directly against `python.viam.dev`: `viam.app.app_client` (all of `app.md`) has been migrated to PEP 585 lowercase generics; `viam.robot.client` (`robot.md`'s `GetOperations`) still uses `typing.List`. The mix in `data.md`/`robot.md`/etc. accurately reflects that the migration is partial upstream, module by module.
- `sensor.md`/`sensor-table.md`'s `Reconfigure` → `Reconstruct` description change reflects the already-merged `sensor.Reconfigure.md` override update from #5042 ("Resource.Reconfigure removal, shipped in RDK v1.0.0") — this page just hadn't been regenerated since.
- `data_sync-table.md` regenerated to a byte-identical match with its current committed content, confirming #5241's regex fix; it was the file that exposed the original truncation bug (`TabularDataCaptureUpload`'s description previously would have truncated to "...through a specific." mid-sentence with the pre-#5241 generator).
- Ran the CLAUDE.md pre-PR checks with the correct scope: `prettier`/`markdownlint` don't apply here — both required CI checks (`prettier-lint.yml`, `markdown-lint.yml`) are scoped to `docs/**`, and `static/` is explicitly excluded from markdownlint's `search_paths`, by design (these files are generator output, not hand-formatted prose — the generator's own style, e.g. single-quoted JS/TS strings, is the canonical committed style, and running prettier over it would just reintroduce the #5234 inconsistency-with-siblings problem). `vale` **does** run repo-wide and is a required, blocking check — ran it locally (`vale sync && vale <changed files>`): 0 errors, 0 warnings, 0 suggestions. `make build-prod` completes cleanly (only expected "old page date" warnings).

This doesn't touch `sdk_protos_map.csv`, so it doesn't move the #5142 method-coverage warning count — that's a separate, ongoing effort (#5184, #5185, and future companions). This is purely about keeping already-documented content in sync with upstream and validating #5241 end-to-end.

Refs #5142
Refs #5234

### DIFF
--- a/static/include/app/apis/generated/app.md
+++ b/static/include/app/apis/generated/app.md
@@ -57,9 +57,7 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/a
 
 ```ts {class="line-numbers linkable-line-numbers"}
 // This method is used internally only. To obtain a user's ID, use the listOrganizationsByUser method.
-const members = await appClient.listOrganizationMembers(
-  '<YOUR-ORGANIZATION-ID>'
-);
+const members = await appClient.listOrganizationMembers('<YOUR-ORGANIZATION-ID>');
 ```
 
 For more information, see the [TypeScript SDK Docs](https://ts.viam.dev/interfaces/AppClient.html#getuseridbyemail).
@@ -120,7 +118,7 @@ For more information, see the [Go SDK Docs](https://pkg.go.dev/go.viam.com/rdk/a
 
 **Returns:**
 
-- (Promise<undefined | [Organization](https://ts.viam.dev/classes/appApi.Organization.html)>): The new organization.
+- (Promise<[Organization](https://ts.viam.dev/classes/appApi.Organization.html) | undefined>): The new organization.
 
 For more information, see the [TypeScript SDK Docs](https://ts.viam.dev/interfaces/AppClient.html#createorganization).
 
@@ -140,7 +138,7 @@ List the {{< glossary_tooltip term_id="organization" text="organizations" >}} th
 
 **Returns:**
 
-- ([List[viam.proto.app.Organization]](https://python.viam.dev/autoapi/viam/proto/app/index.html#viam.proto.app.Organization)): :   The list of organizations.
+- ([list[viam.proto.app.Organization]](https://python.viam.dev/autoapi/viam/proto/app/index.html#viam.proto.app.Organization)): :   The list of organizations.
 
 **Example:**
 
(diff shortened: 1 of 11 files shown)


btshrewsbury-viam (approved):

btshrewsbury-viam (approved):
Adversarially reviewed with independent verification. Approving.

**Strongest evidence:** re-ran `update_sdk_methods.py` on the PR head across all components, services, app APIs, and robot (no language filter) — the result is **byte-identical** to the committed tree. That confirms no hand edits, no upstream drift since the run, a complete file list, and the `data_sync-table.md` byte-identical claim. Also verified: based on main with #5241 merged; no method sections added/removed; every deletion in the net −151 lines is reflow or type-annotation drift (no lost prose/params); `play_stream` matches live upstream and supersedes #5234 as described; "Reconstruct" is verbatim from #5042's merged override; the `List[`→`list[` partial migration is live-confirmed; vale 0/0/0; `make build-prod` clean.

**Non-blocking findings:**

1. **Three upstream docstring bugs now surface on docs pages** (all character-identical to live python.viam.dev, so the generator is faithful — but worth follow-up issues on the Python SDK): `data.md` `update_bounding_box` example calls the wrong method (`update_bounding_box_to_image_by_id`) and has a SyntaxError (missing comma); `robot.md` `transform_pcd` example uses an undefined `pcd` variable; `audio_out.md` example drops its `from_robot` setup line. Docs-side hand edits would recreate the #5234 problem, so upstream fixes are the right path.
2. **One raw-HTML leak:** `billing.md`'s `(Promise<Uint8Array<ArrayBuffer>>)` — browsers parse `<ArrayBuffer>` as an unknown element, so the deploy preview renders it as `(Promise<Uint8Array>)`. Generator-level fix (escape `<` in type strings), out of scope here.
3. PR-body nit: prettier/markdownlint are called "required CI checks" — they're `continue-on-error`; the only required checks are htmltest and vale. This only strengthens the scoping argument.

PR #5251, opened 2026-08-03
Pin @connectrpc/connect-node to v1 in Node.js setup docs

jeremyrose-viam:
## Summary

Pins `@connectrpc/connect-node` to `^1.7.0` in the Node.js setup guide's install command.

## What was wrong

`@viamrobotics/sdk` depends on the Connect-ES **v1** API (`@connectrpc/connect@^1.7.0` even in the latest SDK release, 0.74.2). The guide's install command doesn't pin a version for `@connectrpc/connect-node`, so a fresh `npm install` resolves the latest release, which is **2.1.2** — Connect-ES v2. That version mismatch breaks the SDK's transport at runtime:

```
TypeError: Cannot read properties of undefined (reading 'typeName')
    at createMethodUrl (.../@connectrpc/connect/dist/esm/protocol/create-method-url.js:32:44)
    at Object.unary (.../@connectrpc/connect/dist/esm/protocol-grpc/transport.js:72:26)
    at Object.authenticate (.../@viamrobotics/sdk/dist/main.es.js:3773:23)
```

Found this by hitting the crash directly while validating #5250 — following this guide's own install command with an unpinned `@connectrpc/connect-node` reproduces it for anyone today.

## Fix

Pin the install command to `@connectrpc/connect-node@^1.7.0`, matching the SDK's own dependency range, with a short note explaining why the pin matters.

## Verification

`npx prettier --check`, `npx markdownlint-cli`, and `vale` all pass on the changed file; `make build-prod` completes with no new errors.

## Related

- Split out of #5250, which hit this while validating a separate fix
- The same unpinned-install issue exists in `viam-typescript-sdk`'s own `Node.md` (different repo, tracked separately)

### DIFF
--- a/docs/build-apps/setup/node.md
+++ b/docs/build-apps/setup/node.md
@@ -31,10 +31,8 @@ npm init -y
 Node.js needs three runtime packages:
 
 - `@viamrobotics/sdk` — the Viam SDK
-- `@connectrpc/connect-node` — the Node-compatible gRPC transport
+- `@connectrpc/connect-node@^1.7.0` — the Node-compatible gRPC transport
 - `node-datachannel` — a WebRTC implementation for Node
-
-The SDK depends on the Connect-ES v1 API, so pin `@connectrpc/connect-node` to the matching major version. Installing it unpinned resolves to the incompatible v2 line and fails with a `TypeError` when the SDK tries to connect.
 
 Install everything at once:
 
@@ -152,6 +150,7 @@ If the script hangs or errors, the most common causes are:
 - **Polyfills registered too late.** The polyfill loop and transport factory must run before the first call into the SDK. If you split them into a separate module that is imported after the SDK, the SDK initializes without them.
 - **`.env` not loaded.** The `tsx --env-file=.env` flag loads the file. If you run the script a different way (`node`, `ts-node`, a bundler), use `dotenv` or your runtime's env loader instead.
 - **Wrong transport.** If you see `Unsupported HTTP version` or similar, confirm that `@connectrpc/connect-node` is installed and the transport factory is registered on `globalThis.VIAM`.
+- **Wrong `@connectrpc/connect-node` version.** The SDK depends on the Connect-ES v1 API. If `@connectrpc/connect-node` gets installed unpinned, it resolves to the incompatible v2 line and fails with a `TypeError` when the SDK tries to connect. Pin it to `^1.7.0`, as in the install command above.
 
 ## Next
 
--- a/docs/build-apps/setup/node.md
+++ b/docs/build-apps/setup/node.md
@@ -73,7 +73,6 @@ const wrtc = require("node-datachannel/polyfill");
 const connectNode = require("@connectrpc/connect-node");
 
 // Register a Node-compatible gRPC transport.
-// @ts-expect-error -- globalThis.VIAM is not in standard types
 globalThis.VIAM = {
   GRPC_TRANSPORT_FACTORY: (opts: any) =>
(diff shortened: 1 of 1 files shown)


btshrewsbury-viam (changes requested):
Adversarially reviewed by following the page literally in a scratch project, both pinned and unpinned. Commit 1 is verified correct; commit 2 needs a revert — and the bad advice that prompted it was mine, so apologies for the runaround.

**Commit 1 (the pin) — verified end-to-end, ship it:**
- Unpinned install resolves `connect-node@2.1.2` and the page's script fails at connect with the exact documented error (`TypeError: Cannot read properties of undefined (reading 'typeName')` in `createMethodUrl` → `authenticate`).
- Pinned install resolves `1.7.0` (the latest 1.x, deduped cleanly against the SDK's own `connect@1.7.0`), and the page's instructions run verbatim all the way to the expected dummy-credential `permission_denied` — no pre-auth failure.
- Not stale-on-arrival: SDK latest (0.74.2) still depends on Connect-ES v1. prettier/markdownlint/vale/build-prod all pass.

**Commit 2 (`@ts-expect-error` removal) — please revert.** My #5250 comment claimed the directive was dead here too; that was wrong. It's dead in the 7 sample files because they use ES `import`, which pulls in the SDK's `declare global { var VIAM ... }`. This page's snippet uses `require()`, so that declaration never enters the TS program — without the directive, `tsc` reports TS2339 (`Property 'VIAM' does not exist on type 'typeof globalThis'`; TS7017 under stock config), and every type-aware editor now shows an error on the page's canonical example. Tested both ways with SDK 0.74.2: directive present = clean, absent = error. Runtime is unaffected (tsx strips types), so revert is cosmetically safe — or switch the snippet to `import`, but that's a bigger change than this PR needs. If you revert, the directive's comment could say why it's needed here: `// @ts-expect-error -- the require()-style snippet doesn't load the SDK's global VIAM declaration`.

**Two optional nits:**
- `npm install ... @connectrpc/connect-node@^1.7.0` with an unquoted `^` fails under zsh with `extendedglob` set (`no matches found`) — quoting the package spec avoids a confusing dead-end for some mac users.
- "Installing it unpinned resolves to the incompatible v2 line" → "An unpinned install resolves to the incompatible v2 line".

btshrewsbury-viam (approved):
Approving — both follow-up commits verified independently, all earlier requests addressed.

**015357831 (directive restore + quoting):** Reproduced your `tsc` result from the page's exact install commands: with the directive the require()-style snippet typechecks clean, without it `TS2339: Property 'VIAM' does not exist on type 'typeof globalThis'`. Also confirmed under TypeScript 5.9.3 and 7.0.2 (unpinned `typescript` now resolves 7.x), with and without `skipLibCheck`, and under `nodenext` — the directive is load-bearing in every configuration. Quoted spec is byte-identical in bash/dash and protects zsh-extendedglob and cmd.exe.

**a6001d549 (rationale → troubleshooting):** Sound placement — the dependency bullet now carries `@^1.7.0` itself, so one-at-a-time installers still see the pin. Re-confirmed the bullet's claim: unpinned resolves connect-node 2.1.2 and fails locally with `TypeError ... 'typeName'` during authenticate; pinned 1.7.0 reaches signaling and gets a genuine `permission_denied` with dummy creds.

Gates: prettier, markdownlint, vale (0 errors), `make build-prod` all pass; branch touches one file, +4/−3.

One pre-existing note, not for this PR: the snippet's top-level `const VIAM` collides (TS2451) with the SDK's ambient `declare global { var VIAM }` if a reader adds a typed SDK import elsewhere in the same tsconfig — present on main before this PR, fine as a someday follow-up.

PR #5271, opened 2026-08-18
CSS/Docsy adjustments for navbar spacing

jeremyrose-viam:
## Issue

Spacing between the `DOCS` and `TUTORIALS` items in the nav menu is out of alignment. In addition, the gray left margin line of `DOCS` and the `:before` element of `TUTORIALS` render ever so slightly differently.

## Solution

- remove the `mt-2` Bootstrap class from the element in the template.
- adjust our existing site-specific SCSS tweaks in `assets/scss/_styles_project.scss`:
    - change flex alignment and padding for navbar container
    - add padding to `DOCS` span to increase grey line height
    - switch `TUTORIALS` element to use `border-left` for the line instead of `:before`
    - set nav-link top and bottom padding to 0

## Testing

Built and previewed locally.


js6450 (approved):
Looks great to me! Thanks for fixing this!

PR #5277, opened 2026-08-21
Update pick-and-place uArm workshop (phases 1-6)

elizafarley:
## Summary

Updates phases 1-6 of the vision-guided pick-and-place workshop based on a run-through with the uArm/xArm6 hardware, tightening wording, adding missed steps, and clarifying the frame-conversion logic in the perception phase.

- **Terminology**: match the app's actual tab casing throughout (`Configure`, `Logs`, `Control`, `Connect`, `Test` instead of all-caps `CONFIGURE`/`LOGS`/`CONTROL`/`CONNECT`/`TEST`).
- **Phase 1**: clarify that clicking **Grab**/jogging from the Control tab bypasses the config path entirely; add a `prev` frontmatter link back to the workshop index; fix `host` attribute description (it's an IP address) and the updated add-component flow (`Add to machine` step).
- **Phase 2**: update the RealSense discovery service's model name (`realsense/discovery` instead of `discovery / realsense:discovery`) and switch the `align_color_depth` attribute example from a bullet to a JSON snippet; note the correct joint numbering (0-5, not 5/6) for the arm's Control card test.
- **Phase 3**:
  - document a third way to jog the arm into position: manual mode via `DoCommand` with `enter_manual_mode`.
  - rework the obstacle-positioning workflow: replace the old "measure with a tape measure and type in `REPLACE_WITH_MEASURED_*` placeholders" approach with placeholder values plus a new **Position the obstacles** section that uses the 3D scene tab's build mode to drag obstacles into place visually.
  - add a warning about keeping the bin out of the camera's view in home pose so the vision service doesn't try to pick it up.
- **Phase 5** (the most substantive change): introduce an explicit **Convert to world before you move** step. Previously the workshop tagged both the approach and grasp poses with `reference_frame="cam-1"`, which is only valid for the first move in a chain — by the second move the wrist camera has already relocated with the arm, so a `cam-1`-tagged grasp pose resolved against the wrong camera position. The fix uses `RobotClient.transform_pose` to convert both poses to the fixed `world` frame once, right after detection, before either move executes. Also adds a new troubleshooting entry for the resulting symptom (arm overshoots into the table). Fixes the segmenter model name to `vision/detections-to-segments` (matching the `viam:vision:detections-to-segments` triplet in the companion config).
- **Phase 6**: matches updated Configure UI copy ("Proceed" button, "Viam-hosted" module flow).
- **Index**: adds an intro sentence describing what the workshop builds.
- Adds two new screenshots (`3d-scene-tab.png`, `auth-details.png`) referenced by phases 2 and 4.

Ran the full pre-PR check suite (prettier, markdownlint, vale, `make build-prod`) per `CLAUDE.md`; a prettier formatting fix is included as a follow-up commit.

Companion repo PR here: https://github.com/viam-devrel/pick-and-place/pull/1

## Test plan

- [ ] Review phase-by-phase diffs against actual uArm/xArm6 hardware behavior
- [ ] Spot-check the new screenshots render correctly in the built site
- [ ] Verify the Phase 5 `transform_pose`/world-frame conversion code runs correctly end-to-end on hardware


docs/tutorials/pick-and-place/03-static-positions.md
```diff
@@ -86,19 +88,30 @@ The steps in this phase move the physical arm, both when you jog it into positio
 
 With the `home-pose` switch added, save and verify it. Home is your observation pose, so jog the arm to a spot above the workspace where the wrist camera has a clear, unobstructed view of the blocks; Phase 5 detects from exactly this pose.
 
-You have two ways to jog the arm on its CONTROL card, and you can mix them:
+You have three ways to jog the arm on its Control card, and you can mix them:
 
-- **Joint control (`MoveToJointPositions`)** sets each joint angle directly. Move one joint slider a small amount, press **Execute**, and that joint rotates. It is the most predictable way to make coarse changes and to lift the arm clear before repositioning.
+- **Joint control (`MoveToJointPositions`)** sets each joint angle directly. Move one joint slider a small amount, press **Execute**, and that joint rotates. It is the most predictable way to make coarse changes and to lift the arm clear before repositioning. You can also use quick move mode (indicated by the lightning bolt icon) to move a joint five percent at a time.
```
HipsterBrown:
```suggestion
- **Joint control (`MoveToJointPositions`)** sets each joint angle directly. Move one joint slider a small amount, press **Execute**, and that joint rotates. It is the most predictable way to make coarse changes and to lift the arm clear before repositioning. You can also use quick move mode (indicated by the lightning bolt icon) to move a joint five degrees at a time.
```

docs/tutorials/pick-and-place/03-static-positions.md
HipsterBrown:
```suggestion
This matters both for correctness and for safety. Without the table obstacle, the planner might find a path that swings the arm through the table surface. Virtual safety walls at the workspace boundary also prevent the arm from swinging into people standing nearby. This is the same pattern you would use to keep a production workcell's motion planner honoring the real boundaries of its cell.
```
Missed this previously. We should remove this Claude-prose.

docs/tutorials/pick-and-place/05-perception-guided-picking.md
```diff
@@ -189,53 +197,93 @@ def offset_pose(pose: Pose, z_offset_mm: float) -> Pose:
...
+{{< /expand >}}
 
 {{< checkpoint >}}
-Before wiring up the moves, print `approach_pose` and `grasp_pose` and compare their `z` to `obj_in_cam.pose.z`: the approach `z` should sit about 100 mm toward the camera and the grasp `z` about 60 mm toward it. If either offset went the wrong way, the pick drives into or well short of the block, so fix the sign before running a move.
+Before wiring up the moves, print `approach_pose` and compare its `z` to `obj_in_cam.pose.z`: the approach `z` should sit about 100 mm toward the camera.
 {{< /checkpoint >}}
 
+## Descend in the gripper frame
+
+The approach move runs first, tagged `reference_frame="cam-1"`. At that instant the arm is still at `home-pose`, so `cam-1` sits exactly where the detection assumed and the motion service resolves the standoff correctly. The trouble only starts with the _second_ move: once the arm travels to the standoff it carries `cam-1` with it, so any further pose you tag `cam-1` resolves against the camera's new position, not the one your offset math used.
```
HipsterBrown:
```suggestion
The approach move runs first, tagged `reference_frame="cam-1"`. At that instant the arm is still at `home-pose`, so `cam-1` sits exactly where the detection assumed and the motion service resolves the standoff correctly. This assumption shifts after that move; once the arm travels to the standoff it carries `cam-1` with it, so any further pose you tag `cam-1` resolves against the camera's new position, not the one your offset math used.
```
Slight tweak in prose. Indicating "trouble" before any issue has occurred is distracting to the point of this section.

docs/tutorials/pick-and-place/05-perception-guided-picking.md
```diff
@@ -189,53 +197,93 @@ def offset_pose(pose: Pose, z_offset_mm: float) -> Pose:
...
+
+Rather than chase the camera frame, re-anchoring the block pose after every move, you sidestep it. The gripper is now parked directly above the block at the standoff, already lined up to come straight down. All that is left is to drop the last stretch, and that distance is a property of the gripper, not the camera: descend `grasp_distance` millimeters along the gripper's own `z` axis.
```
HipsterBrown:
```suggestion
Rather than chase the camera frame, re-anchoring the block pose after every move, you sidestep it. The gripper is now parked directly above the block at the standoff, already lined up to come straight down. All that is left is to drop the last stretch, and we can describe that move in the frame of the gripper instead of the camera: descend `grasp_distance` millimeters along the gripper's own `z` axis.
```

docs/tutorials/pick-and-place/06-inline-module.md
```diff
@@ -30,32 +30,37 @@ For your own control code, reach for a module when one of these is true for your
...
+The detection, the pose math, and the motion calls are the same pick-and-place logic, moved into a module's lifecycle methods with no change to what they do. Only the wiring around that logic changes, in a few places:
 
+- **No connection code necessary.** The module is part of the machine, so authentication is no longer necessary.
 - **How the module gets its resources.** A script calls `from_robot` after it connects; a module receives its resources through **dependency injection** instead.
+- **Config-driven resources.** Your local script uses hardcoded component names like "arm-1." The module reads those names dynamically, which allows a module to be used across different machines that may have different naming conventions.
 - **How the logic gets triggered.** A script runs top to bottom under `__main__`; a module runs its logic behind a **`do_command`** entry point that something else calls.
+- **Hard constraints in class structure.** The model identifier and class name are auto-generated and must not change.
```
HipsterBrown: This is only true for inline modules.

HipsterBrown (changes requested):

docs/tutorials/pick-and-place/06-inline-module.md
elizafarley: Addressed in latest revision.

HipsterBrown (approved):

